PR #7, opened 2018-06-07
Add TypeScript definitions for Electron

felixrieseberg:
As you switch versions in the dropdown, we'll magically ✨update the editor with TypeScript definition files for the chosen version of Electron.

That means that the editor will auto-complete Electron-related code <3


zeke:
:meta

PR #22, opened 2018-06-14
Fix first electron version install

gerges:
screen shot 2018-06-14 at 3 07 13 pm

Currently on first run (or running after deleting `~/Application Support/electron-fiddle`) the first electron version download is never reflected in the app.

At the moment `new BinaryManager` is responsible for downloading the first version of electron, which happens in a way `AppState` isn't aware of. A second call to `downloadVersion`, initiated by `AppState` does happen, but at the moment it resolves early as "already being downloaded". By removing the `new BinaryManager` side-effect, everything works as expected.

### DIFF
--- a/src/renderer/binary.ts
+++ b/src/renderer/binary.ts
@@ -19,10 +19,6 @@ const eDownload = promisify(require('electron-download'));
 export class BinaryManager {
   public state: StringMap<'ready' | 'downloading'> = {};
 
-  constructor(version: string) {
-    this.setup(version);
-  }
-
   /**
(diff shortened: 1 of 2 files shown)


felixrieseberg:
Sweet, thank you! Do you want to add [yourself to the contributors whitelist](https://github.com/electron/fiddle/blob/master/tools/fetch-contributors.js#L10-L22) for some credits in the settings?

felixrieseberg:
@gerges I'll just merge this in, feel free to submit another PR so that your name shows up in the credits!

gerges:
Haha, we'll wait until I can get something more meaty merged in

PR #25, opened 2018-06-25
Move to less, improve dev tools

felixrieseberg:
I can already see where this is going: Eventually, we'll move to Webpack. That's cool. For now, I've added `less`, turned all our CSS to less, and added a little bit of developer tooling - most notable, a simple file watcher that'll recompile TypeScript and Less on-the-fly.

Nothing fancy, but good enough until Fiddle is grown-up enough to deserve the time it needs to implement Webpack.


tools/generateAssets.js
```diff
@@ -1,9 +1,11 @@
 /* tslint:disable */
 
-const { maybeFetchContributors } = require('./fetch-contributors');
-const { generateTypeScript } = require('./tsc');
+const { maybeFetchContributors } = require('./contributors');
+const { compileTypeScript } = require('./tsc');
+const { compileLess } = require('./lessc');
 
 module.exports = async () => {
```
MarshallOfSound: Seems silly but can we be Super Fast(tm) and do

```js
await Promise.all([maybeFetchContributors, compileTypescript, compileLess])
```

PR #55, opened 2018-08-09
fix: Install beta.22 of Fiddle

felixrieseberg:
Installing `latest` has unwelcome side effects (mostly, that we're installing older versions). Let's go with the `beta` for now, it does enough for us here.


malept:
Why not beta.27?

PR #58, opened 2018-08-09
feat: Update icons

felixrieseberg:
Icons! They're basically garbage (Hi, it's me, your designer), but they'll do for a first release.


forge.config.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
         authors: 'Electron Community',
         exe: 'electron-fiddle.exe',
         iconUrl: 'https://github.com/electron/electron/blob/fe0f203/atom/browser/resources/win/atom.ico?raw=true',
```
malept: Change the `iconUrl` too? Or is that post-PR?

felixrieseberg: Post-PR, just so that we have a stable path.

PR #64, opened 2018-08-10
Add missing script tag for renderer process

cmpscabral:


felixrieseberg:
Thank you so much! By the way, we have a credits page in the settings - if you’d like to show up there, go and add your username to the whitelist: https://github.com/electron/fiddle/blob/master/tools/contributors.js

PR #71, opened 2018-08-13
fix: Do not use "showItemInFolder" on save

### LINKED ISSUE #69: Save (Ctrl+S) opens the main folder where the project resides
When I hit Ctrl+S to save Nautilius opens the parent folder of where the project is created.
Is this a bug or a feature?

running Ubuntu 18.04.1

felixrieseberg:
It's unclear to me why I added this call in the first place. Showing the item on save makes little sense 😅 

Closes #69

### DIFF
--- a/src/renderer/file-manager.ts
+++ b/src/renderer/file-manager.ts
@@ -103,9 +103,6 @@ export class FileManager {
         }
       }
 
-      // Show in folder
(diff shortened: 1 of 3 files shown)


felixrieseberg:
🤔 Also, it's unclear to me why `static/templates/ipc/index.html` is marked as changed. The lines shown as added here are already in that file?

Issue #60, opened 2018-08-10
Support electron < 1.8.x


bpasero:
It looks like the code in main does not execute when you pick Electron < 1.8.x, due to the missing `loadFile` method:

![image](https://user-images.githubusercontent.com/900690/43941876-1b5f4700-9c76-11e8-9cf4-48bea88da012.png)

savokiss:
We need electron 1.6.5, seems like there are no options 😶

malept:
I think it's more complicated than that:

* Should Fiddle support version series of Electron that the core team does not maintain anymore?
* If so, how far back does Fiddle need to support?
* Does Fiddle provide boilerplate Electron code based on the version selected? For example, #77 makes it pretty clear that `loadFile` makes the code easier to read.

nctl144:
I believe this issue has been resolved in #77 with modification from @felixrieseberg

felixrieseberg:
It was! Thank you for noticing, I'll close this now.

PR #77, opened 2018-08-20
fix: Support Electron <1.8.x

nctl144:
This PR aims to solve the issue #60 . By replacing **loadFile** method with the **loadURL** method, based on the PR electron/electron#11565 😄 Please let me know if this works :)

### DIFF
--- a/src/content/main.ts
+++ b/src/content/main.ts
@@ -1,5 +1,7 @@
 export const main = `// Modules to control application life and create native browser window
 const {app, BrowserWindow} = require('electron')
+const path = require('path')
+const url = require('url')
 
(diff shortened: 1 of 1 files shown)


felixrieseberg:
Thank you so much, this is a great PR! I'd love to have you in the credits - we're using a whitelist to make sure we're not publishing your name against your wishes. Do you want to add yourself? You can [just add your name here](https://github.com/electron/fiddle/blob/master/tools/contributors.js#L10-L23).

nctl144:
Thank you so much for the review @felixrieseberg ! It would be really great to add myself to the contributor list! I just created another PR for that :)

Issue #61, opened 2018-08-10
Toolbar covers editor widget


rebornix:
* Launch Fiddle
* Run Go To Definition on `BrowserWindow` in the first line
* A warning message pops up but it's covered by the toolbar
* You can see the same issue for Hover and other similar editor widgets as well

![editorindex](https://user-images.githubusercontent.com/876920/43967553-7cf18f52-9c79-11e8-87b1-5bd0132f3da9.gif)

wsmd:
Hey, I thought I'd drop a a hint..

It turns out the tooltip isn't cover by the toolbar. It's trimmed by the editor's boundaries.

This is caused by the `overflow: hidden` applied by the [mosaic](https://github.com/palantir/react-mosaic) window [wrapping the editor](https://github.com/electron/fiddle/blob/ff3fbac7ed63b76c38f039e72039379c73fbf155/src/renderer/components/editors.tsx#L95-L97).

The element selector is located at `.mosaic-window.mosaic-drop-target.main`

A more precise selector (copied via devtools):

```css
/* may or may not work based on the app state */
#app > div > div.mosaic-blueprint-theme.mosaic.mosaic-drop-target > div.mosaic-root > div:nth-child(1) > div
```

Removing the styling, _while it breaks a bunch of other things_, seems to fix this specific problem.

![screen shot 2018-08-13 at 4 26 41 pm](https://user-images.githubusercontent.com/2100222/44056243-ac5bb986-9f15-11e8-9712-9b01ef368761.png)

I'm assuming this is something out of Electron Fiddle's control. Perhaps a quick workaround would be to add padding to the mosaic window wrapper to give the first line tooltips some breathing room:

![screen shot 2018-08-13 at 4 39 38 pm](https://user-images.githubusercontent.com/2100222/44056942-c7d91f94-9f17-11e8-833f-ea6401cd9a89.png)

--

a little note: that tooltip looks somehow gigantic compared to other types of tooltips. E.g.

![screen shot 2018-08-13 at 4 43 49 pm](https://user-images.githubusercontent.com/2100222/44057080-1f4b86d6-9f18-11e8-99f3-e874cb802e83.png)

or compared to the same tooltip taken from VSCode

![screen shot 2018-08-13 at 4 42 22 pm](https://user-images.githubusercontent.com/2100222/44057102-2e4d407a-9f18-11e8-920b-7d574c430013.png)

I think resolving that font-size issue could help with making the padding hack mentioned above less drastic.

[...]

PR #78, opened 2018-08-21
fix: Adjust .mosaic-window-body styles

BeniCheni:
Refs #61 

After analyzing the details & suggestions in #61 & inspecting the `.mosaic-window-body` element's styles, this PR attempts to simply adjust the styles to resolve the observation in #61.
* add padding top allowance to support a visible tooltip on any variable on code line 1
* adjust font-size to 12px; local test seems to look okay.

Happy to update to any suggested styles to make the UI look nice. 🙏 for your review in advance.

electron-fiddler

### DIFF
--- a/src/less/mosaic.less
+++ b/src/less/mosaic.less
@@ -34,6 +34,9 @@
     flex-wrap: nowrap;
     align-items: stretch;
     align-content: stretch;
+    padding-top: 20px;
+    background-color: #1d2427;
+    font-size: 12px;
   }
 
   .mosaic {


felixrieseberg:
Thank you so much, this is a fantastic PR! I'd love to have you in the credits - we're using a whitelist to make sure we're not publishing your name against your wishes. Do you want to add yourself? You can [just add your name here](https://github.com/electron/fiddle/blob/master/tools/contributors.js#L10-L23).

PR #89, opened 2018-08-24
test: use enzyme-to-json and fix tests

aaronabramov:
i was trying to run tests locally and they were failing (i think it was a version mismatch or something)

this PR fixes tests by doing two things:
1. use `ensyme-to-json` package to serialize enzyme wrappers, so we only print the HTML and not the internals of enzyme (which seem to be different sometimes)
2. `delete window.localStorage` in setup file, because sometimes it can't be simply replaced by `window.localStorage = {}`


felixrieseberg:
@aaronabramov Let's merge this in!

PR #91, opened 2018-08-28
infra: @jest-runner/electron

aaronabramov:
use jest electron runner to run tests in a real electron process.
Some of the tests relied on an electron mock so i kept them the way the are, but the rest should be running in a real electron.


felixrieseberg:
There hasn't been much movement here in a while, so I'm closing this for now. We could always reopen it!

Issue #85, opened 2018-08-22
Unable to copy the entire GitHub Gist URL


bpasero:
I seem to only be able to select until here:

![image](https://user-images.githubusercontent.com/900690/44456211-3be6ea80-a600-11e8-827d-ba2769ad76f3.png)

This is with macOS

TekSiDoT:
This currently seems to work as intended.

While the element does seem to be a single input element, it actually is a span followed by an input element.

felixrieseberg:
It does suck though, so we should probably fix that.

BeniCheni:
Is there any hint to approach this fix?

(Can use this evening later to dig deeper on my own. But if there’s any hint, that would point my debugging to the right code area, perhaps? Still familiarizing self to the code base)

TekSiDoT:
This would be a great place to start: [Line 112 of address-bar.tsx](https://github.com/electron/fiddle/blob/ff3fbac7ed63b76c38f039e72039379c73fbf155/src/renderer/components/address-bar.tsx#L112)

BeniCheni:
Opened #94 to try this issue. Looking forward for your reviews. 🙏🏻

PR #94, opened 2018-09-05
fix: Adjust gist link address bar

BeniCheni:
Fixes: #85

Use a single `<input>` for the gist link address bar:
* use `Input` with `type=url`
* remove `placeholder` to leverage on `value` for gist link
* update left padding style to look "_okay_"
* update related test snapshot.

(P.S.: These changes are based on my best "_guesses_". Happy to update according to review & advice. 🙏🏻)

##### Test
test

##### Test w. token population
test-w-token


felixrieseberg:
Hey! Solid start, but there are a few issues here: 

1) Typing results in the url being repeated over and over again
2) Would be nice to style the left side of the input box, too
3) The existing box has some nice url coercion that might break.

This really is a solid start though!

felixrieseberg:
Hey @BeniCheni, good stuff! I found a few more small issues and decided to just fix them on your branch. I'm now just waiting to see if I broke any tests - if I didn't, I'll merge this later today!

PR #96, opened 2018-09-09
process hook

jeremy-coleman:
exit the electron process on window close

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -5,7 +5,8 @@ const package = require('./package.json');
 
 module.exports = {
   hooks: {
-    generateAssets: require('./tools/generateAssets')
+    generateAssets: require('./tools/generateAssets'),
+    postStart: async ({}, child) => {child.on('exit', () => process.exit(0))}
   },
   packagerConfig: {
     name: 'Electron Fiddle',


malept:
Thanks for filing a pull request! This is actually [a bug in Electron Forge](https://github.com/electron-userland/electron-forge/issues/545). It would be great if you could help us solve it.

jeremy-coleman:
i mean, that fixes it no problem. I wouldn't so much say it's a bug, just the cli API is not standard usage for process / event hooks. 

also the fact that you have to have
  "config": {
    "forge": "./forge.config.js"
  }, is pretty annoying .. without it , forge won't find electron.. a pretty annoying ass problem for something so simple to solve

anyway,

```js
module.exports = {
  hooks: {
    generateAssets: require('./tools/generateAssets'),
    postStart: async ({}, child) => {child.on('exit', () => process.exit(0))}
  },
rest of config..
,
```
that works just fine, it's just that the child object returned comes second in the hook, which is totally ass backwards so i see how people might think it's a bug or hard to figure out, which is why i made the pull req to save some headaches

felixrieseberg:
@jeremy-coleman Thank you for your fix, I really appreciate it. I've closed this PR simply because we had to actually fix this issue in our dependencies 🙇

PR #97, opened 2018-09-10
fix: don't hang on window close when run via npm start

malept:
Fixed via Electron Forge v6.0.0-beta.28 (https://github.com/electron-userland/electron-forge/pull/567)

Closes #96.


felixrieseberg:
Thank you!

PR #101, opened 2018-09-26
fix: don't pin download link to 0.1 tag

ckerr:
The main page's

**"Download Electron Fiddle for macOS, Windows, and Linux"**

link still points to 0.1.

I was showing Fiddle to an Electron user today and we were both confused for a bit why we couldn't find 0.2 :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ GitHub, anyone can quickly try your Fiddle out by just entering it in the
 address bar.
 
 <h3 align="center">
-  <a href="https://github.com/electron/fiddle/releases/tag/0.1">
+  <a href="https://github.com/electron/fiddle/releases/">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ GitHub, anyone can quickly try your Fiddle out by just entering it in the
 address bar.
 
 <h3 align="center">
-  <a href="https://github.com/electron/fiddle/releases/tag/0.1">
+  <a href="https://github.com/electron/fiddle/releases/">
```
MarshallOfSound: There's a magic link we can use here:

`https://github.com/electron/fiddle/releases/latest`

felixrieseberg: Ohh, nice!

PR #104, opened 2018-10-07
feat: provide error message for Token dialog

HashimotoYT:
Added error message in the Token dialog.

## Before

before1

## After

after1


felixrieseberg:
Nice, thank you for the added feature!

I'm wondering though - why did you delete all the templates? Was that just a mistake?

HashimotoYT:
@felixrieseberg, This folder deleted after i'm clone repo. All files from template folder added to staged files and it cannot be reverted. Also when i'm trying to unstage git changes, it change case sensitivity how this view below.

gifca

HashimotoYT:
@felixrieseberg, all templates reverted.

felixrieseberg:
@HashimotoYT: Huh, interesting. Thanks for dealing with this for now, I'll try to look into what's happening later.

One more thing: We have a credits page and I'd love to include you! If you're okay with being on said credits page, would you mind adding yourself to this list?

https://github.com/electron/fiddle/blob/master/tools/contributors.js#L10-L25

PR #106, opened 2018-10-11
feat: Add key property for buttons, update electron and typescript

HashimotoYT:
List of changes:

- Added key property for buttons in token dialog
- Updated Electron to 3.0.4 and TypeScript to 3.1.2
- Added myself to contributors page.


felixrieseberg:
Good stuff, thank you!

zeke:
🙌

PR #111, opened 2018-11-09
fix: friendlier button text for intro tour

nornagon:
In general I like it when buttons are labeled with what's going to happen when I click on them


nornagon:
Remaining CI failures seem unrelated to this PR, but let me know if there's anything i can do to help!

PR #118, opened 2018-12-11
docs: New icon

### LINKED ISSUE #117: New icon
> @felixrieseberg ... an icon that'll allow people to distinguish the default Electron app from Electron Fiddle

simurai:
This PR replaces the current icon with a new version:

### README

Electron Fiddle icon

Electron Fiddle screenshot

### App icon

app icon

### Loading Gif

loading

----

The following 👇  is **not** part of this PR, but _could_ be used in the future if needed, like for Twitter or so.

Electron Fiddle avatar

Electron Fiddle promo

---

Closes #117


malept:
If we add an SVG version of the icon, we can use that for the Linux app (e.g., on [Debian/Ubuntu](https://github.com/electron-userland/electron-installer-debian/blob/v1.0.0/README.md#optionsicon)).

simurai:
> If we add an SVG version of the icon, we can use that for the Linux app (e.g., on Debian/Ubuntu).

Added an SVG version here: https://github.com/electron/fiddle/blob/new-icon/assets/icons/fiddle.svg

I assume it will need to be reference somewhere?

malept:
> > If we add an SVG version of the icon, we can use that for the Linux app (e.g., on Debian/Ubuntu).
> 
> Added an SVG version here: https://github.com/electron/fiddle/blob/new-icon/assets/icons/fiddle.svg
> 
> I assume it will need to be reference somewhere?

Yep, I can add that to this PR.

PR #119, opened 2018-12-12
chore: update windows installer icon

malept:
Now that #118 is merged, there's a stable URL for the new icon that `electron-winstaller` can use.


felixrieseberg:
Thank you!

PR #120, opened 2018-12-17
feat: Minor redesign, new component library

felixrieseberg:
This fairly big PR contains a minor redesign. 

It switches Fiddle from a homegrown and modified version of [Miligram](https://milligram.io/) to Blueprint.js. The former is great, but the latter is just much better suited to our use case – I want Fiddle to be an effective tool for people getting started or working with Electron and good UI components are an important building step. Having some experience with Blueprint, I'm really impressed with the code quality behind the library.

It also:
• Updates parcel and improves the build
• Refactors a bunch of components

screen shot 2018-12-12 at 6 21 53 pm

screen shot 2018-12-13 at 4 19 51 pm

screen shot 2018-12-12 at 6 21 59 pm


forge.config.js
```diff
@@ -1,8 +1,9 @@
 /* tslint:disable */
 
-const path = require('path');
-const package = require('./package.json');
+const path = require('path')
+const packackeJson = require('./package.json')
```
malept: Interesting rename.

forge.config.js
```diff
@@ -1,8 +1,9 @@
...
 
+const { arch, version } = packackeJson;
```
malept: How does one get `arch` from `package.json`?

forge.config.js
felixrieseberg: `package` is a reserved word in strict mode and I should have never used it 😢

forge.config.js
felixrieseberg: It doesn't, this was a bad copy/paste job from another internal app of mine 😅

forge.config.js
```diff
@@ -45,7 +46,7 @@ module.exports = {
         loadingGif: './assets/loading.gif',
         noMsi: true,
         remoteReleases: '',
-        setupExe: `electron-fiddle-${package.version}-setup-${process.arch}.exe`,
+        setupExe: `electron-fiddle-${version}-setup.exe`,
```
malept: Are we not going to distribute setup files for multiple Windows arches, then?

felixrieseberg: We are, I'll just have to handle this manually for now. Is there a way to get whatever architecture `forge` is actually using?

I'll merge this for now, but I'd love to do something like this:

```
import { args } from '@electron-forge/cli'
const { arch } = args
```

malept: I think it'd have to be a [`postMake` hook](https://v6.electronforge.io/configuration#postmake) to rename the artifact with the correct arch. I'd verify with Sam but he's :palm_tree: :smile:

MarshallOfSound: I only muted `electron/electron` 😄 I'd go crazy if I didn't use GH at all for 5 weeks 😆 

You're looking for this hack 😄 Ignore the comment, I'm probably gonna just make it official API 👍 

https://github.com/electron-userland/electron-forge/blob/master/packages/maker/base/src/Maker.ts#L60-L68

Basically, your `config` property can be function instead that takes in the target arch and returns the config object

MarshallOfSound: I'll probably end up doing something similar to the `fromBuildIdentifier` helper --> https://v6.electronforge.io/configuration#build-identifier

PR #125, opened 2018-12-18
chore: automatically add arch to squirrel installer filename

malept:
Per https://github.com/electron/fiddle/pull/120#discussion_r242381413 _(view without whitespace changes to see a better diff)_

If Sam ends up doing https://github.com/electron/fiddle/pull/120#discussion_r242381665 we can change when we upgrade Forge.

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -1,9 +1,9 @@
 /* tslint:disable */
 
 const path = require('path')
-const packackeJson = require('./package.json')
+const packageJson = require('./package.json')
 
-const { version } = packackeJson;
+const { version } = packageJson;
 const iconDir = path.resolve(__dirname, 'assets', 'icons');
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


forge.config.js
```diff
@@ -1,85 +1,87 @@
...
-      name: '@electron-forge/maker-deb',
-      platforms: ['linux'],
-      config: {
-        icon: {
-          scalable: path.resolve(iconDir, 'fiddle.svg')
+    makers: [
+      {
+        name: '@electron-forge/maker-squirrel',
+        platforms: ['win32'],
+        config: {
```
MarshallOfSound: It's only this line that has to be a function, not the whole config afaik

malept: Aha, I'll fix that.

malept: Force-pushed the change: https://github.com/electron/fiddle/commit/0d0a51a900e4f2dafac40e70533e797d20d9d7c3

PR #126, opened 2018-12-18
feat: Use npm registry to fetch releases

felixrieseberg:
We've used the GitHub API in the past to fetch releases. That wasn't as clever as I thought:

* We sometimes have different releases in `electron/electron/releases` than aren't actually on `npm` ( `nightly` support broke because of that)
* GitHub's API sends us way more stuff than we want, forcing us to paginate.

This PR switches over to the npm registry. Bonus points: We no longer need to ask the user how many pages they want to fetch. We'll just make one call.


malept:
FYI, it sounds like there's going to be a separate package for nightlies soon (https://github.com/electron/electron/pull/15938), hopefully this code can be future-proofed for that (if it's not already, I haven't checked).

felixrieseberg:
@malept It doesn't handle that, but none of Fiddle does right now. For reference, we'd have to update `electron-download` and a bunch of other dependencies, too. Presumably, `forge` would have to learn to accept it, too.

So in that sense, this PR doesn't cover this case, but covering it will involve a whole lot more :D

PR #132, opened 2018-12-21
feat: prompt users to move to application folder

codebytere:
This PR asks users to move the app to the application folder on the first time they open it if they have not already done so.

/cc @felixrieseberg

### DIFF
--- a/src/utils/check-first-run.ts
+++ b/src/utils/check-first-run.ts
@@ -1,0 +1,30 @@
+import { app } from 'electron';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+const getConfigPath = () => {
+  const userDataPath = app.getPath('userData');
+  return path.join(userDataPath, 'FirstRun', 'electron-app-first-run');
+};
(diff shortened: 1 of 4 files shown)


src/utils/check-first-run.ts
```diff
@@ -0,0 +1,30 @@
+const fs = require('fs');
```
felixrieseberg:
```suggestion
import * as fs from 'fs-extra';
```

src/utils/check-first-run.ts
```diff
@@ -0,0 +1,30 @@
...
+const path = require('path');
```
felixrieseberg:
```suggestion
import * as path from 'path';
```

src/utils/check-first-run.ts
```diff
@@ -0,0 +1,30 @@
...
+const { app } = require('electron');
```
felixrieseberg:
```suggestion
import { app } from 'electron';
```

PR #135, opened 2018-12-24
ci: Allow Travis to create signed macOS builds

felixrieseberg:
In a perfect world, we could release Fiddle with nothing but CI builds. As of today, I'm still signing things by hand on my local machines, but I'd like to get away from that.

In this PR, I'm enabling Travis to create signed macOS builds.

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -3,7 +3,7 @@
 const path = require('path')
 const packageJson = require('./package.json')
 
- const { version } = packageJson
+const { version } = packageJson
 const iconDir = path.resolve(__dirname, 'assets', 'icons')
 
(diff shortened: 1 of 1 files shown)


forge.config.js
```diff
@@ -4,7 +4,7 @@ const path = require('path')
 const packageJson = require('./package.json')
 
 const { version } = packageJson;
```
malept: For consistency:
```suggestion
 const { version } = packageJson
```

felixrieseberg:
@malept Maybe eventually, but at least for now, I think notarizing is still optional. That said, getting it into `electron-forge` would be amazing.

PR #154, opened 2019-01-08
Add instructions to get Notifications working on Windows 10

### LINKED ISSUE #153: Notifications API don't work on Windows 10
I noticed that we need to add the electron.exe to the start menu. And add `app.setAppUserModelId(process.execPath)` to the Main Process. Creating a PR to the Readme to save other people's sanity & time.
image

vviikk:
Fixes #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,13 @@ fiddle has grown up, export it as a project with or without
 [electron-forge][electron-forge]. Then, use your favorite editor and take on
 the world!
 
+## Troubleshooting
+### Notifications not showing on Windows 10 Fall Update and Newer
+This is because Windows is only allowing known applications to fire notifications.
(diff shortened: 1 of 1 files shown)


malept:
I think it would be better to add some code and/or comments to https://github.com/electron/fiddle/blob/master/static/templates/Notification/main.js rather than increase the size of the README.

Also, doesn't this duplicate the [Notification documentation](https://electronjs.org/docs/tutorial/notifications) (which is linked to in the notification template)?

felixrieseberg:
I agree with @malept here, I think the `README` of this project isn't the right place for this kind of documentation. I really appreciate your effort and am thankful, but I think we should think this over and possibly find another solution.

vviikk:
No problem. I've added this to the Electron docs as well, and the PR just got merged.

https://github.com/electron/electron/pull/16320

Let me know if you need anything else. And yes, we probably need to find a workaround, or do something in electron-forge.

PR #163, opened 2019-01-17
chore: update Electron Forge to 6.0.0-beta.32

malept:
Forge v6 beta 32 has an updated Electron Packager and `electron-installer-*` modules. Offhand, there are two things that are relevant to Fiddle:

* Linux packages no longer include a dependency on `GConf` (dropped in Electron 3)
* macOS notarize support


felixrieseberg:
Thank you!

Issue #76, opened 2018-08-18
"Show Me" not working on Windows


neighthan:
I just downloaded the executable from [here][exe]. Running it pops up the blank Electron Fiddle window. When I select any of the options from the "Show Me" menu, though, (e.g. Show Me > Electron APIs > App), nothing happens. I assumed this would pull up some simple demo project that showcases the named feature? Are there any other setup steps that I need to take besides just running the executable?

Actually, it might be more than just "Show Me" - I tried File > New Fiddle, and I'm still left with the same blank window (just the title bar with, e.g., the button to close the window and the tool bar with File, etc.).

This is on Windows 10. I'm happy to provide any additional information that might be helpful / test things.

[exe]: https://github.com/electron/fiddle/releases/download/0.1/electron-fiddle-0.1.0-setup-win32-x64.exe

TekSiDoT:
Reproducible in latest Win10.

I took a look at the respective source code and suspect [this __Dirname call](https://github.com/electron/fiddle/blob/b074071333d1fa775dbbfc46e5df3c5fa50ff72c/src/renderer/templates.ts#L19); I remember webpack having similar issues with production Electron applications.

However, building with `npm run make` on Windows with the latest code version works just fine (both the resulting executable as well as the squirrel installer), so I guess this might be a result of building the Win32 executable on a non Win32 machine?

felixrieseberg:
Interestingly enough, I spent a lot of time in Windows 10 and don't think I understand what could break. If you open the developer tools, do you see anything in the console?

TekSiDoT:
I took a closer look at the problem. While the dev console itself does not offer any hints on whats happening, [getFile](https://github.com/electron/fiddle/blob/38b7b08691fc0988cdc4416bcaf694b7a1db5e50/src/renderer/templates.ts#L21) in [templates.ts](https://github.com/electron/fiddle/blob/master/src/renderer/templates.ts) catches the following error:

```
"Error: ENOENT, static\templates\DesktopCapturer\renderer.js not found in C:\Users\hab\AppData\Local\electron-fiddle\app-0.1.0\resources\app.asar
    at notFoundError (ELECTRON_ASAR.js:115:19)
    at fs.readFile (ELECTRON_ASAR.js:518:16)
    at go$readFile (C:\Users\hab\AppData\Local\electron-fiddle\app-0.1.0\resources\app.asar\node_modules\graceful-fs\graceful-fs.js:73:14)
[...]

PR #166, opened 2019-01-23
chore: Bump version to 0.5.0

felixrieseberg:


malept:
I wonder if https://github.com/electron/fiddle/issues/76 should be reinvestigated before 0.5.0.

felixrieseberg:
@malept Fair point, I'll take a look today.

Issue #158, opened 2019-01-10
Option to only list downloaded versions of Electron in the dropdown


miniak:
The list of Electron versions to the left of the run button shows way too many items. Make it possible to only display the downloaded versions, which would keep the list small. New versions can always be downloaded in the settings.

malept:
I think if we were to do that, there should be a link to get to that settings page from the dropdown.

felixrieseberg:
Alternative solution: I'll keep the default as it is, but we'll offer this configuration from the settings.

PR #170, opened 2019-01-27
feat: Display only Electron versions with a certain state

felixrieseberg:
This implements #158. In short, it allows users to specify whether or not to display versions that have been downloaded, are downloading, or haven't been downloaded yet.


felixrieseberg:
Coverage actually rose, that piece of information here is deceptive 🗡

PR #183, opened 2019-03-31
feat: Panel management (close and open editors)

felixrieseberg:
This PR is in support of future plans for Fiddle to be more helpful than it is today - for which I'd like to use more panels. Currently, you can't really do anything with panels - short of dragging them around and resizing them.

This PR adds proper _panel management_ to Fiddle. In short, it'll allow users to enable and disable panels! It's a pretty cool feature that'll enable all kinds of neat stuff.

This PR is fairly big but I wanted to make sure that I'm implementing this _properly_.

panels


felixrieseberg:
:rocket: Here we go!

PR #195, opened 2019-05-09
Fix: Require not defined bug

ajphukan:
**Issue:**  Uncaught ReferenceError: require is not defined  
**Files Affected:** All the Show Me example file  which uses './renderer.js'  
**Solution:** Can be solved by adding the webPreferences in main.js file.
`   webPreferences: {
      nodeIntegration: true
    }`


felixrieseberg:
Thank you!

PR #203, opened 2019-05-27
feat: add support for launch examples from the Electron docs

MarshallOfSound:
This adds support for loading fiddles directly from the Electron repository.  E.g. 

electron-fiddle://electron/fiddle-in-docs/docs/fiddles/single-instance

URL is of the format `electron-fiddle://electron/{ref}/{path}`

This will let us write fiddles as examples in our docs and let users one-click launch them to try them out


src/renderer/components/commands-address-bar.tsx
```diff
@@ -96,13 +121,85 @@ export class AddressBar extends React.Component<AddressBarProps, AddressBarState
...
+      if (appState.gitHubToken) {
+        octo.authenticate({
+          type: 'token',
+          token: appState.gitHubToken!
+        });
+      }
+
+      // TODO: Maybe we should fetch the package.json for the given ref to use the correct version of Electron?
+
+      const folder = await octo.repos.getContents({
```
felixrieseberg: I wonder if we can even hit this endpoint more than once without being authenticated 🤔

felixrieseberg:
This is very, very cool!

src/renderer/components/commands-address-bar.tsx
MarshallOfSound: If we're ok with making more requests and relying on 404's I think we can make this work without using the API (just hitting the `raw.*` URLs and guessing for `main.js`, `renderer.js` and `index.html`

Issue #188, opened 2019-04-11
Add an about dialog


malept:
As far as I can tell, there's no way to figure out what version of Electron Fiddle a user is running without asking them to dig into the source code of the bundled app, or if they're on Linux, using their package manager. Including this will make it easier to troubleshoot GitHub issues.

Saoskia:
What OS are you using? I'm on a Mac, and there already exists an About dialog by clicking on the app name and selecting 'About Electron Fiddle'. I don't have access to other OS to check them.

![Screen Shot 2019-05-27 1](https://user-images.githubusercontent.com/18042873/58420806-a36f2080-80d1-11e9-852a-3ab82ec0d467.png)
![Screen Shot 2019-05-27 2](https://user-images.githubusercontent.com/18042873/58420814-a833d480-80d1-11e9-8974-64740dbc0af6.png)

codebytere:
@Saoskia the issue is that it's not showing what version of _Electron_ is used in a given version of Fiddle, not what version of _Fiddle itself _is being used!

Saoskia:
@codebytere My apologises, that's not what I understood the issue to be.

PR #204, opened 2019-06-01
feat: add more information to about panel

codebytere:
Resolves https://github.com/electron/fiddle/issues/188.

Adds more information to the About panel on macOS and Linux, including the Electron version being used. @felixrieseberg I wasn't sure what you'd like to put for copyright, so i put `Electron Authors` for now but if there's something else you'd prefer I can use that instead!

new panel on macOS:

Screen Shot 2019-05-31 at 10 29 58 PM

### DIFF
--- a/src/utils/set-about-panel.ts
+++ b/src/utils/set-about-panel.ts
@@ -1,0 +1,26 @@
+import { AboutPanelOptionsOptions, app } from 'electron';
+
+/**
+ * Sets Fiddle's About panel options on Linux and macOS
+ *
+ * @returns
+ */
+export function setupAboutPanel(): void {
+  const options: AboutPanelOptionsOptions = {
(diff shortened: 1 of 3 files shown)


src/utils/set-about-panel.ts
```diff
@@ -0,0 +1,26 @@
+import { AboutPanelOptionsOptions, app } from 'electron';
```
malept: It's really called `AboutPanelOptionsOptions`? :laughing:

MarshallOfSound: That's um, questionable on my part 😆 The TS method takes `${methodName}${paramName}` as the default.  In this case the method is `setAboutPanelOptions` and the param is `options` 🤔 Maybe it should prevent `OptionsOptions` 🤔

PR #207, opened 2019-06-07
feat: handle incoming electron-fiddle://token/* launches

MarshallOfSound:
This will be paired with a new service to do an oauth dance and provide fiddle with this token.  This will be the new user flow way easier (no manual process for creating and setting up the Personal Access Token)


MarshallOfSound:
Just adding the ability to do this launch for now, at some point in the future we should update `dialog-token.tsx` to instead ask people to "Sign In" and open the token service.

PR #209, opened 2019-06-11
Fix: submit token

amorist:
Fixed:
1. Please provide a valid GitHub Personal Access Token
2. Can't perform a React state update on an unmounted component.


felixrieseberg:
Test error looks like a flake. Thank you very much for the fix!

PR #211, opened 2019-06-11
fix: Don't attempt to set about panel on Windows

### LINKED ISSUE #210: File menu regression on Windows
All menu items except for Exit are missing in v0.8.0 on Windows

Fiddle v0.7.0
Screen Shot 2019-06-11 at 12 46 09 PM

Fiddle v0.8.0
Screen Shot 2019-06-11 at 12 46 28 PM

felixrieseberg:
Turns out, `setAboutPanelOptions` doesn't work on Windows, so we're crashing there 🙈 

fixes #210

### DIFF
--- a/src/utils/set-about-panel.ts
+++ b/src/utils/set-about-panel.ts
@@ -6,6 +6,8 @@ import { AboutPanelOptionsOptions, app } from 'electron';
  * @returns
  */
 export function setupAboutPanel(): void {
+  if (process.platform === 'win32') return;
+
   const options: AboutPanelOptionsOptions = {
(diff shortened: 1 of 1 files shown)


miniak:
what about checking if `setAboutPanelOptions` exists instead of the platform check?

felixrieseberg:
> what about checking if setAboutPanelOptions exists instead of the platform check?

Even if it does, it's not work we _want_ to do on Windows. I'm a bit `/shrug` on this, but in the interest of getting a fix out, I think this is fundamentally the same. It's not like one solution is slower or faster than the other

PR #214, opened 2019-06-12
snap

mikeroyal:
Includes a wrapper script and snapcraft.yaml to build [snap](https://snapcraft.io/).


malept:
The wrapper script & hand-coded YAML isn't necessary. Fiddle uses Electron Forge, which has built-in support for creating snaps for Electron apps (like Fiddle) if the config is set.

That being said, @felixrieseberg do you want to add this app to the Snap Store?

mikeroyal:
Hi @malept, thank you for update. :+1:

PR #179, opened 2019-03-22
Welcome Tour - formatting bugs (spacing)

marclundgren:
I loaded the welcome tour up to find a few places where a `code` block was missing a space between itself and the proceeding word. See the screenshots below for all 3 places I saw the problem:

## before
electron fiddle tutorial - main script - formatting bug
Screen Shot 2019-03-22 at 2 22 46 PM
Screen Shot 2019-03-22 at 2 25 27 PM

## after
Screen Shot 2019-03-22 at 2 32 49 PM
Screen Shot 2019-03-22 at 2 32 54 PM
Screen Shot 2019-03-22 at 2 33 01 PM

PR #216, opened 2019-06-14
adds missing spaces

drewswaycool:
Adds spaces to the tour experience where they were missing.

_another option would be to add `{' '}` directly if desired; this could be done in place of the spaces added or on the line above as needed._

issue that this resolves:
Screen Shot 2019-06-14 at 3 43 40 PM


malept:
Thanks for the contribution, but it seems like a duplicate of #179? Perhaps if you update the tests as well, we can get this one merged instead.

felixrieseberg:
I'll merge this, the test failure is caused by something else (which has been fixed on `master` on the meantime). Thank you!

PR #218, opened 2019-06-16
feat: add ability to clear console output

### LINKED ISSUE #196: Feature Request: Add ability to clear console output
It would be really useful to allow clearing the console output by:
- adding a button to the toolbar (next to the "Console" button)
- adding a context menu item (below "Run Fiddle" for example)

DeerMichel:
**What's Included**
- context menu item to clear console below "Run Fiddle" (Closes #196)
- top bar button to clear console next to "Console" (#196)
- setting to automatically clear console when fiddle is executed

**Discussion**
- I dont like the top bar button. Maybe a setting to optionally en/disable it would be fine - or do we need it at all since there is an auto-clear-on-run function? And if we keep it, should there be a Touch Bar button too?

Hope I got everything right and extended the tests properly :)


src/renderer/runner.ts
```diff
@@ -42,6 +42,8 @@ export class Runner {
     const { binaryManager, currentElectronVersion } = this.appState;
     const { version, localPath } = currentElectronVersion;
 
+    if (this.appState.isClearingConsoleOnRun)
+      this.appState.clearConsole();
```
felixrieseberg: Dumb nit: I'd love some `{ }` around the if. We're in the process of adding `prettier` to avoid stupid style comments, but in the meantime, apologies for the nit 👯‍♂

felixrieseberg:
> I dont like the top bar button. Maybe a setting to optionally en/disable it would be fine - or do we need it at all since there is an auto-clear-on-run function? And if we keep it, should there be a Touch Bar button too?

I agree with you. We could add a button to the console itself (maybe the top right?). That said, I think I'd intuitively open the context menu in the console to clear it.

DeerMichel:
Removed the button for now. If someone has a need for it, go for it.

Since i dont have write access for this repo, feel free to merge at any time ;)

codebytere:
@felixrieseberg this is oddly failing with: 

```
Contributors array is empty
Error: Contributors array is empty
    at /Users/travis/build/electron/fiddle/tools/contributors.js:147:17
    at processTicksAndRejections (internal/process/task_queues.js:89:5)
```

even though [this diff](https://github.com/electron/fiddle/pull/218/files#diff-c9ca73536ff0a746f7c7d8e5a4d137fe) shows that @DeerMichel added himself; any ideas what might be up?

felixrieseberg:
Losing the 4.5% in coverage is a bit much, but I'll add some tests tomorrow. For now, I'll merge this.

As for the contributors error: That's because the build isn't trusted, therefore no GITHUB_TOKEN is set, and GitHub is rate-limiting Travis CI.

PR #220, opened 2019-06-27
feat: clarify UX for Publish Gist feature

erickzhao:
Heya 👋 

I've been sharing Electron example apps from Fiddle via GitHub Gists over the past few weeks. I noticed a few UX improvements that could be made and packaged them into this PR. These are subjective, so please provide feedback on whether they'd be useful!

### 1. Clearer visual feedback when publishing Gist

Disabling the address bar during the publish action might be helpful, as you generally don't want to load another gist during the process of publishing one.
screencast 2019-06-27 11-34-56

Moreover, this fixes a visual inconsistency where the privacy toggle button wouldn't be `disabled` on `isPublishing`. This was because the `disabled` option in the [`Popover`](https://blueprintjs.com/docs/#core/components/popover) API from Blueprint only prevents the popover menu from being opened, but the button would still appear as clickable.

_Note that this proposed change moves the `isPublishing` flag from the component state in `commands-publish-button` to the global store so that it can be shared with `commands-address-bar`._

**Also, should I be adding a spec for this change?** 

### 2. Consistent language for publishing action

This is a nitpick, but it might be clearer to users if "publish" was the action verb across the app for uploading a Fiddle to Gist. Previously, the language was "Save to Gist" in the Menu, "Publish" on the commands bar, and "pushed" in the README.

new

### 3. Correction of callout in settings menu

Another nitpick, but the callout in the Gist settings previously only mentioned that you could publish **public** Gists. I think it'd be clearer to remove that word from the settings.


felixrieseberg:
Great improvements, thank you very much!

> Also, should I be adding a spec for this change?

Yes please!

erickzhao:
Note that the latest update to this PR updates `enzyme` from `^3.9.0` to `^3.10.0`. There was a bug in `@types/enzyme` that prevented me from using an argument in `ShallowWrapper.setProps()`. The fix was only pushed on the `3.10.0` version of that package, so I updated both the types and the main package.

See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18215

PR #222, opened 2019-07-03
fix: correct button code for BrowserWindow show-me

erickzhao:
Little fix so that the button in the `BrowserWindow` show-me does what the code seemed like it was supposed to do. Upon clicking the `Close this window` button, the window will close! ✨ 
screencast 2019-07-03 15-54-31


felixrieseberg:
Thank you!

PR #227, opened 2019-07-12
chore: add offline notice when loading fiddles

DeerMichel:
This PR adds a more verbose warning message when trying to load a fiddle while computer is offline.


DeerMichel:
@MarshallOfSound yeah I'm now thinking about catching 3 cases manually and display only our own error message for those: offline, not found and GitHub down (if distinguishable from offline case). I would only show the underlying error, if it doesn't fall under one of this cases.

PR #229, opened 2019-07-12
feat: add dev mode support for protocol

erickzhao:
Adds support for dev mode for the Protocol Launch feature. Allows you to add the protocol URL as an arg to your `npm start command`.

Example:
```
npm start -- electron-fiddle://electron/fiddle-in-docs/docs/fiddles/single-instance
```

### DIFF
--- a/src/main/protocol.ts
+++ b/src/main/protocol.ts
@@ -68,7 +68,7 @@ export const scanArgv = (argv: Array<string>) => {
 
 export const scanNpmArgv = (argv: string) => {
   const parsedArgv = JSON.parse(argv);
-  const { remain: args } = parsedArgv;
+  const { original: args } = parsedArgv;
   scanArgv(args);
 }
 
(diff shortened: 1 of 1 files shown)


src/main/protocol.ts
```diff
@@ -75,7 +82,10 @@ export const listenForProtocolHandler = () => {
     }
   });
 
-  if (process.platform === 'win32') {
+  // pass protocol URL via npm start args in dev mode
+  if (isDevMode() && process.env.npm_config_argv) {
+    scanNpmArgv(process.env.npm_config_argv);
+  } else if (process.platform === 'win32') {
```
MarshallOfSound: This probably shouldn't be an else if.

These conditions should both be able to happen at the same time

erickzhao: Does this mean that if we're running `win32` and we're in dev mode, we want `handlePotentialProtocolLaunch` to run twice? It seems to me like duplicate code, but on the other hand, running `scanArgv` twice doesn't seem to have any negative side effects (tested on my end by just duplicating the `scanNpmArgv` function call).

felixrieseberg: I think we do just want to run this once.

PR #232, opened 2019-07-15
feat: force electron version when loading from example protocol

erickzhao:
### Changes
- ✨When loading from example protocol, switches Electron versions depending on the version specified in `package.json` in the protocol ref
  - If version in disabled release channel, prompt user to enable the release channel before proceeding
  - If version unsupported, throw error dialog
- ♻️Encapsulates all remote loading logic to its own `RemoteLoader` class.
- ✅ Adds tests for everything!

### Testing
Working version (`5.0.6`):
```
npm start -- electron-fiddle://electron/fiddle-protocol/docs/fiddles/single-instance
```
Unsupported version (old nightly):
```
npm start -- electron-fiddle://electron/fiddle-in-docs/docs/fiddles/single-instance    
```

### Discussion
Sorry for the large PR!

cc @felixrieseberg @MarshallOfSound


src/renderer/remote-loader.ts
```diff
@@ -0,0 +1,231 @@
...
+import { INDEX_HTML_NAME, MAIN_JS_NAME, RENDERER_JS_NAME } from '../shared-constants';
+import { getTitle } from '../utils/get-title';
+import { getOctokit } from '../utils/octokit';
+import { sortedElectronMap } from '../utils/sorted-electron-map';
+import { ELECTRON_ORG, ELECTRON_REPO } from './constants';
+import { getContent } from './content';
+import { AppState } from './state';
+import { ElectronReleaseChannel, getReleaseChannel } from './versions';
+
+export class RemoteLoader {
```
felixrieseberg: This is right and good

PR #233, opened 2019-07-17
chore: `toogle` -> `toggle`

erickzhao:
Tiny nit where we were calling a bunch of variables `toogle` rather than `toggle` in the code. Will make toggles easier to search for 👀


erickzhao:
Seems like `continuous-integration/appveyor/branch` is failing. Looks like a flake to me (failing on install), but I don't have the most context on this!

felixrieseberg:
Surely a flake, let’s merge this 👍

PR #247, opened 2019-08-12
feat: intellisense for local builds

erickzhao:
This PR adds IntelliSense for local versions of Electron by loading and watching the `electron.d.ts` file generated by the local build. By using `fs.watch`, the definitions will be renewed whenever they are updated locally.

Since the TypeScript definitions are generated from documentation, changes to the IntelliSense will only be reflected upon modifying the documentation locally and rebuilding.


felixrieseberg:
This is pretty impressive stuff.

PR #253, opened 2019-09-13
fix: ensure state consistency when loading an external Fiddle

### LINKED ISSUE #251: Cancelling overwrite when opening new Fiddle does not cancel local file path change
Ok, first time posting an issue on GitHub so bear with me.

Lets say you want to switch fiddle from Fiddle1 to Fiddle2 but you have unsaved changes on Fiddle1.
You will get a dialog box asking:
"Opening this fiddle will replace your unsaved changes. Do you want to proceed?"
With options: Cancel and Yes

If you cancel the dialog box the code will stay the same (Fiddle1) but the path will change to the project you were about to open (Fiddle2).
So if you Ctrl-S at this point you will overwrite the code in Fiddle2 with the code of Fiddle1.

- - -

Fiddle version: 0.9.0
Windows 10

erickzhao:
Fixes #251 by refactoring the prompt logic.

### Context
Previously, ancillary state variables for a Fiddle (`gistId`, `filePath`, `templateName`) were being reset even if a user cancelled the loading via a prompt.

### Solution

_Note: The LoC setting these variables simply could have been moved such that they were blocked by a prompt cancellation. However, I noticed that I had to do this many times in the code, so it seemed like a code smell._

* Renamed functions to make purpose clearer:
  * `app.setValues() -> app.setEditorValues()`
  * `app.getValues() -> app.getEditorValues()`

* Added a new `replaceFiddle()` function:
  * **Purpose:** to have a DRY abstraction for replacing the existing Fiddle with a new one.
  * **Motivation:** Previously, local Fiddles, show-mes, Gists, and Fiddle Protocols each had their own loading implementation. This involved code duplication for calling `setEditorValues()`, showing the Unsaved prompt, and setting supporting state variables.
  * The new implementation gets rid of the `warn` boolean parameter, as we always warn when loading an external Fiddle. `setEditorValues()` is called directly if warning is unnecessary (e.g. when switching versions).
  * **The bug is addressed in here because `replaceFiddle` avoids setting the new Fiddle's supporting state variables if the Unsaved prompt is cancelled.**

Example of new API:
```javascript
// local file
await replaceFiddle(editorValues, { localPath: '/some/path' });
```

### DIFF
--- a/src/renderer/app.tsx
+++ b/src/renderer/app.tsx
@@ -2,11 +2,17 @@ import { when } from 'mobx';
 import * as MonacoType from 'monaco-editor';
 
 import { ipcRenderer } from 'electron';
-import { ALL_EDITORS, EditorId, EditorValues } from '../interfaces';
+import {
+  ALL_EDITORS,
+  EditorId,
+  EditorValues,
+  SetFiddleOptions
+} from '../interfaces';
 import { WEBCONTENTS_READY_FOR_IPC_SIGNAL } from '../ipc-events';
 import { updateEditorLayout } from '../utils/editor-layout';
 import { getEditorValue } from '../utils/editor-value';
 import { getPackageJson, PackageJsonOptions } from '../utils/get-package';
+import { getTitle } from '../utils/get-title';
 import { isEditorBackup } from '../utils/type-checks';
 import { FileManager } from './file-manager';
 import { RemoteLoader } from './remote-loader';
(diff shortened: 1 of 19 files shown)


src/renderer/app.tsx
```diff
@@ -31,36 +37,60 @@ export class App {
   public touchBarManager: TouchBarManager | undefined;
 
   constructor() {
-    this.getValues = this.getValues.bind(this);
-    this.setValues = this.setValues.bind(this);
+    this.getEditorValues = this.getEditorValues.bind(this);
```
felixrieseberg: 👍 I like how this is a lot clearer

src/renderer/file-manager.ts
```diff
@@ -38,13 +37,12 @@ export class FileManager {
   /**
    * Opens a template.
    *
-   * @param {string} name
+   * @param {string} templateName
```
felixrieseberg: Again, solid rename 👍

PR #257, opened 2019-09-26
feat: add Bisect helper

### LINKED ISSUE #244: Introduce a bisect helper
This might be a bit of a big project but I have a _terrible_ local script I use nowadays to effectively "git bisect" but for electron nightly releases.  The idea being to track down when something broke using our new nightly releases.

Chromium has a similar thing --> https://www.chromium.org/developers/bisect-builds-py

If fiddle had a "bisect" helper where you could give a start nightly and an end nightly and go through the bisect logic to narrow down which release broke it that would be _awesome_.

I'm imagining a UI where you choose two nightly releases and then two new "good" and "bad" buttons appear which you click until you've finished bisecting

erickzhao:
Closes #244.

Adds a version bisect helper for Electron Fiddle. Inspired by [git bisect](https://git-scm.com/docs/git-bisect), this feature helps Electron contributors or Electron users reporting a bug to pinpoint the exact version where a bug started occurring.

## Algorithm
Taken directly from the Chromium bisect code:
https://github.com/chromium/chromium/blob/master/tools/bisect-builds.py#L839

## Screenshots
Dialog allowing the selection of a version range.
image

Once range is chosen, immediately jump to pivot version and allow user to continue w/ 3 options: good version/bad version/cancel.
image

## Bonus
This PR also adds graceful edge behavior for the commands bar. It will now scroll horizontally when overflowing.

## Considerations
- Is the experience intuitive?
- Is code re-use between the existing Version selector and the new Bisect dialog optimized?


erickzhao:
Currently wondering if I should be adding this to the Preferences or as a View sub-menu item.

felixrieseberg:
`+1,194 −173 ` Damn son this is some serious work 👏

src/renderer/components/dialog-bisect.tsx
```diff
@@ -0,0 +1,161 @@
+
```
felixrieseberg:
```suggestion
```

src/renderer/state.ts
```diff
@@ -66,47 +67,50 @@ export class AppState {
...
-      this.retrieve('statesToShow') as Array<ElectronVersionState>
-      || [ ElectronVersionState.downloading, ElectronVersionState.ready, ElectronVersionState.unknown ];
+    this.retrieve('statesToShow') as Array<ElectronVersionState>
+    || [ElectronVersionState.downloading, ElectronVersionState.ready, ElectronVersionState.unknown];
   @observable public isKeepingUserDataDirs: boolean = !!this.retrieve('isKeepingUserDataDirs');
   @observable public isEnablingElectronLogging: boolean = !!this.retrieve('isEnablingElectronLogging');
-
-  @observable public binaryManager: BinaryManager = new BinaryManager();
+  @observable public isClearingConsoleOnRun: boolean = !!this.retrieve('isClearingConsoleOnRun');
+  @observable public isBisectCommandVisible: boolean = !!this.retrieve('isBisectCommandVisible');
```
felixrieseberg: You should only need the `retrieve` if you're saving the value to `localStorage` with an autorun later

erickzhao:
@felixrieseberg just went and did a final validation pass and realized that my bisect algorithm needs some tweaks @_@ will rework soon

erickzhao:
@felixrieseberg @MarshallOfSound finally found time to fix this. Old algorithm was very flawed, so I just used Chromium's and it seems to be doing well. Completed Bisect output is now shown in the console.

src/renderer/bisect.ts
```diff
@@ -0,0 +1,58 @@
+import { ElectronVersion } from '../interfaces';
+
+export class BisectInstance {
```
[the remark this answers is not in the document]
erickzhao: That makes sense. I like `Bisector`.

Issue #259, opened 2019-10-09
Missing instructions to build fiddle


arnitkun:
A very trivial case, but the readme does not contain the steps to build fiddle. While this is something almost everyone would be knowing and can be figured out in a couple of minutes, I think somethings that would be of great use to beginners should have all the basic information in one single place. I would be happy to send a pull request for this.

Apologies if this is redundant/unnecessary.

malept:
Fixed in #311.

PR #264, opened 2019-10-11
docs: Added the steps to develop fiddle.

arnitkun:
Added the steps to run fiddle, as mentioned in #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,13 @@ fiddle has grown up, export it as a project with or without
 [electron-forge][electron-forge]. Then, use your favorite editor and take on
 the world!
 
+### Running Fiddle
+
+1. Clone this repository: `git clone https://github.com/electron/fiddle.git`.
+2. Change current directory to fiddle: `cd ./fiddle`
+3. Run `npm install`
+4. Run `npm start`
+
 ## License
 
 [MIT, please see the LICENSE file for full details](https://github.com/electron/fiddle/blob/master/LICENSE.md).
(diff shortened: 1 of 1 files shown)


malept:
Wouldn't the steps be "Download the binary and run"? Installing it from the repo is only if you want to work on Fiddle itself.

arnitkun:
yes, I was only thinking from the point of view of someone who wants to develop himself, but then he/she should know how to do this. 

Adding the instruction for "end-users".

malept:
In that case, the "how to download Fiddle to develop on it" instructions should be in a `CONTRIBUTING.md` file (which I'm kind of surprised that it's not in this repo yet) instead of `README.md`.

README.md
```diff
@@ -56,13 +56,17 @@ fiddle has grown up, export it as a project with or without
...
 
 1. Clone this repository: `git clone https://github.com/electron/fiddle.git`.
 2. Change current directory to fiddle: `cd ./fiddle`
 3. Run `npm install`
 4. Run `npm start`
 
+### Running Fiddle
+
+Download and run the latest binary from the electron [releases page](https://electronjs.org/releases/stable).
+
```
malept: I don't think this is necessary, since there's already a Download link after the first paragraph in the README.

arnitkun: okay, removing this and placing the build instructions in contributing.md. 

One quick question, should I put a generic contribution guide or would one/more of the members do it?

felixrieseberg: I think a contributor's guide is a wonderful idea!

felixrieseberg:
There hasn't been much movement here in a while, so I'm closing this for now. We could always reopen it!

arnitkun:
Yeah, Apologies for that, very busy right now.

PR #267, opened 2019-10-25
feat: Electron 7 (and more updates)

felixrieseberg:
This PR updates us to Electron 7. It also updates all other packages _and_ fixes small typing issues that came with that.


src/renderer/remote-loader.ts
```diff
@@ -174,7 +179,8 @@ export class RemoteLoader {
       path: 'package.json'
     });
 
-    const packageJsonString = Buffer.from(packageJsonData.content, 'base64').toString('utf8');
+    // This is bug in Octokit's typing
+    const packageJsonString = Buffer.from((packageJsonData as any).content, 'base64').toString('utf8');
```
erickzhao: Seems like this is a TypeScript feature rather than an Octokit bug -- it's telling us to guard our types because the Octokit response can generate either a single object or an array of other objects.

```typescript
  type ReposGetContentsResponse =
    | {
        _links: ReposGetContentsResponseLinks;
        content?: string;
        download_url: string | null;
        encoding?: string;
        git_url: string;
        html_url: string;
        name: string;
        path: string;
        sha: string;
        size: number;
        type: string;
        url: string;
        target?: string;
        submodule_git_url?: string;
      }
    | Array<ReposGetContentsResponseItem>;
```

For instance, this snippet makes the code transpile correctly:
```typescript
if (!Array.isArray(packageJsonData) && !!packageJsonData.content) {
      const packageJsonString = Buffer.from(packageJsonData.content, 'base64').toString('utf8');
      const { version } = JSON.parse(packageJsonString);
      return version;
    }
```

felixrieseberg: Oh, good catch, I'll put that in.

PR #271, opened 2019-11-02
fix: update folder selection dialog for Electron 7

### LINKED ISSUE #269: Unable to select folder in local Electron build dialog
picture1:
image
picture2:
image
I can't add local Electron bulid in Electron setting. As you see in picture1 and picture 2 .The `F:\Program-Files\Electron` have included `electron.exe`, but The picture 1 show that `We did not find a electron.exe  in this folder...`. How can I solve this Problem?

erickzhao:
WIP: Need to fix specs and commit to good test coverage.

Fixes #269.

### Context
Our upgrade to Electron 7 inadvertently broke our usage of `<input type="file" webkitdirectory/>`, which is a non-standard browser API designed to allow the file input to select entire directories ([MDN link](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory)).

In Electron <= 6, the returned `files` object returned the selected folder itself as. This was incorrect behaviour, and was fixed in this PR: https://github.com/electron/electron/pull/18343.

Starting in Electron 7, the returned `files` object returns the correct value: a `FileList` of all files. However, we no longer have a direct reference to our selected folder.

### Change
~Instead, there's some jank path manipulation going on, where we grab an arbitrary file, and using its absolute path and relative path from the selected folder, obtain the absolute path of the selected folder.~

No more jank. PR instead leverages Electron's `dialog.showOpenDialog()` API to select the folder and retrieve its path.


MarshallOfSound:
IMO we should just use `dialog.showOpenDialog({ properties: ['openDirectory'] })`.  There isn't a Good ™️ way to do this with web APIs.  Let's use the Electron one designed for this use case.

I.e. Change this to a text input and a button and emulate the behavior using the Electron API

mklnz:
https://github.com/electron/electron/pull/18343#issuecomment-549018047

> So previously I was able to get the path of the directory selected. However with the "fixed" version I'm only able to get a list of files within that directory, how to get the directory path if there are no files inside said directory since the target.files array is empty?
> ```
> files: FileList
>   length: 0
> ```

erickzhao:
@MarshallOfSound hadn't thought of that -- definitely sounds less hacky. I'll redo the code here accordingly.

@mklnz, the above suggestion `dialog.showOpenDialog({ properties: ['openDirectory'] })` would be a good alternative.

src/main/dialogs.ts
```diff
@@ -17,6 +17,10 @@ export function setupDialogs() {
     showConfirmationDialog(args);
   });
 
+  ipcMainManager.on(IpcEvents.SHOW_LOCAL_VERSION_FOLDER_DIALOG, async (_event) => {
+    await showOpenDialog(_event);
```
felixrieseberg: This should probably be `event` if we actually forward it (TypeScript uses the `_` to denote "I don't use this")

miniak:
@erickzhao you could have used the new `ipcRenderer.invoke()` / `ipcMain.handle()` APIs

PR #272, opened 2019-11-03
fix: update dialog example for promisification

codebytere:
Fixes the dialog example - `showOpenDialog` now returns a Promise in supported versions so we should likely be using that. The previous example would not properly load the chosen file. 

cc @felixrieseberg

### DIFF
--- a/static/show-me/dialog/main.js
+++ b/static/show-me/dialog/main.js
@@ -16,7 +16,7 @@ app.on('ready', () => {
     properties: ['openFile']
   }).then(result => {
     if (result.canceled) {
-      console.log('Dialog was cancelled')
+      console.log('Dialog was canceled')
     } else {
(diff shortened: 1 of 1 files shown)


static/show-me/dialog/main.js
```diff
@@ -12,13 +12,16 @@ app.on('ready', () => {
...
-  //
-  // If you provide a callback, the method will be asynchronous,
-  // but in thise case, we'll just use the synchronous variant.
-  const file = dialog.showOpenDialog({
-    title: 'Hello!',
+  dialog.showOpenDialog(mainWindow, {
     properties: ['openFile']
+  }).then(result => {
+    if (result.canceled) {
+      console.log('Dialog was cancelled')
```
erickzhao:
```suggestion
      console.log('Dialog was canceled')
```

PR #275, opened 2019-11-12
fix: call isDevMode in main/devtools.ts

Letra312:
isDevMode is of function type but used as boolean.


Letra312:
@coveralls OK! No problem!

Letra312:
Create another PR #277

PR #277, opened 2019-11-13
fix: ensure dev tools only open in dev mode

### LINKED ISSUE #276: isDevMode should be called in main/devtools.ts
isDevMode is of function type but used as boolean.

Letra312:
Closes #276 

isDevMode is of function type but used as boolean.

Fix devtools-spec.ts for not calling isDevMode

### DIFF
--- a/tests/main/devtools-spec.ts
+++ b/tests/main/devtools-spec.ts
@@ -1,13 +1,5 @@
 import { setupDevTools } from '../../src/main/devtools';
 
-let mockIsDevMode = false;
-
-jest.mock('../../src/utils/devmode', () => ({
-  get isDevMode() {
-    return mockIsDevMode;
-  }
-}));
-
 jest.mock('electron-devtools-installer', () => ({
   default: jest.fn(),
   REACT_DEVELOPER_TOOLS: 'REACT_DEVELOPER_TOOLS',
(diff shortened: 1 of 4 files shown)


erickzhao:
@Letra312 ready for re-review now? :)

Letra312:
@erickzhao Yeah! Just now CI building succeeded!

tests/main/devtools-spec.ts
```diff
@@ -1,34 +1,54 @@
...
+  const old = (process as any).defaultApp; // for tsconfig error
+
+  afterEach(() => {
+    Object.defineProperty(process, 'defaultApp', { value: old });
+  });
+
   it('does not set up developer tools if not in dev mode', () => {
     const devtools = require('electron-devtools-installer');
-
+    Object.defineProperty(process, 'defaultApp', {
```
erickzhao: Instead of setting `process.defaultApp` and calling the actual `isDevMode()` function in the tests, could you use the `jest.Mock()` approach to instead mock the return value of the `isDevMode()` function.

The motivation is that we don't want to test if `isDevMode()` returns the correct value of `process.defaultApp`. Rather, assuming that `isDevMode()` returns a certain results, we want to test how our function works.

See code below:
```ts
import { isDevMode } from '../../src/utils/devmode';
jest.mock('../../src/utils/devmode');
// ...
it('sets up developer tools if in dev mode', () => {
    // ...
    (isDevMode as jest.Mock).mockReturnValue(true); // mock false value for the other test
});
```

tests/main/devtools-spec.ts
```diff
@@ -1,34 +1,54 @@
...
+      writable: true
+    });
     setupDevTools();
 
     expect(devtools.default).toHaveBeenCalledTimes(1);
   });
+
+  it('catch error in setting up developer tools', async (done) => {
+    const devtools = require('electron-devtools-installer');
+    // throw devtool erros
```
erickzhao:
```suggestion
    // throw devtool errors
```

Letra312: OK. I'll fix these bugs right now.

Letra312:
I have another question, code like 
``` javascript
jest.mock('../../src/utils/devmode', () => ({
  isDevMode: jest.fn()
}));
// ...
  it('does not set up developer tools if in dev mode', () => {
    const devtools = require('electron-devtools-installer');
    const { isDevMode } = require('../../src/utils/devmode');
    isDevMode.mockReturnValue(true);

    setupDevTools();
    expect(devtools.default).toHaveBeenCalledTimes(1);
  });
```
cannot pass the test.

Like code above, `isDevMode` is not called in '../../src/utils/devmode'?

Letra312:
Maybe I have known  the differences. Mock one total module instead of `isDevMode` single export.
I'll go to see how it works. :)

PR #281, opened 2019-12-12
fix: Incorrect dropdowns

felixrieseberg:
This code was introduced in #257 and seems to have broken popovers and dropdowns. I don't believe it's actually required, I ran some bisect operations without it and things worked fine.

### Before

Screen Shot 2019-12-12 at 2 24 53 PM

### After

Screen Shot 2019-12-12 at 2 24 32 PM


felixrieseberg:
cc @erickzhao: You have finals, focus on those, but I still wanted you to know 👍

erickzhao:
Can confirm this was unrelated to the Bisect feature itself. I added those lines in because the command bar was cramped with so many elements (once you added the bisect stuff in).

PR #293, opened 2020-01-15
fix: Create initial backup for preload editor

### LINKED ISSUE #284: Preload.js is not loaded nor uploaded
Hi there, 

**What works**
- preload.js is saved when you save or "save as forge" the project
- preload.js is loaded when you downloaded the gist

**What didn't works**
- preload.js is not loaded when you open existing project
- preload.js is not uploaded if you publish to gist

I haven't tried to package or make installer

thank you...

felixrieseberg:
When loading a Fiddle that contains a `preload` file, we didn't correctly load that content unless the user has previously opened the preload editor. The underlying cause was that we didn't create an initial text model for that editor, so we didn't have a place to store that value.

Closes #284

### DIFF
--- a/src/renderer/components/editor.tsx
+++ b/src/renderer/components/editor.tsx
@@ -99,33 +99,52 @@ export class Editor extends React.Component<EditorProps> {
   }
 
   /**
+   * Create a model and attach it to the editor
+   *
+   * @private
+   * @param {string} value
+   */
+  private async createModel(value: string) {
+    const { monaco } = this.props;
+
+    const model = monaco.editor.createModel(value, this.language);
+    model.updateOptions({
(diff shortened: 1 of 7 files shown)


src/renderer/components/editor.tsx
```diff
@@ -98,34 +98,53 @@ export class Editor extends React.Component<EditorProps> {
...
    */
   private async setContent() {
-    const { appState, id, monaco } = this.props;
+    const { appState, id } = this.props;
     const { version } = appState;
 
     const backup = appState.getAndRemoveEditorValueBackup(id);
 
-    if (backup && backup.model) {
+    console.log(id);
```
malept: Stray debug line?

PR #294, opened 2020-01-21
fix: prevent other editors from clipping tooltips

### LINKED ISSUE #289: Tooltips obfuscated by other editors
Method tooltips are covered by other editors in the Fiddle UI and should not be.

Example:

Screen Shot 2020-01-07 at 7 29 17 PM

erickzhao:
Fixes #289 

Leverages the CSS `z-index` property and Monaco's `onDidFocusEditorText()` event handler to make the latest focused Editor pane higher in stack order than the others.

The Less selectors are a bit complicated because I couldn't add the `focused` class directly on each `<MosaicWindow/>`. This is due to the `<Mosaic/>` element not calling `renderTile()` unless the child `<MosaicWindow/>` elements were re-arranged.

### DIFF
--- a/src/less/components/editors.less
+++ b/src/less/components/editors.less
@@ -24,6 +24,21 @@
   }
 }
 
+// update if list of Editor ID changes
+@editor-ids: main, renderer, html, preload;
+
+// takes each editor ID and increase its z-index if the parent Mosiac root
+// has focused__id class for that specific id.
+each(@editor-ids, {
+  .focused__@{value}.mosaic .mosaic-window.@{value} {
+    z-index: 2;
+  }
+});
+
+.mosaic-window {
+  z-index: 1;
+}
+
 @keyframes fadein {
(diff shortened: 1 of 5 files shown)


src/less/components/editors.less
```diff
@@ -24,6 +24,21 @@
   }
 }
 
+// update if list of Editor ID changes
+@editor-ids: main, renderer, html, preload;
```
erickzhao: Less compiles to this CSS snippet:
```css
.focused__main.mosaic .mosaic-window.main {
  z-index: 2;
}
.focused__renderer.mosaic .mosaic-window.renderer {
  z-index: 2;
}
.focused__html.mosaic .mosaic-window.html {
  z-index: 2;
}
.focused__preload.mosaic .mosaic-window.preload {
  z-index: 2;
}
.mosaic-window {
  z-index: 1;
}
```

PR #303, opened 2020-01-26
fix: Make example warning more descriptive

### LINKED ISSUE #262: Infobox for launch from documentation, not informative
The formulation of this message it's not great for new users or not only for new.

> Are you sure you sure

😂. Yes I'm sure

> `example from the Electron docs`?

Here I'm planning to see the name of the example that launched, it gives some confidence it's really that what's user want.

> from fiddle path 'v7.0.0-beta.4'?

Maybe `this "name of fiddle" for Electron v7.0.0-beta.4`?

---

The Picture of the Box:

issacgerges:
Fixes #262.

Make example warning more descriptive by adding path, remove repeated text. 

## Screenshots
example

gist

## Considerations
* We may want to preflight the gist so we can tell the user the author name, at the moment we just say 'gist'
* The string replace to pretty up the example name is a bit weak, but the code is currently written to allow us to point to directories outside the fiddle directory. If we want to ban this all together I'm happy to catch it on the protocol handler side, only sending the pretty name to the app, which reconstructs the path later.
* Words are hard.

### DIFF
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -25,17 +25,20 @@ export class RemoteLoader {
 
   public async loadFiddleFromElectronExample(_: any, exampleInfo: { path: string; ref: string }) {
     console.log(`Loading fiddle from Electron example`, _, exampleInfo);
-    const ok = await this.verifyRemoteLoad('example from the Electron docs', exampleInfo.ref);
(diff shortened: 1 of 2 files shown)


issacgerges:
@felixrieseberg notice you mentioned on the issue you were creating a PR, happy to close if this is in-flight somewhere

PR #306, opened 2020-01-28
feat: bisect success dialog

### LINKED ISSUE #299: Add bisect complete dialog
I've noticed that the bisect result is pretty easy to lose in the console scrollback. In addition to logging the result, I'd love a dialog to display it and let me know the bisect is complete.

issacgerges:
Closes #299.

This change adds a dialog to show when the bisect is complete (to prevent the user from losing the complete log in their console scrollback).

bisect-complete


src/renderer/state.ts
```diff
@@ -270,23 +271,23 @@ export class AppState {
...
     }
   }
 
   @action public toggleBisectDialog() {
     this.isBisectDialogShowing = !this.isBisectDialogShowing;
   }
 
-  @action public toggleConfirmationPromptDialog() {
-    this.isConfirmationPromptShowing = !this.isConfirmationPromptShowing;
+  @action public toggleBisectCompleteDialog() {
```
erickzhao: Seems like we don't need this function after the refactor :)

PR #308, opened 2020-01-28
feat: focus on quit warning

### LINKED ISSUE #307: Show inactive windows on quit warning
I often close apps with `cmd+tab` pressing `q` over an app I want to close. At the moment, the quit warning shows, but the app isn't brought to the front.

Here's an example of how I'd like it to work
show-on-quit

issacgerges:
Fixes #307.

show-on-quit

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -171,6 +171,9 @@ export class AppState {
 
           this.isWarningDialogShowing = true;
 
+          const remote = require('electron').remote;
+          remote.getCurrentWindow().showInactive();
+
           // We'll wait until the warning dialog was closed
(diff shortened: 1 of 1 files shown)


src/renderer/state.ts
```diff
@@ -171,14 +171,16 @@ export class AppState {
 
           this.isWarningDialogShowing = true;
 
+          const remote = require('electron').remote;
```
erickzhao: Could we handle this with IPC directly rather than add an instance of `remote`?

issacgerges: Wasn't sure how ya'll felt about `remote` in this repo, happy to switch away. Will update this

PR #315, opened 2020-02-03
chore: don't retain global mainWindow ref

codebytere:
Ref https://github.com/electron/electron/pull/21957.

We no longer need to hold a global reference to `mainWindow` in order to prevent it being gc'd.

cc @erickzhao @felixrieseberg


malept:
I think I had the same idea 😅 #314

codebytere:
@malept ah - well, you beat me by about 18 minutes, so i will cede this one 😆

PR #318, opened 2020-02-05
chore: add sentry for crash / js error reporting

MarshallOfSound:
We have sentry, may as well use it to collect crash reports


felixrieseberg:
Hey, this is cool 🙇 

We probably want to start sentry only if we're not testing, which we could probably do in a single utility file.

Issue #323, opened 2020-02-11
Unable to select latest version in Bisect dialog


HashimotoYT:

felixrieseberg:
Closed in #324!

PR #324, opened 2020-02-12
fix: Reorganize the bisect helper

felixrieseberg:
This PR improves the existing bisect helper by turning our "Electron Version Select" into a generic component that can be consumed by both the command bar as well as the bisect helper. I have a feeling that we'll need more of these in the future.

It also ensures that the bisect helper will show the same versions as the command bar as configured in the preferences.

(This PR _seems_ to add more lines than it deletes, but that's just snapshots. The code is _actually_ smaller)

### DIFF
--- a/src/renderer/components/dialog-bisect.tsx
+++ b/src/renderer/components/dialog-bisect.tsx
@@ -1,22 +1,19 @@
-import { Button, Dialog, Label, MenuItem } from '@blueprintjs/core';
-import { Select } from '@blueprintjs/select';
+import { Button, ButtonGroup, Dialog, Label } from '@blueprintjs/core';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
 import { ElectronVersion } from '../../interfaces';
 import { Bisector } from '../bisect';
 import { AppState } from '../state';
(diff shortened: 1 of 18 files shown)


src/renderer/components/dialog-bisect.tsx
```diff
@@ -118,39 +118,23 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
...
-              onItemSelect={this.onEndSelect}
-              noResults={<MenuItem disabled={true} text='No results.' />}
-              disabled={!startIndex}
-            >
-              <Button
-                text={endIndex ? `v${allVersions[endIndex].version}` : ``}
-                icon={endIndex ? getItemIcon(allVersions[endIndex]) : 'small-minus'}
-                fill={true}
-                disabled={!startIndex}
+            <ButtonGroup fill={true}>
```
erickzhao: Now that there's no UI restriction preventing the Version Select from choosing a Latest Version that's earlier than the Earliest Version, could we add a ToolTip to the disabled button to explain the requirements to begin the Bisect?

erickzhao:
Also, this fixes #323 as far as I can tell?

felixrieseberg:
> Also, this fixes #323 as far as I can tell?

It does! You raise a good point about the usability, so let me add that feature in another PR 👍

PR #326, opened 2020-02-16
feat: Add some help to the bisect

felixrieseberg:
@erickzhao suggested, correctly, that the reorganization of code in #324 resulted in a bisect helper that wasn't _quite_ as helpful as it was before. I'm hoping that these suggestions go a bit above and beyond, making the result even more helpful than we had before - to make up for how I made it worse :wink:

Two things here: First, I added some actual help text. Secondly, the dropdowns now let you only select versions that make sense, ensuring that you never end up in a situation where you can't start the bisect. I left our sanity checks in there though, even if you should never be able to select versions that result in inability to start the bisect.

Help, by default mostly minimized:

Then extended:

Disabled versions:
image

### DIFF
--- a/src/renderer/components/dialog-bisect.tsx
+++ b/src/renderer/components/dialog-bisect.tsx
@@ -92,7 +92,6 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
    * Can we get this show on the road?
    */
   get canSubmit(): boolean {
-    console.log(`Calculating cansubmit`);
     return this.state.startIndex > this.state.endIndex;
   }
 
@@ -220,7 +219,7 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
     // 1: 4.0.0
     // 2: 3.0.0
     // ...
-    return allVersions.indexOf(version) < endIndex;
+    return allVersions.indexOf(version) < endIndex + 1;
   }
 
   /**
@@ -232,6 +231,6 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
   public isLatestItemDisabled(version: ElectronVersion): boolean {
     const { allVersions, startIndex } = this.state;
 
-    return allVersions.indexOf(version) > startIndex;
+    return allVersions.indexOf(version) > startIndex - 1;
   }
 }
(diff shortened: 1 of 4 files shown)


src/renderer/components/dialog-bisect.tsx
```diff
@@ -78,18 +81,31 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
...
+   */
+  public showHelp() {
+    this.setState({ showHelp: true });
+  }
+
+  /**
+   * Can we get this show on the road?
+   */
+  get canSubmit(): boolean {
+    console.log(`Calculating cansubmit`);
```
erickzhao: nit: unnecessary log statement?

src/renderer/components/dialog-bisect.tsx
```diff
@@ -145,4 +205,33 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
...
+  /**
+   * Should an item in the "latest version" dropdown be disabled?
+   *
+   * @param {ElectronVersion} version
+   * @returns {boolean}
+   */
+  public isLatestItemDisabled(version: ElectronVersion): boolean {
+    const { allVersions, startIndex } = this.state;
+
+    return allVersions.indexOf(version) > startIndex;
```
erickzhao:
```suggestion
    return allVersions.indexOf(version) > startIndex - 1;
```

Off by one -- current change still allows same version to be selected in both selectors.

src/renderer/components/dialog-bisect.tsx
```diff
@@ -145,4 +205,33 @@ export class BisectDialog extends React.Component<BisectDialogProps, BisectDialo
...
+   */
+  public isEarliestItemDisabled(version: ElectronVersion): boolean {
+    const { allVersions, endIndex } = this.state;
+
+    // In the array, "newer" versions will have a lower index.
+    // 0: 5.0.0
+    // 1: 4.0.0
+    // 2: 3.0.0
+    // ...
+    return allVersions.indexOf(version) < endIndex;
```
erickzhao:
```suggestion
    return allVersions.indexOf(version) < endIndex + 1;
```

Off by one -- current change still allows same version to be selected in both selectors.

src/renderer/components/dialog-bisect.tsx
felixrieseberg:
```suggestion
```

PR #330, opened 2020-02-27
Update dependencies

felixrieseberg:
This PR updates various dependencies, most notably Electron from v7 to v8.

PR #331, opened 2020-03-02
build: auto-update the Electron Releases JSON file every day

malept:
Follow-up from https://github.com/electron/fiddle/pull/330#discussion_r385443905, have Electron Bot (via GitHub Actions) automate adding new Electron releases to the JSON file.

Example run: https://github.com/malept/fiddle/runs/480768703?check_suite_focus=true (although on push instead of cron)
Example commit: https://github.com/electron/fiddle/commit/2ceadbbd7c125519f28a5c7f5f57d5aa8b1d1d4d

### DIFF
--- a/.github/workflows/electron-releases.yml
+++ b/.github/workflows/electron-releases.yml
@@ -1,0 +1,41 @@
+name: Auto-update Releases JSON file
+on:
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  autoupdate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+    - name: Get npm cache directory
+      id: npm-cache
+      run: |
+        echo "::set-output name=dir::$(npm config get cache)"
+    - uses: actions/cache@v1
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/electron-releases.yml
```diff
@@ -0,0 +1,41 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        echo "machine github.com login $GITHUB_ACTOR password $GITHUB_TOKEN" > ~/.netrc
+        chmod 600 ~/.netrc
+        git add static/releases.json
+        if test -n "$(git status -s)"; then
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "electron-bot@users.noreply.github.com"
+          git diff --cached
+          git commit -m "Update Electron Releases"
```
erickzhao: nit: If this commit is going directly onto the `master` branch, do we want to keep semantic commits?

malept: Yes, you're right. It should be
```suggestion
          git commit -m "build: update Electron Releases"
```

felixrieseberg:
Oh neat, this is cool. Thanks 🙇

malept:
FYI, this PR has the same status issue I mentioned in https://github.com/electron/fiddle/pull/333#issuecomment-593677419

PR #333, opened 2020-03-02
build: Use Ubuntu 18.04 in Travis CI

felixrieseberg:
We need to upgrade our Linux image so that we create proper `rpm` packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ node_js: "12"
 os:
 - linux
 - osx
-dist: trusty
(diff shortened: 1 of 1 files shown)


malept:
I'm not sure why the Travis CI status check is hanging, the build passed: https://travis-ci.org/electron/fiddle/builds/657520763

felixrieseberg:
It might just be that I have to re-authorize the webhook since we moved to the electronjs.org GSuite. Thanks for taking a look!

PR #340, opened 2020-03-06
chore: add test report generation helper

codebytere:
Add a command to generate a json report for a given test run. I'm working on consuming these in a reportable way, so this helps me out with that venture a bit.

cc @erickzhao @felixrieseberg

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "test": "jest --config=jest.json",
     "test:ci": "jest --config=jest.json --coverage --runInBand",
     "test:coverage": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
+    "test:report": "jest --config=jest.json --json --outputFile=report.json | exit 0",
     "tsc": "tsc --noEmit -p .",
     "electron-releases": "node ./tools/fetch-releases.js"
   },
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ yarn-error.log
 # Coverage
 coverage
 
+# Test report
+report.json
+
 # npm package
 /npm/dist
 /npm/path.txt
--- a/jest.json
+++ b/jest.json
@@ -4,7 +4,6 @@
   },
   "testURL": "http://localhost",
   "testRegex": "(-spec)\\.(ts|tsx)$",
-  "bail": true,
   "resetMocks": true,
   "resetModules": true,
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "lint:tests": "tslint \"./tests/**/*.ts{,x}\" -c tslint.json --fix",
     "lint:templates": "standard \"./static/show-me/**/*.js\"",
     "lint": "npm-run-all \"lint:*\"",
+    "generate-test-report": "jest --config=jest.json --json > report.json",
```
erickzhao: Can we call this `test:report`, since we already have `test:ci` and `test:coverage`?

package.json
```diff
@@ -22,6 +22,7 @@
     "test": "jest --config=jest.json",
     "test:ci": "jest --config=jest.json --coverage --runInBand",
     "test:coverage": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
+    "test:report": "jest --config=jest.json --json --outputFile=report.json | exit 0",
```
ckerr: The pipe-exit-0 call is a little weird

If you're wanting to test:report to unconditionally report success, this is probably what you want:

```suggestion
    "test:report": "jest --config=jest.json --json --outputFile=report.json || true",
```

Some background about the `|| true` idiom @ https://serverfault.com/questions/228281/bash-run-an-executable-giving-a-good-exit-status

OTOH, if you're wanting test:report to report success iff report.json was created -- e.g. if your caller doesn't care if the tests passed or failed, but does care that there is a report -- then you want to append another test:

```suggestion
    "test:report": "jest --config=jest.json --json --outputFile=report.json || true; test -f report.json",
```

ckerr: Separate issue I discovered while testing the above suggestions:

Because of the `"bail": true,` line in `jest.json`, this report won't get created if any tests fail, so a report will only exist if it's going to report 100% success.

I suspect that's not what you're after since that's not a very useful report, so you'd want to force the jest invocation here to override the config file by changing `bail`:

`"test:report": "jest --config=jest.json --bail=false --json ...`

Paired with the above suggestions, maybe something like this:

```suggestion
    "test:report": "jest --config=jest.json --bail=false --json --outputFile=report.json || true; test -f report.json",```

ckerr:
Welp, looks like I was a little too slow on this review :laughing:

PR #343, opened 2020-03-09
fix: test:report makes a report even if tests fail

ckerr:
Fixes the invocation of `test:report` to not bail out when tests fail. Bailing causes jest to not generate a report.

Discussion @ https://github.com/electron/fiddle/pull/340#pullrequestreview-371336100 but my code review was a few minutes too late :smile: 

CC @codebytere @erickzhao

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test": "jest --config=jest.json",
     "test:ci": "jest --config=jest.json --coverage --runInBand",
     "test:coverage": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "test:report": "jest --config=jest.json --json --outputFile=report.json | exit 0",
+    "test:report": "jest --config=jest.json --bail=false --json --outputFile=report.json || true; test -f report.json",
     "tsc": "tsc --noEmit -p .",
     "electron-releases": "node --unhandled-rejections=strict ./tools/fetch-releases.js"
   },


package.json
```diff
@@ -22,7 +22,7 @@
     "test": "jest --config=jest.json",
     "test:ci": "jest --config=jest.json --coverage --runInBand",
     "test:coverage": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "test:report": "jest --config=jest.json --json --outputFile=report.json | exit 0",
+    "test:report": "jest --config=jest.json --bail=false --json --outputFile=report.json || true; test -f report.json",
```
erickzhao: The `--bail=false` flag seems redundant here since #340 removed the `bail: true` from `jest.json` and Jest does not bail by default.

For good measure, I tried pasting `"test:report": "jest --config=jest.json --json --outputFile=report.json | exit 0"` into this branch's `package.json` and it still generated the test report.

```
> electron-fiddle@0.13.0 test:report /Users/ezhao/Developer/fiddle
> jest --config=jest.json --json --outputFile=report.json | exit 0

Test Suites: 1 failed, 96 passed, 97 total
Tests:       1 failed, 601 passed, 602 total
Snapshots:   60 passed, 60 total
Time:        7.954s, estimated 8s
Ran all test suites.
Test results written to: report.json

erickzhao: The rest of the change LGTM, though, although if my comment above is correct, we'd have to rename this PR.

ckerr: Oop, you are right!

No, IMO the rest of this isn't important enough to rename the PR.

PR #348, opened 2020-03-11
feat: show inactive window on quit warning

### LINKED ISSUE #307: Show inactive windows on quit warning

erickzhao:
Fixes #307.

Implements #308, but uses a new IPC event instead of `remote`.


src/main/windows.ts
```diff
@@ -61,6 +63,11 @@ export function createMainWindow(): Electron.BrowserWindow {
     shell.openExternal(url);
   });
 
+  ipcMainManager.on(IpcEvents.SHOW_INACTIVE, () => {
+    console.info('erick was here');
```
felixrieseberg: 😄

felixrieseberg:
```suggestion
    console.info('erick was here');
```

tests/main/windows-spec.ts
```diff
@@ -95,5 +97,12 @@ describe('windows', () => {
       (getOrCreateMainWindow().webContents as any).emit('will-navigate', e);
       expect(e.preventDefault).toHaveBeenCalled();
     });
+
+    it('shows the window on IPC event', () => {
+      const w = getOrCreateMainWindow();
+      ipcMainManager.emit(IpcEvents.SHOW_INACTIVE);
+      expect(w.showInactive).toHaveBeenCalled();
+
```
felixrieseberg:
```suggestion
```

src/main/windows.ts
erickzhao: 🙈

PR #349, opened 2020-03-11
chore: fix test:report bailing

codebytere:
Follow up to https://github.com/electron/fiddle/pull/340#pullrequestreview-371336100.

I accidentally forgot to cherry-pick over the `--bail=false` update from `jest.json` but you can also just specify it at the cli level so this does that. Also uses more idiomatic `|| true` instead of `|| exit 0`

cc @erickzhao @ckerr


codebytere:
@erickzhao no you're right - i intended to replace the default and only set bail to false for the test report generation. Should be fixed now!

PR #357, opened 2020-03-16
build: Only notarize during make

felixrieseberg:
This PR ensures that we only notarize during `npm run make` and only if we're on a Travis tag.

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -106,8 +106,8 @@ function notarizeMaybe() {
     return;
   }
 
-  if (process.env.npm_lifecycle_event !== 'make') {
-    console.log(`Not in "make" mode, skipping notarization`);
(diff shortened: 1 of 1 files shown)


forge.config.js
```diff
@@ -106,11 +110,21 @@ function notarizeMaybe() {
     return;
   }
 
+  if (process.env.npm_lifecycle_event !== 'make') {
+    console.log(`Not in "make" mode, skipping notarization`);
```
malept: I think we only want to do this when we're publishing. Although this will change back once we move to CircleCI.

```suggestion
  if (process.env.npm_lifecycle_event !== 'publish') {
    console.log(`Not in "publish" mode, skipping notarization`);
```

felixrieseberg:
@malept all good now?

PR #358, opened 2020-03-18
Refactor BinaryManager, add download progress indicator

felixrieseberg:
This is a fairly big refactor of what used to be `BinaryManager`. Instead of being a class, I've turned into a collection of methods, allowing a bit more flexibility and removing one more class that keeps internal state about Electron versions. This makes the code a bit more flexible and enabled me to replace `electron-download` with `@electron/get` - with the end goal of providing visual feedback for download progress. A picture probably says more, so here it is:

progress

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2185,10 +2185,9 @@
       }
     },
     "@electron/get": {
-      "version": "1.8.0",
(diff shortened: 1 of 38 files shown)


felixrieseberg:
@erickzhao Good catch! Not intentional, just didn't wire it up.

Issue #321, opened 2020-02-11
Downloading state drop to default when updating release list


HashimotoYT:
Repro steps:

1. Go to preferences -> electron
2. Start downloading
3. Click "Update Electron Release List"

Note: Downloading continues but no give visual feedback

Repro Screencast:

![hiworld](https://user-images.githubusercontent.com/24681191/74285947-75993d00-4d2f-11ea-9736-781dac8d7673.gif)

PR #361, opened 2020-03-24
feat: Add progress to settings page, too

felixrieseberg:
This PR adds a progress indicator to the settings page, too. It's a slightly different design than in the runner since the settings page uses slightly different components.

### DIFF
--- a/src/less/components/settings.less
+++ b/src/less/components/settings.less
@@ -52,6 +52,10 @@
     .action {
       text-align: right;
     }
+
+    .bp3-progress-bar {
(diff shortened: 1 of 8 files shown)


erickzhao:
Screen-Recording-2020-03-24-at-3 06 08-PM

Getting some weird behaviour when testing, with no errors logged to console. I believe that #321 also reappears with this change.

felixrieseberg:
Hmmm, looks fishy. You're right, I'll have to dig into that one.

PR #362, opened 2020-03-26
refactor: prefer app.whenReady() over .on('ready')

ckerr:
This is a "syntactic sugar" type change to make the `static/show-me/` demos follow the preferred practice of using `app.whenReady().then(...)` instead of `app.on('ready', ...)` so that any future app developers cribbing from Fiddle will use it too.

More discussion is at https://github.com/electron/electron/pull/21972


src/content/main-1-x-x.ts
```diff
@@ -21,7 +21,7 @@ function createWindow () {
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
-app.on('ready', createWindow)
+app.whenReady().then(createWindow)
```
malept: According to https://github.com/electron/electron/commit/fcc82ebd353d1704e111d4d84a8a1e5675840dff `whenReady` was added in v3, so we shouldn't change it here.

```suggestion
app.on('ready', createWindow)
```

erickzhao: Hmm... Good catch. If we're supporting Electron 2 in Fiddle, we would have to hack together another `src/content/main-2-x-x.ts` (not ideal).

ckerr:
> This is good, but we should probably be prepared for the (unlikely) event that there will be issues filed when people try to test Fiddles with Electron 2.x...

I agree with this point and am happy to make changes to the PR, but I think I need more clarity on what specific change you're requesting?

erickzhao:
Otherwise, another option would be to leave the `.on('ready')` event as-is for the default main process editor, but use `whenReady()` for the demos, as we don't really guarantee that they'll work for all versions of Electron.

erickzhao:
@ckerr I think I'll just merge this in as it's better and we should prime a better solution for this anyways.

PR #373, opened 2020-04-14
chore: update @electron/get version

codebytere:
This PR updates the bundled version of `@electron/get` to 1.10.0 in order to incorporate this new functionality: https://github.com/electron/get/pull/155 which allows for overriding the downloaded version of Electron with a custom one via environment variable.

Alternately, we might consider not committing the package lock, to prevent this problem; i'm open to either option!

cc @erickzhao @felixrieseberg

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,952 +10,445 @@
       "integrity": "sha1-nK+xca+CMpSQNTtIFvAzR6oVCjA="
     },
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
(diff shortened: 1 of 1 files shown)


malept:
Another alternative is that we enable dependabot on this repo.

The Travis CI failure is strange and seems like it should be rerun to make sure it's consistent.

erickzhao:
Restarted, failure is weird because we usually don't encounter many flakes in the unit tests. 🤷‍♂️

malept:
🤔 The failure appears to be in sending coverage info to coveralls.

felixrieseberg:
```
Error: Command failed: git cat-file -p d407ea9b0d2ce7ac01762cf47d1e2831102d1242
fatal: Not a valid object name d407ea9b0d2ce7ac01762cf47d1e2831102d1242
```

No idea what `git` is smoking here, so I'll just force-merge this in.

Issue #297, opened 2020-01-23
Add ability to reset default editor view


erickzhao:
Although we have a drag-and-drop interface that allows you to re-arrange the editors as much as possible. However, it's kind of a struggle to reset your arrangement if you misclick or whatnot.

It would be cool to add a button that programmatically sets the arrangement in the react-mosaic library to the `DEFAULT_MOSAIC_ARRANGEMENT` constant.

kimberleejohnson:
Hi, @erickzhao! I chatted with Shelley Vohr yesterday via her Twitter office hours, and she flagged this as a good first open source issue for a junior JavaScript developer with a bit of React experience like me. I've started poking around locally, and I hope to have something to contribute (following the Contributing.MD guidelines) soon. Thank you!

erickzhao:
Hey @kimberleejohnson, that's awesome to hear! Let me know if you have any questions. :)

kimberleejohnson:
Hi, @erickzhao! I took a first pass at this, but I had a few outstanding questions for you here: https://github.com/electron/fiddle/pull/374 

Please let me know if I missed something in the Contributing guidelines, and thank you for your help! 
Kimberlee

PR #374, opened 2020-04-16
feat: reset layout button

kimberleejohnson:
Addresses [Issue 297](https://github.com/electron/fiddle/issues/297). 

I added a button to reset the editors to their default layout by: 
- Adding a `resetEditorLayout` function in state.ts
- Calling that function onClick of a new Button in commands.tsx

I also: 
- Updated the jest Snapshot 
- Rearranged where the button is (I'd love feedback on where this looks best, and if Reset Layout is the best name. 

I have questions about: 
- Is there more testing I should add for this button? I thought not because the test as is for commands covers it? 
- I'm seeing a lot of odd code flagged as added in these files. Is this because of something funky with a formatter (Prettier), or are there other conventions I missed? 

Thank you for your help!

### DIFF
--- a/src/renderer/components/commands-editors.tsx
+++ b/src/renderer/components/commands-editors.tsx
@@ -22,10 +22,7 @@ export interface EditorDropdownProps {
  * @extends {React.Component<EditorDropdownProps, EditorDropdownState>}
  */
 @observer
-export class EditorDropdown extends React.Component<
-  EditorDropdownProps,
-  EditorDropdownState
-> {
+export class EditorDropdown extends React.Component<EditorDropdownProps, EditorDropdownState> {
   constructor(props: EditorDropdownProps) {
     super(props);
 
--- a/src/renderer/components/commands-editors.tsx
+++ b/src/renderer/components/commands-editors.tsx
@@ -1,4 +1,4 @@
-import { Button, Menu, MenuItem, Popover, Position } from '@blueprintjs/core';
+import { Button, Menu, MenuDivider, MenuItem, Popover, Position } from '@blueprintjs/core';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
@@ -83,11 +83,15 @@ export class EditorDropdown extends React.Component<EditorDropdownProps, EditorD
     }
 
     result.push(
-      <Button
-        icon='grid-view'
-        text='Reset Layout'
-        onClick={appState.resetEditorLayout}
-      />
+      <React.Fragment key={'fragment-reset-layout'}>
+      <MenuDivider />
+        <MenuItem
+          icon='grid-view'
+          key='reset-layout'
+          text='Reset Layout'
+          onClick={appState.resetEditorLayout}
+        />
+      </React.Fragment>
     );
 
     return result;
--- a/src/renderer/components/commands-editors.tsx
+++ b/src/renderer/components/commands-editors.tsx
@@ -1,11 +1,11 @@
-import { Button, Menu, MenuItem, Popover, Position } from '@blueprintjs/core';
-import { observer } from 'mobx-react';
-import * as React from 'react';
+import { Button, Menu, MenuItem, Popover, Position } from "@blueprintjs/core";
+import { observer } from "mobx-react";
+import * as React from "react";
 
-import { ALL_EDITORS, MosaicId, PanelId } from '../../interfaces';
-import { getVisibleMosaics } from '../../utils/editors-mosaic-arrangement';
-import { AppState } from '../state';
-import { TITLE_MAP } from './editors';
+import { ALL_EDITORS, MosaicId, PanelId } from "../../interfaces";
+import { getVisibleMosaics } from "../../utils/editors-mosaic-arrangement";
+import { AppState } from "../state";
+import { TITLE_MAP } from "./editors";
 
 export interface EditorDropdownState {
   value: string;
@@ -22,7 +22,10 @@ export interface EditorDropdownProps {
  * @extends {React.Component<EditorDropdownProps, EditorDropdownState>}
  */
 @observer
-export class EditorDropdown extends React.Component<EditorDropdownProps, EditorDropdownState> {
+export class EditorDropdown extends React.Component<
+  EditorDropdownProps,
+  EditorDropdownState
+> {
   constructor(props: EditorDropdownProps) {
     super(props);
 
@@ -33,7 +36,7 @@ export class EditorDropdown extends React.Component<EditorDropdownProps, EditorD
     return (
       <>
         <Popover content={this.renderMenu()} position={Position.BOTTOM}>
-          <Button icon='applications' text='Editors' />
+          <Button icon="applications" text="Editors" />
         </Popover>
         {this.renderDocsDemos()}
       </>
@@ -47,8 +50,8 @@ export class EditorDropdown extends React.Component<EditorDropdownProps, EditorD
 
     return (
       <Button
-        icon='help'
-        text='Docs & Demos'
+        icon="help"
+        text="Docs & Demos"
         id={PanelId.docsDemo}
         onClick={this.onItemClick}
         active={!this.props.appState.closedPanels.docsDemo}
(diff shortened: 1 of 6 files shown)


kimberleejohnson:
Hi, @erickzhao! Let me know if I should convert this over to a draft, and if you have any advice for debugging the coverage/coveralls failed test, thank you so much!

erickzhao:
Hey @kimberleejohnson, thanks for the ping! I'll give a review to discuss the code itself, but regarding your more general questions:

> I'm seeing a lot of odd code flagged as added in these files. Is this because of something funky with a formatter (Prettier), or are there other conventions I missed?

Prettier or another formatter installed globally would potentially conflict with our TSLint setup, so I'd put my money on that. For instance, I see that your formatter changed all single-quotes to double-quotes.

Ideally, you'd want to revert these formatting changes because they create inconsistencies in the codebase with respect to formatting, as well as making reviewing harder (since it takes longer to see which important lines changed.)

I think you could apply the repo's current formatting rules by installing the TSLint plugin for your editor of choice (e.g. [VSCode](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin)) and disabling. 

> if you have any advice for debugging the coverage/coveralls failed test

The Coveralls status check fails any time our test coverage dips, which is a nice goal, but sometimes unrealistic for each PR. If you want to see which new lines of code are uncovered, you'd want to click on [the coveralls.io page related to the build](https://coveralls.io/builds/30109820), which has a "NEW MISSED LINES IN DIFF" section.

Either way, the check is set up to be optional as well!

kimberleejohnson:
Ack, thank you so much for all your help, Erick! I'm getting to this this week and will keep you posted!

src/renderer/components/commands-editors.tsx
```diff
@@ -82,6 +82,14 @@ export class EditorDropdown extends React.Component<EditorDropdownProps, EditorD
...
     }
 
+    result.push(
+      <Button
+        icon='grid-view'
+        text='Reset Layout'
+        onClick={appState.resetEditorLayout}
+      />
+    );
+
```
kimberleejohnson, line 92: Once all of the Editors have been set in the Array, I thought it made sense to then push the Reset Layout button at the end. 

This made more sense to me as a user, but please shout if you disagree!

src/renderer/state.ts
```diff
@@ -635,6 +635,15 @@ export class AppState {
...
 
+  /**
+   * Resets editor view to default layout
+   *
+   *
+   */
+  @action public resetEditorLayout() {
+    this.mosaicArrangement = DEFAULT_MOSAIC_ARRANGEMENT;
+  }
+
```
kimberleejohnson, line 646: Function added to reset the editors.

tests/renderer/components/__snapshots__/commands-editors-spec.tsx.snap
```diff
@@ -203,6 +211,10 @@ exports[`EditorDropdown component renders the extra button if the FIDDLE_DOCS_DE
           shouldDismissPopover={true}
           text="Stylesheet (styles.css)"
         />
+        <Blueprint3.Button
+          icon="grid-view"
+          text="Reset Layout"
+        />
```
kimberleejohnson, line 217: Updated tests.

tests/renderer/state-spec.ts
```diff
@@ -560,4 +560,11 @@ describe('AppState', () => {
       });
     });
   });
+
+  describe('resetEditorLayout()', () => {
+    it('Puts editors in default arrangement', () => {
+      appState.resetEditorLayout();
+      expect(appState.mosaicArrangement).toEqual(DEFAULT_MOSAIC_ARRANGEMENT);
+    });
+  });
```
kimberleejohnson, line 569: Added test to check to make sure that the arrangement displays as the default.

src/renderer/state.ts
```diff
@@ -539,7 +540,6 @@ export class AppState {
       this.outputBuffer += strData;
       strData = this.outputBuffer;
       const parts = strData.split('\r\n');
-
```
kimberleejohnson, line 542: It seemed like this extra space violated TSLint rules? But let me know if that's wrong!

kimberleejohnson:
> Hey @kimberleejohnson, the change works great! A few general things:
> 
> * Suggestion: I think it'd crowd the commands bar a bit less if the new button was under the "Editors" dropdown.
> * As mentioned above, you'll have to revert the code formatting would have to be reverted before accepting the PR.
> * As for testing, I think you could add a test in `state.spec.ts` that checks if the correct editor layout is set when calling `resetEditorLayout()`. See [a similar test for `toggleConsole()`](https://github.com/electron/fiddle/blob/8ee42c290e3339115861064b0869d4b2e6e1b5d6/tests/renderer/state-spec.ts#L173-L181) for reference.

Thanks again, @erickzhao! I just applied changes and re-requested your review. Please shout if I missed anything.

tests/renderer/state-spec.ts
```diff
@@ -560,4 +560,11 @@ describe('AppState', () => {
       });
     });
   });
+
+  describe('resetEditorLayout()', () => {
+    it('Puts editors in default arrangement', () => {
+      appState.resetEditorLayout();
+      expect(appState.mosaicArrangement).toEqual(DEFAULT_MOSAIC_ARRANGEMENT);
```
HashimotoYT, line 567: It's a good test but then not very exponential. State by default generates mosaic with default arrangement and tests can pass when nothing is changed.

Here most optimal it's changing the default arrangement of mosaic for more credibility that test passed when something is really changed.

Here example for your case:
```suggestion
      appState.hideAndBackupMosaic(EditorId.main);

      expect(appState.mosaicArrangement).toEqual({
        direction: 'row',
        first: EditorId.renderer,
        second: EditorId.html
      });

      appState.resetEditorLayout();
      expect(appState.mosaicArrangement).toEqual(DEFAULT_MOSAIC_ARRANGEMENT);
      expect(appState.mosaicArrangement).toEqual({
        direction: 'row',
        first: EditorId.main,
        second: {
          direction: 'column',
          first: EditorId.renderer,
          second: EditorId.html,
        }
      });
```

Here we remove `main` window and verify that the only `renderer` and `html` are visible, and at the end, we check if all reset by default.

src/renderer/components/commands-editors.tsx
```diff
@@ -82,6 +82,14 @@ export class EditorDropdown extends React.Component<EditorDropdownProps, EditorD
       );
     }
 
+    result.push(
+      <Button
+        icon='grid-view'
+        text='Reset Layout'
+        onClick={appState.resetEditorLayout}
+      />
+    );
```
HashimotoYT, line 91: This button is very stood out from the content of that menu. Here for more beautiful be that if that button also uses `<MenuItem />` as the above, that allows to not stand out from styles. For creating the space between this button and above you can use the `<MenuDivider />` component.

Here example for your case:

```suggestion
    result.push(
      <React.Fragment key={'fragment-reset-layout'}>
        <MenuDivider />
        <MenuItem
          icon='grid-view'
          key='reset-layout'
          text='Reset Layout'
          onClick={appState.resetEditorLayout}
        />
      </React.Fragment>
    );
```

And how it's look in the app:

Untitled

@erickzhao Here used `React.Fragment`  instead of short-form because also need to have a unique `key`.

tests/renderer/state-spec.ts
erickzhao: Good point, it'd be better to test it this way since we indeed verify that the state changes. However, I would say that the last `expect()` call is redundant here since `{
        direction: 'row',
        first: EditorId.main,
        second: {
          direction: 'column',
          first: EditorId.renderer,
          second: EditorId.html,
        }
      }` is already hard-coded to `DEFAULT_MOSAIC_ARRANGEMENT`.

kimberleejohnson: Great flag, thank you so much @HashimotoYT! You and @erickzhao are helping me learn a lot about testing, and I updated the test based on your comments.

src/renderer/components/commands-editors.tsx
kimberleejohnson: Wow, that looks so much better, thank you so much! I've updated to this so the button stands out as something different than one of the Editor buttons, but doesn't look so clunky and out of place. Also used those components recommended and added in the missing key, thanks again!

erickzhao:
Will give this PR another look today!

kimberleejohnson:
> Will give this PR another look today!

Thank you so much 🥳 ! I made all the changes mentioned above and am looking forward to your review!

PR #386, opened 2020-04-28
feat: add ability to update gist instead of publishing

### LINKED ISSUE #290: Update existing Fiddle gist when publishing with non-empty URL field
After I publish a gist and change something and click `Publish` again, a new gist is created.

The gist should be updated instead (new revision). A new gist should only be created, iff the fiddle url address field is empty.

Maybe this is a crazy idea, but IMO a lot more intuitive as it is now.

HashimotoYT:
This PR adds the ability to update the existing Gist instead of publishing the new one. Also added toasts notifications when the gist published or updated.

GIFS 😃😃

### Publishing

publish the gist

### Updating

updating the gist

Fixes #290

### DIFF
--- a/src/renderer/components/commands-publish-button.tsx
+++ b/src/renderer/components/commands-publish-button.tsx
@@ -1,8 +1,9 @@
-import { Button, ButtonGroup, Menu, MenuItem, Popover, Position } from '@blueprintjs/core';
+import { Button, ButtonGroup, IToastProps, Menu, MenuItem, Popover, Position, Toaster } from '@blueprintjs/core';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
 import { when } from 'mobx';
+import { EditorValues } from '../../interfaces';
 import { IpcEvents } from '../../ipc-events';
 import { INDEX_HTML_NAME, MAIN_JS_NAME, PRELOAD_JS_NAME, RENDERER_JS_NAME, STYLES_CSS_NAME } from '../../shared-constants';
 import { getOctokit } from '../../utils/octokit';
@@ -13,6 +14,10 @@ export interface PublishButtonProps {
   appState: AppState;
 }
 
+interface IPublishButtonState {
+  readonly isUpdating: boolean;
+}
+
 /**
  * The "publish" button takes care of logging you in.
  *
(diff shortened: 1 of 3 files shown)


src/renderer/components/commands-publish-button.tsx
```diff
@@ -158,28 +159,55 @@ export class PublishButton extends React.Component<PublishButtonProps> {
...
+                content={privacyMenu}
+                position={Position.BOTTOM}
+              >
+                <Button
+                  icon={privacyIcon}
+                />
+              </Popover>
               <Button
-                icon={privacyIcon}
+                onClick={this.handleClick}
```
erickzhao, line 174: When we're in an updatable state, can we make this turn into `Update` button?

src/renderer/components/commands-publish-button.tsx
```diff
@@ -158,28 +159,55 @@ export class PublishButton extends React.Component<PublishButtonProps> {
...
-      <fieldset disabled={isPublishing}>
-        <ButtonGroup className='button-publish'>
-            <Popover
-              content={privacyMenu}
-              position={Position.BOTTOM}
-            >
+      <>
+        <fieldset disabled={isPublishing}>
+          <ButtonGroup className='button-publish'>
+              <Popover
```
erickzhao, line 165: This button currently doesn't do anything when updating. We should either:
* Not show the privacy button when updating
* Update the privacy state as well

HashimotoYT: Updating the privacy of gist is not available from API, so I just removed that button when we have gist 🙃

src/renderer/components/commands-publish-button.tsx
HashimotoYT: Done 😃

src/renderer/components/commands-publish-button.tsx
```diff
@@ -158,28 +159,55 @@ export class PublishButton extends React.Component<PublishButtonProps> {
...
-              onClick={this.handleClick}
-              loading={isPublishing}
-              icon='upload'
-              text={isPublishing ? 'Publishing...' : 'Publish'}
-            />
-        </ButtonGroup>
-      </fieldset>
+          </ButtonGroup>
+        </fieldset>
+        <Toaster position={Position.TOP_RIGHT} ref={this.refHandlers.toaster} />
```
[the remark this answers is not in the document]
HashimotoYT: I changed position to `BOTTOM_RIGHT. I don't change colors and unable to verify because Octokit broke for me. 🤷‍♂️

HashimotoYT:
@erickzhao I reverted changes to octokit since they create some problems, he should be updated in the future PRs.

It's ready for re-review and sorry for override history.

PR #389, opened 2020-05-09
fix: set the window to null after destroying

### LINKED ISSUE #379: TypeError: Object has been destroyed
Sentry Issue: [ELECTRON-FIDDLE-29](https://sentry.io/organizations/electronjs/issues/1565863761/?referrer=github_integration)

```
TypeError: Object has been destroyed
  File "app:///dist/src/main/main.js", line 528, in IpcMainManager.<anonymous>
    browserWindow.showInactive();
  File "app:///dist/src/main/main.js", line 577, in IpcMainImpl.<anonymous>
    electron_1.ipcMain.on(name, (...args) => this.emit(name, ...args));
...
(5 additional frame(s) were not displayed)
```

HashimotoYT:
How Electron docs say for [`closed` event](https://www.electronjs.org/docs/api/browser-window#event-closed) after this event is fired we should remove the `browserWindow` and never use them. Previously we still can use these `browserWindow` if the window doesn't more exist. This PR makes `browserWindow` null after they removed.

Fixes #379 
/cc @erickzhao

### DIFF
--- a/src/main/windows.ts
+++ b/src/main/windows.ts
@@ -5,7 +5,7 @@ import { ipcMainManager } from './ipc';
 
 // Keep a global reference of the window objects, if we don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
-export let browserWindows: Array<BrowserWindow> = [];
+export let browserWindows: Array<BrowserWindow | null> = [];
 
 /**
  * Gets default options for the main window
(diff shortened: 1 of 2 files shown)


src/main/windows.ts
```diff
@@ -37,11 +37,14 @@ export function getMainWindowOptions(): Electron.BrowserWindowConstructorOptions
...
   console.log(`Creating main window`);
-  const browserWindow = new BrowserWindow(getMainWindowOptions());
+  let browserWindow: BrowserWindow | null
+  browserWindow = new BrowserWindow(getMainWindowOptions());
   browserWindow.loadFile('./dist/static/index.html');
 
   browserWindow.webContents.once('dom-ready', () => {
-    browserWindow.show();
+    if (browserWindow) {
+      browserWindow.show();
```
erickzhao, line 46: Could move this to the `if (browserWindow)` at line 49

src/main/windows.ts
```diff
@@ -37,20 +37,23 @@ export function getMainWindowOptions(): Electron.BrowserWindowConstructorOptions
...
+
       createContextMenu(browserWindow);
     }
   });
 
   browserWindow.on('closed', () => {
     browserWindows = browserWindows
       .filter((bw) => browserWindow !== bw);
+
+    browserWindow = null
```
erickzhao, line 56: nit: should end line with `;` for consistency

(didn't see this yesterday, sorry 🙇 )

HashimotoYT: efb0b19

PR #401, opened 2020-05-25
refactor: replace app.getName() with app.name

erickzhao:
As `app.getName` is deprecated, this PR replaces it with the `app.name` property.


vantuan1994:
src/main/menu.ts

PR #404, opened 2020-05-27
fix: make hidden editors interact properly with remote content

### LINKED ISSUE #387: Content from Gist won't load if the editor is hidden
1. Load the fiddle (any fiddle with preload.js) https://gist.github.com/Prinzhorn/dfe34f7b9e6e90fbde99f5f20e402747 
2. Click Editors -> preload.js to show the editor
3. It is empty

Now if you load the gist again with the visible preload.js editor it is populated.

This means gists with a preload.js are broken by default, because the preload.js editor is hidden by default.

### LINKED ISSUE #400: Publishing to Gist does not work if editor is hidden
If a non-empty editor is hidden for some reason, publishing a gist will upload an empty file instead of the editor's contents.

This is kind of the inverse of #387

erickzhao:
Fixes #387
Fixes #400

### Context

We didn't account for hidden editors when considering our interactions with publishing Gists. This presented itself in two ways:

1) Remote content (including Gists) would only load if the editor was loaded at some point. This is because we checked each editor for its corresponding remote content.
2) Published Gists would be missing content if an editor was hidden.

### Changes

1) When loading remote content, we now first open all editors that have non-placeholder content. This ensures that the content can load into the editors, and provides an all-around better experience since all the content from the Gist will be loaded on-screen.

2) When hiding an editor, we now delete it from the `editors` mapping before creating its backup in `appState.closedPanels`. We also now check the value of backups when calling `getEditorValue`. These two changes ensure that any hidden content gets properly published into the Gist.

3) We unfortunately didn't get the above two fixes for free. I had to make `setVisibleMosaics` asynchronous -- it now waits until all Monaco editors are loaded before resolving in order to avoid any race conditions with assuming editors are ready right after we open their panel. This is done in a hacky way by periodically polling the editors object until their contents match the mosaics that we've opened.

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -604,6 +604,9 @@ export class AppState {
         this.closedPanels[id] = isEditorId(id)
           ? getEditorBackup(id)
           : true;
+
+        // if we have backup, remove active editor
+        delete window.ElectronFiddle.editors[id];
       }
 
       // Remove the backup for panels now. Editors will remove their
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -596,7 +596,7 @@ export class AppState {
     return null;
   }
 
-  @action public setVisibleMosaics(visible: Array<MosaicId>) {
+  @action public async setVisibleMosaics(visible: Array<MosaicId>) {
     const currentlyVisible = getVisibleMosaics(this.mosaicArrangement);
 
     for (const id of ALL_MOSAICS) {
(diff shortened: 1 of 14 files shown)


src/renderer/state.ts
```diff
@@ -617,6 +620,31 @@ export class AppState {
     console.log(`State: Setting visible mosaic panels`, visible, updatedArrangement);
 
     this.mosaicArrangement = updatedArrangement;
+
+    // after the mosaicArrangement loads, we want to wait for the Mosaic editors to
+    // mount to ensure that we can load content into the editors as soon as they're
+    // declared visible.
+
+    const waitForEditorsToMount = () => {
```
felixrieseberg, line 628: Should we maybe break this out?

PR #408, opened 2020-06-05
fix: correct sense of bisect buttons

nornagon:
this matches 'git bisect'.


nornagon:
ci is sad about the fact that this is a fork, i think. i don't have permissions to create a branch on e/fiddle.

HashimotoYT:
@nornagon Need to update the tests:

https://github.com/electron/fiddle/blob/6aa19eada69b01f35b423edcff571b601dc25b37/tests/renderer/components/commands-bisect-spec.tsx#L54

https://github.com/electron/fiddle/blob/6aa19eada69b01f35b423edcff571b601dc25b37/tests/renderer/components/commands-bisect-spec.tsx#L59

PR #413, opened 2020-06-14
feat: allow to publish if user would to do it

### LINKED ISSUE #412: Update vs publish DX can trip up users
Recently fiddle added the ability to update gists when they already exist, such that new fiddles don't get published when one is already working within the app. This is great, but creates a bit of a developer experience trap if someone _does_ want to create a new fiddle after they change code in currently open editors. I ran into this issue today when creating a fiddle for a bug report and then swapping to a new bug - it updated my initial gist and removed my code for the initial bug report. 

Ideally, i'd like to split these two options out - the default I think can/should remain update when a fiddle is already open, but i'd like to retain the option to forcibly publish a new one. If the flow we want to push users toward is to ask them to open a new fiddle every time i do understand that, so in that case we might want to somehow make that clearer to users?

I'm not sure exactly what yall prefer that to look like, but i'm happy to take a stab at it!

HashimotoYT:
Fixes #412 

GIF 🙌

/cc @codebytere

### DIFF
--- a/src/renderer/components/commands-publish-button.tsx
+++ b/src/renderer/components/commands-publish-button.tsx
@@ -17,6 +17,7 @@ export interface PublishButtonProps {
 
 interface IPublishButtonState {
   readonly isUpdating: boolean;
+  readonly wouldPublish: boolean;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


erickzhao:
cc @codebytere wondering what you thought of this UI!

codebytere:
I like it! it definitely solves the pain point i initially found myself hitting ✨

PR #422, opened 2020-06-25
fix: package nightlies correctly

codebytere:
Fixes packaging nightly builds. 

Nightlies come from `electron-nightly` and not `electron`, so we need to change the name in package.json. We also need to force a new ABI number as none has been assigned for nightlies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "chokidar": "^3.3.1",
     "coveralls": "^3.0.9",
     "electron": "8.0.2",
+    "electron-rebuild": "^1.11.0",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
     "enzyme-to-json": "^3.4.4",
--- a/package-lock.json
+++ b/package-lock.json
@@ -6508,9 +6508,9 @@
       }
     },
     "electron-rebuild": {
-      "version": "1.10.1",
-      "resolved": "https://registry.npmjs.org/electron-rebuild/-/electron-rebuild-1.10.1.tgz",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -84,6 +84,7 @@
     "chokidar": "^3.3.1",
     "coveralls": "^3.0.9",
     "electron": "8.0.2",
+    "electron-rebuild": "^1.11.0",
```
malept, line 87: Where is this directly used? All I see is an addition to Forge's `electronRebuildConfig`, and Forge already depends on `electron-rebuild`.

codebytere: the version there doesnt include the latest release, which has the ABI override functionality - i can bump it in forge though and then bump for here if you'd prefer

malept: I think you can just run `npm upgrade --depth 100 electron-rebuild` or something. _(This is also known as the "why isn't node-abi updated" problem :sob:)_

HashimotoYT: > This is also known as the "why isn't node-abi updated" problem 😭

Maybe time for additional documentation about that in `electron-rebuild` readme?

malept: I think the error just needs to be clarified.

codebytere: @malept can you expand? the issue here is that there is no _assigned_ ABI number, and so one needs to force one.

malept: Sorry, my point was that `electron-rebuild` is a transitive dependency of Fiddle, so it's not necessary to add it as an explicit dependency just for the purposes of forcing a version bump.

I only mention `node-abi` because Electron Forge has the same transitive dependency problem with that module.

codebytere:
Closing for now.

PR #424, opened 2020-06-27
fix: transparent BrowserWindow show-me on Windows

shastriUF:
The `transparent` configuration on Windows OS requires the window to also be frameless. Updating the show-me to handle win32. 

The show-me now also demonstrates the `frame` configuration. 

This behavior is documented at the entry for `transparent` at the [BrowserWindow API doc](https://www.electronjs.org/docs/api/browser-window#new-browserwindowoptions).

### DIFF
--- a/static/show-me/browserwindow/main.js
+++ b/static/show-me/browserwindow/main.js
@@ -23,15 +23,27 @@ app.on('ready', () => {
       y: 200
     })
   )
-
-  // A transparent window
-  windows.push(
-    new BrowserWindow({
-      transparent: true,
-      x: 300,
-      y: 300
-    })
-  )
+  
+  // A transparent window. On Windows OS, a transparent window must be frameless
+  if (process.platform != 'win32') {
+    windows.push(
+      new BrowserWindow({
+        transparent: true,
+        x: 300,
+        y: 300
+      })
+    )  
+  }
+  else {
+    windows.push(
+      new BrowserWindow({
+        transparent: true,
+        frame: false,
+        x: 300,
+        y: 300
+      })
+    )
+  }
 
   // A window that's fixed and always on top
   windows.push(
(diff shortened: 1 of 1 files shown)


static/show-me/browserwindow/main.js
```diff
@@ -23,15 +23,27 @@ app.on('ready', () => {
...
+  else {
+    windows.push(
+      new BrowserWindow({
+        transparent: true,
+        frame: false,
+        x: 300,
+        y: 300
+      })
+    )
+  }
```
malept, line 46: We want to keep it as simple as possible to maintain, so it would be best to minimize the amount of duplicate code. Additionally, for future reference, you generally want to start your if-else conditionals with the positive check.

```suggestion

  // A transparent window.
  const windowOptions ={
    transparent: true,
    x: 300,
    y: 300
  }
  // On Windows platforms, a transparent window must be frameless
  if (process.platform === 'win32') {
    windowOptions.frame = false
  }
  windows.push(new BrowserWindow(windowOptions))
```

shastriUF:
The builds failed because of some style issues,.. Will run linter and push again.

static/show-me/browserwindow/main.js
shastriUF: Thanks, this is much cleaner! Pushed new changes that include this.

shastriUF: @malept I resolved this comment thread since I pushed the update, but that doesn't make the "Requested changes must be addressed" blocker go away. I reset your review status, hopefully if you approve the review that will clear out the "change requested" blocker?

Apologies for the confusion, I'm a little new to the GitHub PR workflow. In Azure DevOps PRs, marking a comment thread as "resolved" clears that request throughout the PR, so that's what I expected would happen here too. But evidently GitHub works a little differently.

shastriUF:
@erickzhao Friendly reminder for this review :) 

I believe I have addressed all the issues raised previously and have a passing build, so if there's anything more I'm expected to do before this can be merged in, please let me know?

PR #426, opened 2020-06-30
fix: audio/video entitlements crash

codebytere:
Fiddle was blowing up spectacularly when i tried to test 

```
const { systemPreferences } = require('electron')

const microphone = systemPreferences.askForMediaAccess("microphone")
```

and a journey into the Console app informed me:

```
This app has crashed because it has a hardened runtime and attempted to access privacy-sensitive data without an entitlement indicating its intent to access this data.  The app must have the 'com.apple.security.device.audio-input' entitlement.
```

and so this adds relevant entitlements to fiddle to fix said crash.

cc @felixrieseberg @erickzhao

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -16,12 +16,11 @@ const config = {
     executableName: 'electron-fiddle',
     asar: true,
     icon: path.resolve(__dirname, 'assets', 'icons', 'fiddle'),
-    // TODO: FIXME?
-    // ignore: [
-    //   /^\/\.vscode\//,
-    //   /^\/tools\//
-    // ],
     appBundleId: 'com.electron.fiddle',
+    usageDescription: {
+      Camera: 'Access is needed by certain built-in fiddles in addition to any custom fiddles that use the Camera',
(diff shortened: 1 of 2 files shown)


malept:
Perhaps we should use this too? http://electron.github.io/electron-packager/v15.0.0/interfaces/electronpackager.options.html#usagedescription

codebytere:
@malept we only really need those to override the defaults from [here](https://github.com/electron/electron/blob/8045152e592e115fafa578c09802956336a4cd99/shell/browser/resources/mac/Info.plist#L39-L42) and i think for our purposes it's probably fine to leave them generic?

malept:
It might be nice to say "this is used by certain built-in fiddles plus any custom fiddles that uses camera/mic" (as appropriate)

codebytere:
@malept done!

codebytere:
@malept looks like i angered coveralls - any ideas?

codebytere:
@erickzhao that's my current understanding!

Issue #418, opened 2020-06-18
Remove usage of Electron `remote` 


vhashimotoo:
In Electron from version 10 `enableRemoteModule` default value changed to `false`. This issue created to track replacing the `remote` import by the `IPC` layer.

1. [x] [`src/renderer/components/show-me/app.tsx`] (https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx) (Need to figure where it's used)
	- https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx#L18
	- https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx#L20
	- https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx#L21
	- https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx#L45
	- https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx#L63
	- https://github.com/electron/fiddle/blob/374a85e2f65c5acf99f62001ade139ae0e29e441/src/renderer/components/show-me/app.tsx#L71
2. [x] [`src/renderer/components/chrome-mac.tsx`](https://github.com/electron/fiddle/blob/dca2cd67e545dbe39ca705db526c7325eb25a111/src/renderer/components/chrome-mac.tsx) (See #598)
	- https://github.com/electron/fiddle/blob/dca2cd67e545dbe39ca705db526c7325eb25a111/src/renderer/components/chrome-mac.tsx#L15
	- https://github.com/electron/fiddle/blob/dca2cd67e545dbe39ca705db526c7325eb25a111/src/renderer/components/chrome-mac.tsx#L16
3. [x] [`src/renderer/constants.ts`](https://github.com/electron/fiddle/blob/b98472f973ecb60dd4f2fe07b94be4ae3d658cef/src/renderer/constants.ts)
	- https://github.com/electron/fiddle/blob/b98472f973ecb60dd4f2fe07b94be4ae3d658cef/src/renderer/constants.ts#L10
[...]

PR #430, opened 2020-07-03
feat: set the `appData` dir from the main process

HashimotoYT:
This PR moves the getting the `app data` folder from remote to single IPC call on the start of the application and saves them to state, what's can be used around the app in another place.

Ref #418


src/renderer/runner.ts
```diff
@@ -281,7 +280,7 @@ export class Runner {
     }
 
     const name = await this.appState.getName();
-    const appData = getAppDataDir(name);
+    const appData = this.appState.getAppDataDir(name);
```
erickzhao, line 283: Feel like we could get rid of `this.appState.getAppDataDir()` and just call `path.join` directly here.

HashimotoYT: 171b2ae

PR #437, opened 2020-07-14
fix: clearer document titles

### LINKED ISSUE #396: Saving a Fiddle locally after publishing a gist shows both the gist URL and the local path in the titlebar
Repro:
* Publish Fiddle
* Save Fiddle locally

Observe something like this:

image

I don't know what the ideal solution here is, but we should be choosing one or the other.

erickzhao:
Fixes #396 

This PR clears the `localPath` variable in the App State whenever a `gistId` is saved, and vice-versa. Although I do still think that the user experience here still has a lot of room for improvement, this change makes it slightly clearer because you only have one source of truth with each save/publish.

Caveat: if you save a Fiddle that was previously published as a Gist, the Gist ID disappears in the publish bar.

### DIFF
--- a/src/renderer/app.tsx
+++ b/src/renderer/app.tsx
@@ -54,6 +54,7 @@ export class App {
     editorValues: Partial<EditorValues>,
     { filePath, gistId, templateName }: Partial<SetFiddleOptions>
   ) {
+    console.warn('@@@@@', filePath, gistId, templateName)
     // if unsaved, prompt user to make sure they're okay with overwriting and changing directory
     if (this.state.isUnsaved) {
       this.state.setGenericDialogOptions({
@@ -81,7 +82,6 @@ export class App {
     }
 
 
-    document.title = getTitle(this.state);
     this.state.gistId = gistId || '';
     this.state.localPath = filePath;
     this.state.templateName = templateName;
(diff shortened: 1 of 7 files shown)


src/renderer/app.tsx
```diff
@@ -54,6 +54,7 @@ export class App {
     editorValues: Partial<EditorValues>,
     { filePath, gistId, templateName }: Partial<SetFiddleOptions>
   ) {
+    console.warn('@@@@@', filePath, gistId, templateName)
```
HashimotoYT, line 57: 🔥🔥🔥
```suggestion
```

tests/utils/gist-spec.ts
```diff
@@ -31,5 +31,10 @@ describe('gist', () => {
       const result = urlFromId(mockId);
       expect(result).toBe(`https://gist.github.com/${mockId}`);
     });
+
+    it('returns an empty string if id is undefined', () => {
+      const result = urlFromId(undefined);
```
HashimotoYT, line 36: It's still right but can be also used without providing `undefined`, leave that for you

```suggestion
      const result = urlFromId();
```

src/renderer/app.tsx
erickzhao: Big oopsies, thanks for the catch!

src/utils/gist.ts
```diff
@@ -21,6 +21,6 @@ export function idFromUrl(input: string): string | null {
  * @param {string} input
  * @returns {string}
  */
-export function urlFromId(input: string): string {
-  return `https://gist.github.com/${input}`;
+export function urlFromId(input: string | undefined): string {
```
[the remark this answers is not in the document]
erickzhao: Clean!

PR #439, opened 2020-07-16
fix: proper toggling for devtools

### LINKED ISSUE #433: Toggle Developer Tools Not Toggling
Clicking on Toggle Developer Tools ( or using Ctrl + Alt + I ) from the Application Menu opens the developer tools , however , it does not closes on clicking over the option again . User has to manually click the cross button from the opened dev tools.

## To Reproduce (Linux)
Open Fiddle and tap the Toggle Developer Tools sub-menu within the Help menu.
Dev tools show up.
Attempt to quit it by clicking the same sub-menu again.

## Expected behaviour
Dev tools must toggle .

## Actual behaviour
Nothing happens.

erickzhao:
Closes #433

### DIFF
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -45,12 +45,7 @@ function getHelpItems(): Array<MenuItemConstructorOptions> {
         const browserWindow = BrowserWindow.getFocusedWindow();
 
         if (browserWindow && !browserWindow.isDestroyed()) {
-          const { webContents } = browserWindow;
-          if (webContents.isDevToolsOpened()) {
-            browserWindow.webContents.closeDevTools();
-          } else {
-            browserWindow.webContents.openDevTools();
-          }
+          browserWindow.webContents.toggleDevTools();
         }
       }
     },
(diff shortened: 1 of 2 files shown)


src/main/menu.ts
```diff
@@ -45,7 +45,12 @@ function getHelpItems(): Array<MenuItemConstructorOptions> {
         const browserWindow = BrowserWindow.getFocusedWindow();
 
         if (browserWindow && !browserWindow.isDestroyed()) {
-          browserWindow.webContents.openDevTools({ mode: 'bottom' });
+          const { webContents } = browserWindow;
+          if (webContents.isDevToolsOpened()) {
+            browserWindow.webContents.closeDevTools();
+          } else {
+            browserWindow.webContents.openDevTools();
+          }
```
HashimotoYT, line 53: If we removing the `options: { mode: 'bottom' }`, why not use the `toggleDevTools` ([docs](https://www.electronjs.org/docs/api/web-contents#contentstoggledevtools)) what do the same thing. If we still require the `options`, please return them back and that be look more reasonable.

erickzhao: @HashimotoYT good catch, didn't see that API method. I don't know why the `mode: 'bottom'` was required there, maybe @felixrieseberg could chime in?

PR #444, opened 2020-07-20
build(deps): upgrade electron-default-menu to 1.0.2

### LINKED ISSUE #441: `electron-default-menu` package uses deprecated app.getName() API
When running Fiddle, we get the following warning message, even though we've removed all direct instances of `app.getName()` in the Fiddle code:

```console
(electron) 'getName function' is deprecated and will be removed. Please use 'name property' instead.
```

This package we depend on for setting up a default menu in Electron still uses the old API. I've submitted a pull request (https://github.com/carter-thaxton/electron-default-menu/pull/11) to update it, but it's something we should look out for when we eventually upgrade to an Electron version that doesn't support the `app.getName()` API anymore.

erickzhao:
Closes #441

Updates to `^1.0.2`

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@octokit/rest": "^16.43.1",
     "@sentry/electron": "^1.2.1",
     "classnames": "^2.2.6",
-    "electron-default-menu": "^1.0.1",
+    "electron-default-menu": "1.0.2",
     "electron-devtools-installer": "^2.2.4",
     "electron-squirrel-startup": "^1.0.0",
     "extract-zip": "^1.6.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "@octokit/rest": "^16.43.1",
     "@sentry/electron": "^1.2.1",
     "classnames": "^2.2.6",
-    "electron-default-menu": "^1.0.1",
+    "electron-default-menu": "1.0.2",
```
malept, line 46:
```suggestion
    "electron-default-menu": "^1.0.2",
```

erickzhao: I re-read the caret did and indeed I shouldn't have pinned it.

PR #206, opened 2019-06-07
build: Move to prettier

felixrieseberg:
This PR doesn't change much - we're primarily just moving to `prettier`. TSLint is deprecated and was just a little bit more annoying than it needed to be.

PR #445, opened 2020-07-23
build: replace TSLint with ESLint and Prettier

erickzhao:
Ref #206

We've been wanting to move from TSLint to ESLint and Prettier for a while. The maintainers of TSLint deprecated the package last year (https://github.com/palantir/tslint/issues/4534), so this is a good time to transition.

I tried to adapt our previous TSLint config to match the new ESLint one as much as possible, but tried to keep it close to the default Prettier settings as well.

Note that this PR also adds `lint-staged` and `husky` for pre-commit hook linting.


package.json
```diff
@@ -81,29 +80,51 @@
...
+  "lint-staged": {
+    "./**/*.{js,ts,tsx}": [
+      "npm run lint:ts"
+    ],
+    "./static/show-me/**/*.js": [
+      "npm run lint:templates"
+    ],
+    "./src/less/*.less": [
+      "npm run lint:style"
+    ]
```
malept, line 128: This isn't going to work quite the way you want it to (I know this from experience). What should happen is that only files that are changed should get run through the appropriate linter. In practice, all of the files specified in the lint subcommand will always get processed. I'd replace the `npm run` commands with the bare linter commands (sans the file globs).

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  semi: true,
+  trailingComma: "all",
+  singleQuote: true,
+  tabWidth: 2
```
malept, line 5: I think the only custom option is `singleQuote: true`?
```suggestion
  singleQuote: true,
```

erickzhao: Looked at the options over at https://prettier.io/docs/en/options.html and found that `trailingComma` is also set to `es5` by default. I'll remove the other ones, though!

PR #447, opened 2020-07-24
feat: add ability to select package between npm and yarn

HashimotoYT:
This PR adds the setting that gives the ability to change the default package manager from npm to Yarn and back.

### DIFF
--- a/src/renderer/components/settings-execution.tsx
+++ b/src/renderer/components/settings-execution.tsx
@@ -1,8 +1,36 @@
-import { Callout, Checkbox, FormGroup, InputGroup } from '@blueprintjs/core';
+import {
+  Button,
+  Callout,
+  Checkbox,
+  FormGroup,
+  InputGroup,
+  MenuItem,
+} from '@blueprintjs/core';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
+import { ItemRenderer, Select } from '@blueprintjs/select';
 import { AppState } from '../state';
+
+const PMSelect = Select.ofType<'npm' | 'yarn'>();
+
+export const renderItem: ItemRenderer<'npm' | 'yarn'> = (
+  item,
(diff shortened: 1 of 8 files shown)


src/renderer/components/settings-execution.tsx
```diff
@@ -125,7 +153,27 @@ export class ExecutionSettings extends React.Component<ExecutionSettingsProps, {
             />
           </FormGroup>
         </Callout>
+        <br />
+        <Callout>
+          <FormGroup>
+            <span style={{ marginRight: 4 }}>
+              You can change the default package manager to 'npm' or 'yarn'. The
```
erickzhao, line 160: Something like this might make sense for the text to give more context on what the selected package manager does:

> Electron Fiddle will install packages on runtime if they are imported within your fiddle with `require`. It uses [npm](https://www.npmjs.com/) as its package manager by default, but [Yarn](https://classic.yarnpkg.com/lang/en/) is also available.

src/utils/exec.ts
```diff
@@ -14,6 +14,7 @@ export async function exec(dir: string, cliArgs: string): Promise<string> {
 
     const { exec: cpExec } = await import('child_process');
 
+    console.log(cliArgs);
```
erickzhao, line 17: stray log
```suggestion
```

src/renderer/runner.ts
```diff
@@ -258,10 +261,10 @@ export class Runner {
    * @returns
    * @memberof Runner
    */
-  public async npmInstall(dir: string): Promise<boolean> {
+  public async npmInstall(options: PMOperationOptions): Promise<boolean> {
```
erickzhao, line 264: maybe rename to `packageInstall` for clarity?

erickzhao:
@HashimotoYT seems like appveyor failed? not sure how to restart the build from their UI 🤐

PR #452, opened 2020-08-05
chore: ensure yarn is installed before launching fiddle

vhashimotoo:
Doing the same thing what's we do with `npm`. It's a super simple solution that's not represented in the UI but represents on the launch.

### DIFF
--- a/src/renderer/npm.ts
+++ b/src/renderer/npm.ts
@@ -34,43 +34,37 @@ const isUnique = (item: any, idx: number, arr: Array<any>): boolean => {
 };
 
 /**
- * Checks if yarn is installed by checking if a binary
+ * Checks if package manager is installed by checking if a binary
  * with that name can be found.
  */
-export async function getIsYarnInstalled(
+export async function getIsPackageManagerInstalled(
+  packageManager: IPackageManager,
   ignoreCache?: boolean,
 ): Promise<boolean> {
-  if (isYarnInstalled !== null && !ignoreCache) return isYarnInstalled;
+  if (packageManager === 'npm' && isNpmInstalled !== null && !ignoreCache)
+    return isNpmInstalled;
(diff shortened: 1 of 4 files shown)


src/renderer/npm.ts
```diff
@@ -32,23 +33,44 @@ const isUnique = (item: any, idx: number, arr: Array<any>): boolean => {
...
+    console.warn(`getIsYarnInstalled: "${command}" failed.`, error);
+    return (isYarnInstalled = false);
+  }
+}
+
 /**
  * Checks if npm is installed by checking if a binary
  * with that name can be found.
  */
 export async function getIsNpmInstalled(
```
erickzhao, line 61: Instead of making a new function, we could also keep the existing one and pass in `npm` or `yarn` as an arg, since the logic for checking if a package is installed is the same for both.

HashimotoYT: Well seems reasonable, done in 0eec84c3ea7a133dfb89b51220be272febb4fe81

src/renderer/npm.ts
```diff
@@ -33,22 +34,37 @@ const isUnique = (item: any, idx: number, arr: Array<any>): boolean => {
...
+  if (packageManager === 'npm' && isNpmInstalled !== null && !ignoreCache)
+    return isNpmInstalled;
+  if (packageManager === 'yarn' && isYarnInstalled !== null && !ignoreCache)
+    return isYarnInstalled;
 
-  const command = process.platform === 'win32' ? 'where.exe npm' : 'which npm';
+  const command =
+    process.platform === 'win32'
+      ? `where.exe ${packageManager}`
+      : `which ${packageManager}`;
```
malept, line 52: Perhaps you can use `hasYarn` from [`yarn-or-npm`](https://www.npmjs.com/package/yarn-or-npm) (which Electron Forge already uses).

Alternatively, perhaps use the [`which`](https://www.npmjs.com/package/which) module.

PR #470, opened 2020-08-27
feat: system-synced dark mode

erickzhao:
This PR adds a new setting that syncs your Fiddle theme with your OS. It's toggled on by default and will alternate between the default dark and light themes depending on your OS setting.

_(Please test this on Windows because I haven't yet 😇 )_

screencast 2020-08-28 15-52-21

### DIFF
--- a/src/main/system-theme.ts
+++ b/src/main/system-theme.ts
@@ -9,13 +9,15 @@ export function setupSystemTheme() {
     maybeUpdateTheme();
   });
 
-  ipcMainManager.on(IpcEvents.ERICK, (_event, source) => {
+  ipcMainManager.on(IpcEvents.SET_THEME_SOURCE, (_event, source) => {
     nativeTheme.themeSource = source;
   });
 }
 
 export function maybeUpdateTheme() {
   if (nativeTheme.themeSource === 'system') {
-    ipcMainManager.send(IpcEvents.ERICK, [nativeTheme.shouldUseDarkColors]);
+    ipcMainManager.send(IpcEvents.SET_SYSTEM_THEME, [
+      nativeTheme.shouldUseDarkColors,
+    ]);
   }
 }
--- a/src/main/system-theme.ts
+++ b/src/main/system-theme.ts
@@ -15,7 +15,7 @@ export function setupSystemTheme() {
     ]);
   });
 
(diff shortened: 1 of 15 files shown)


src/renderer/state.ts
```diff
@@ -808,7 +834,8 @@ export class AppState {
       | null
       | boolean,
   ) {
-    if (value) {
+    // null or undefined
+    if (value != null) {
```
erickzhao, line 838: This is the one unrelated change I added. Previously, non-null falsy values would never get saved to localStorage.

src/main/system-theme.ts
```diff
@@ -0,0 +1,23 @@
...
+  nativeTheme.on('updated', () => {
+    maybeUpdateTheme();
+  });
+
+  ipcMainManager.on(IpcEvents.SET_THEME_SOURCE, (_event, source) => {
+    nativeTheme.themeSource = source;
+  });
+}
+
+export function maybeUpdateTheme() {
```
HashimotoYT, line 17: really need export? never used anywhere rather that file

src/renderer/components/settings-general-appearance.tsx
```diff
@@ -96,6 +106,11 @@ export class AppearanceSettings extends React.Component<
       this.setState({ themes, selectedTheme });
     });
 
+    autorun(async () => {
```
HashimotoYT, line 109: What's the difference here when use `autorun`? It's something that I never see before so yeah, that looks strange for me.

vhashimotoo (dismissed):
Looks interesting some questions:

Should be this enabled by default from start? We use the default theme from start, and now we change to the system default theme. Maybe leave that disabled by default.

Not sure if this bug, but when you select light theme but system default is dark the first click to the checkbox doesn't change the theme to dark:

recording

src/renderer/state.ts
felixrieseberg, line 838:
```suggestion
    if (value !== null && value !== undefined) {
```

erickzhao:
@felixrieseberg now that you mention it, CSS might be a better approach here because it wouldn't require any main process work or IPC, and might have a lower maintenance burden. I think I tunnelled into the Electron-specific solution really quickly, but I'll take a closer look.

PR #481, opened 2020-09-15
feat: show a link to the comparison range on bisect complete

nornagon:
For easy clickin'!

Screenshot_9_18_20__10_43_AM

### DIFF
--- a/src/renderer/components/commands-bisect.tsx
+++ b/src/renderer/components/commands-bisect.tsx
@@ -29,12 +29,24 @@ export class BisectHandler extends React.Component<BisectHandlerProps> {
 
       const [minRev, maxRev] = response;
       const [minVer, maxVer] = [minRev.version, maxRev.version];
-      const message = `Check between versions ${minVer} and ${maxVer}.`;
+      const message = (
+        <>
+          Check the range{' '}
+          <a
+            target="_blank"
+            rel="noreferrer"
+            href={`https://github.com/electron/electron/compare/v${minVer}...v${maxVer}`}
+          >
+            {minVer}...{maxVer}
+          </a>
+          .
+        </>
+      );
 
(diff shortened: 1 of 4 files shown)


codebytere:
@nornagon do you have a screenshot of the new UI?

codebytere (commented):
@nornagon do you have a screenshot of the new UI?

This is also struggling here:

```
FAIL tests/renderer/components/commands-spec.tsx
  ● Test suite failed to run
    src/renderer/components/commands-bisect.tsx:49:9 - error TS2322: Type 'Element' is not assignable to type 'string'.
    49         label: <>Bisect complete. {message}</>,
               ~~~~~
      src/interfaces.ts:60:3
        60   label: string;
             ~~~~~
        The expected type comes from property 'label' which is declared here on type 'GenericDialogOptions'
```

nornagon:
@codebytere screenshot added to description & tests fixed.

codebytere (approved):

src/renderer/components/commands-bisect.tsx
```diff
@@ -29,12 +29,24 @@ export class BisectHandler extends React.Component<BisectHandlerProps> {
...
+          </a>
+          .
+        </>
+      );
 
       appState.pushOutput(`[BISECT] Complete: ${message}`);
       appState.setGenericDialogOptions({
         type: GenericDialogType.success,
-        label: `Bisect complete. ${message}`,
+        label: <>Bisect complete. {message}</>,
```
HashimotoYT, line 49: I think for more organizing we can move this "Bisect complete." also inside the `message` constant, and just return the single constant with all inside. This also should help to remove the second `React.Fragment`.

HashimotoYT (approved):
Looks cool, one small improvement.

Issue #479, opened 2020-09-15
Cmd+A doesn't work in text fields


nornagon:
Version 0.16.0 (9.1.1)
macOS 10.15.3

To repro:
1. Click any text field (e.g. the fiddle URL box, or the search box in the version selector)
2. Type some text so the field isn't empty
3. Press Cmd+A

Expected behavior:
All the text in the field should be selected

Actual behavior:
Selection does not change

codebytere:
Looks like this is broken here: https://github.com/electron/fiddle/blob/7bab1578636894925dcc227148439f6848d3a7e4/src/main/menu.ts#L268-L276

PR #483, opened 2020-09-16
fix: selectAll should work in text fields

codebytere:
Closes https://github.com/electron/fiddle/issues/479.

Also sends the action to the first responder so that this also works in the version picker and gist text field.

cc @nornagon @erickzhao @HashimotoYT


HashimotoYT (approved):
yeah
alt: nothing to say, just 👍

Issue #482, opened 2020-09-15
Bisect misunderstands ordering between stable, nightly and beta


nornagon:
Fiddle version: 0.16.0

To repro:
1. Enable bisect button
2. Click bisect
3. Select 10.0.0-nightly.20200209 as the "known good" version
4. Open the "known bad" version selector

Expected behavior:
It should be possible to select a beta version, e.g. 10.0.0-beta.23, as the "known bad" version, as all betas are logically later than all nightlies.

Actual behavior:
All the 10.0.0-beta.* versions are greyed out and unselectable.

PR #485, opened 2020-09-18
fix: version prerelease sorting

codebytere:
Closes https://github.com/electron/fiddle/issues/482.

The issue opened by @nornagon exposed a greater issue, which was that we weren't previously sorting prerelease tags at all in our logic. This meant that for a given major version the stable tags would correctly show at the top, but they would be followed by nightly tags instead of beta tagged version, which is incorrect since beta are strictly newer than nightlies. 

Semver does not handle this on its own: https://www.npmjs.com/package/semver#prerelease-tags

As such this PR seeks to address that by adding extra logic to `sortedElectronMap` to ensure prerelease tags sort properly as well as a test to ensure that's the case. Also tested with @nornagon's repro.

<details>
<summary>Before</summary>
Screen Shot 2020-09-17 at 11 10 03 PM
</details>

<details>
<summary>After</summary>
Screen Shot 2020-09-17 at 11 11 53 PM
</details>

cc @HashimotoYT @felixrieseberg

### DIFF
--- a/src/utils/sorted-electron-map.ts
+++ b/src/utils/sorted-electron-map.ts
@@ -20,12 +20,28 @@ export function sortedElectronMap<T>(
         return a.localeCompare(b);
       }
 
-      if (!semver.valid(a)) {
-        return -1;
-      }
+      if (!semver.valid(a)) return -1;
+      if (!semver.valid(b)) return 1;
 
-      if (!semver.valid(b)) {
-        return 1;
+      // Handle prerelease tag sorting - semver does not handle this.
+      // See https://www.npmjs.com/package/semver#prerelease-tags for more.
+      const bothPre = semver.prerelease(a) && semver.prerelease(b);
+      const sameCoerced =
+        semver.valid(semver.coerce(a)) === semver.valid(semver.coerce(b));
+      if (bothPre && sameCoerced) {
+        const [tagA, valueA] = semver.prerelease(a)!;
+        const [tagB, valueB] = semver.prerelease(b)!;
+
+        if (tagA === 'beta' && tagA === tagB) {
+          // Both are betas - compare beta number.
+          return valueA > valueB ? -1 : 1;
+        } else if (tagA === 'nightly' && tagA === tagB) {
+          // Both are nightlies - compare nightly dates strings.
+          return -`${valueA}`.localeCompare(`${valueB}`);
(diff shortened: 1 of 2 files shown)


src/utils/sorted-electron-map.ts
```diff
@@ -20,12 +20,27 @@ export function sortedElectronMap<T>(
...
 
-      if (!semver.valid(a)) {
-        return -1;
-      }
+      if (!semver.valid(a)) return -1;
+      if (!semver.valid(b)) return 1;
+
+      // Handle prerelease tag sorting - semver does not handle this.
+      // See https://www.npmjs.com/package/semver#prerelease-tags for more.
+      if (semver.prerelease(a) && semver.prerelease(b)) {
```
nornagon, line 28:
```suggestion
      if (semver.major(a) === semver.major(b) && semver.prerelease(a) && semver.prerelease(b)) {
```

nornagon: (maybe check minor/patch versions too? i know we don't currently have any prerelease versions that aren't `x.0.0`, but we might in future have `x.1.0-beta.1` or something so it'd be best to compare the full version here i think)

src/utils/sorted-electron-map.ts
```diff
@@ -20,12 +20,27 @@ export function sortedElectronMap<T>(
...
 
-      if (!semver.valid(b)) {
-        return 1;
+        if (semver.major(a) === semver.major(b)) {
+          if (tagA === 'beta' && tagA === tagB) {
+            // Both are betas - compare beta number.
+            return valueA > valueB ? -1 : 1;
+          } else if (tagA === 'nightly' && tagA === tagB) {
+            // Both are nightlies - compare nightly date.
+            return new Date(valueA) > new Date(valueB) ? -1 : 1;
```
nornagon, line 38: no need to convert these into dates as the strings themselves will sort the right way.
```suggestion
            return valueA.localeCompare(valueB);
```

src/utils/sorted-electron-map.ts
codebytere: This gets checked on line 32 - happy to move it but logically this should already work as expected?

nornagon (approved):

HashimotoYT (approved):

Issue #487, opened 2020-09-18
The publish button on existed gist just delete the gist


HashimotoYT:
Ref #484 

![recording (1)](https://user-images.githubusercontent.com/24681191/93615977-c3200b80-f9dc-11ea-9b9c-e0378733fd67.gif)

/cc @codebytere

PR #488, opened 2020-09-18
fix: gist action state management

codebytere:
Closes https://github.com/electron/fiddle/issues/487.

Fixes an issue where if multiple gists are published on top of one another, the delete state is not properly updated. To clean this up a bit, i've also pulled out the various gist actions into their own functions. Also fixed several pre-existing issues with publishing state being persisted at the same time as updating state.

To test, I:
1. Published Fiddle 1 ✅ 
2. Edited and updated Fiddle 1 ✅
3. Published Fiddle 2 ✅ 
4. Deleted Fiddle 1 ✅ 
5. Updated Fiddle 2 ✅ 
6. Loaded and deleted Fiddle 2 ✅ 

Which i _believe_ should cover all state transitions.

cc @HashimotoYT

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -102,105 +102,136 @@ export class GistActionButton extends React.Component<
   }
 
   /**
-   * Connect with GitHub, publish the current Fiddle as a gist,
-   * and update all related properties in the app state.
+   * Publish a new GitHub gist.
    */
-  public async performGistAction(): Promise<void> {
-    console.log('publishOrUpdateFiddle()');
+  public async handlePublish() {
     const { appState } = this.props;
-    const { actionType } = this.state;
-    appState.isPublishing = true;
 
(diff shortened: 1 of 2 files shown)


felixrieseberg (approved):

HashimotoYT (changes requested):
Seems working fine, just one case where something goes not okay:

1. Load the non-you gist: https://gist.github.com/pacificocean-bot/1021a526610a8662ba6c404b5aeda85c
2. Try to delete, you should engage error about not your gist
3. The load gist bar and buttons reset to the default state.

This seems should be doing the same thing what's done for "Update" where the state is saved but nothing more happens.

codebytere:
Ah hmm yeah if you paste in a new one then the `gistId` isn't set in state until we run `Load Fiddle` 🤔

HashimotoYT (approved):
I think we can discuss the better error in some other time. Currently, that looks fine to me

PR #491, opened 2020-09-21
refactor: clean up gist action states

codebytere:
Fiddle can't perform more than one action at a given time, so it makes more sense to place all potential actions in one state variable.

We also had some state variables in the Mobx store and some not, and as we should disable the command bar during _any_ action it makes the most sense to place this overarching state variable in the Mobx store.

cc @felixrieseberg


erickzhao (approved):
Looks good!

Issue #477, opened 2020-09-14
It should be possible to set a custom Gist description


codebytere:
Right now it's hardcoded to 'Electron Fiddle Gist' which is pretty annoying when there are hundreds of them. 

I'll take this one if you all think it's a good idea!

erickzhao:
I'm definitely a 👍  on this.

On a related note, since we use [namor](https://www.npmjs.com/package/namor) to generate the Fiddle app title, we could set the Gist's title to that to at least have some sort of uniqueness to the Gist names even if a custom title is not set,

codebytere:
@erickzhao ooh as a stopgap that sounds great to me for now to set as a default! Wrangling custom descriptions will require some new UI that'll probably require some more thought 🤔

PR #492, opened 2020-09-25
feat: allow setting gist description

codebytere:
Closes https://github.com/electron/fiddle/issues/477.

This introduces a means to set a custom description for each Gist that's published to GitHub. This is useful when creating many gists since they're hard to differentiate with the same name.

Screen Shot 2020-09-29 at 8 01 29 PM

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -14,6 +14,7 @@ import * as React from 'react';
 import { when } from 'mobx';
 import {
   EditorValues,
+  GenericDialogType,
   GistActionState,
   GistActionType,
 } from '../../interfaces';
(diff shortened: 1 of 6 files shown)


erickzhao (commented):
I feel like this UI is good enough to ship -- I don't have any issues with the way it's handled!

codebytere:
@erickzhao should be good to go!

src/renderer/components/commands-action-button.tsx
```diff
@@ -105,6 +106,26 @@ export class GistActionButton extends React.Component<
...
+      label: 'Please provide a brief description for your Fiddle Gist',
+      cancel: 'Skip',
+      wantsInput: true,
+      placeholder: 'Electron Fiddle Gist',
+    });
+
+    appState.isGenericDialogShowing = true;
+    await when(() => !appState.isGenericDialogShowing);
+
+    return appState.genericDialogLastInput ?? 'Electron Fiddle Gist';
```
erickzhao, line 126: The Okay button here does the same thing as the Skip one. 

If we want a true skip, this should also short-circuit to `Electron Fiddle Gist` if  `appState.genericDialogLastResult` is false.

However, I'd personally be okay with not having a `Skip` button altogether, and just publishing the placeholder if someone presses "Okay" without entering a string.

src/renderer/components/commands-action-button.tsx
```diff
@@ -105,6 +106,26 @@ export class GistActionButton extends React.Component<
     }
   }
 
+  public async getFiddleDescriptionFromUser(): Promise<string> {
+    const { appState } = this.props;
+
+    // Reset potentially non-null last description.
+    appState.genericDialogLastInput = null;
+
+    appState.setGenericDialogOptions({
```
erickzhao, line 115: Maybe set `ok: 'Publish'` so that the copy is clearer in the dialog.

src/renderer/components/dialog-generic.tsx
```diff
@@ -34,8 +38,16 @@ export class GenericDialog extends React.Component<GenericDialogProps> {
...
+    const {
+      type,
+      ok,
+      cancel,
+      label,
+      wantsInput,
+      placeholder,
+    } = genericDialogOptions;
 
+    const shouldShowInput = wantsInput && placeholder;
```
erickzhao, line 50: Would it make sense to make this placeholder optional?

erickzhao (changes requested):

erickzhao (approved):

PR #495, opened 2020-10-02
feat: add ability to change electron download binary path

### LINKED ISSUE #310: Allow user to change install dir of Electron binaries
Just why not™

vhashimotoo:
This PR adds the ability to change the download directory for launching binary files from the default directory to any.

### Screenshot of the settings menu

image

Fixes #310

### DIFF
--- a/src/renderer/binary.ts
+++ b/src/renderer/binary.ts
@@ -4,7 +4,6 @@ import * as path from 'path';
 import { VersionState } from '../interfaces';
 import { fancyImport } from '../utils/import';
 import { normalizeVersion } from '../utils/normalize-version';
-import { USER_DATA_PATH } from './constants';
 import { removeTypeDefsForVersion } from './fetch-types';
 import { AppState } from './state';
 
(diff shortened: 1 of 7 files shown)


src/renderer/binary.ts
```diff
@@ -233,8 +235,8 @@ async function download(appState: AppState, version: string): Promise<string> {
  * @param {string} version
  * @returns {string}
  */
-function getDownloadPath(version: string): string {
-  return path.join(USER_DATA_PATH, 'electron-bin', version);
+function getDownloadPath(downloadPath: string, version: string): string {
```
erickzhao, line 238: Is there a case where we don't want to use `appState.downloadBinaryPath` for the `downloadPath` param? Can we just hardcode the variable inside `path.join`?

erickzhao (commented):
Design question: do we want to add a note saying that users will have to redownload their Electron binaries when switching paths, or should we automatically move the old files over?

Issue #322, opened 2020-02-11
Clicking "esc" in Theme search filter closes the preferences window


vhashimotoo:
Repro:

1. Go to preferences -> general
2. Click the "Choose your theme" box
3. Start searching for something
4. Drop your plans and click "esc" to close this box.

Maybe something for newcomers, just need to disable "esc" when we in the input box.

mwoodhousemckenzie:
:wave: I took a stab at fixing this problem, I am new to OSS contributions so I am not sure if I went about this correctly 😅

erickzhao:
@mwoodhousemckenzie I'll take a look at this tomorrow!

Issue #449, opened 2020-07-27
Intermittent Coveralls failures on CI


erickzhao:
### Problem

Coveralls will frequently fail on CI, leading to blocked pull requests. This seems to be an upstream issue that we've been dealing with for the last few months.

Ref https://github.com/lemurheavy/coveralls-public/issues/1427
Ref https://github.com/lemurheavy/coveralls-public/issues/750

This error will look something like this in the TravisCI logs:
```shell
0.46s$ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then npm run test:coverage; fi
> electron-fiddle@0.15.1 test:coverage /home/travis/build/electron/fiddle
> cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
[error] "2020-07-27T00:17:32.111Z"  'error from getOptions'
/home/travis/build/electron/fiddle/node_modules/coveralls/bin/coveralls.js:19
      throw err;
      ^
Error: Command failed: git cat-file -p f540f5d7d65076553db0c902506d2945b3d1b09b
fatal: Not a valid object name f540f5d7d65076553db0c902506d2945b3d1b09b
    at ChildProcess.exithandler (child_process.js:303:12)
    at ChildProcess.emit (events.js:315:20)
    at maybeClose (internal/child_process.js:1021:16)
    at Socket.<anonymous> (internal/child_process.js:443:11)
    at Socket.emit (events.js:315:20)
    at Pipe.<anonymous> (net.js:674:12) {
  killed: false,
  code: 128,
  signal: null,
  cmd: 'git cat-file -p f540f5d7d65076553db0c902506d2945b3d1b09b'
}
```

### Workaround

If your PR fails, you can force-push to re-trigger the build.

```shell
git checkout my-failing-branch
git commit --amend # no changes needed to the message; this is just to change the commit hash
git push --force-with-lease
```

Note that these steps don't guarantee that Coveralls will pass. You might need to repeat a few times to get a pass.

erickzhao:
We switched to GitHub Actions in https://github.com/electron/fiddle/pull/504 and haven't encountered the failure in the last month or so. Will close this issue for now.

PR #498, opened 2020-10-05
fix: theme select popover closing settings when pressing escape

mwoodhousemckenzie:
Closes https://github.com/electron/fiddle/issues/322

I added a state listener for when someone opens the theme selector. I could not figure out a way to stop the events from propagating and closing the settings besides this solution. It might be better to make this more generic of `prevenSettingsClose` or something, but I am not sure.

ezgif com-optimize

I saw that this is a good first issue to test my OSS contributions on, I am mainly an Angular developer and have only dabbled a little in React so if there is a better way to make these changes, please let me know.

### DIFF
--- a/src/renderer/components/settings-general-appearance.tsx
+++ b/src/renderer/components/settings-general-appearance.tsx
@@ -59,6 +59,7 @@ export const renderItem: ItemRenderer<LoadedFiddleTheme> = (
 
 export interface AppearanceSettingsProps {
   appState: AppState;
+  toggleHasPopoverOpen: () => void;
 }
 
 export interface AppearanceSettingsState {
@@ -185,14 +186,15 @@ export class AppearanceSettings extends React.Component<
             itemPredicate={filterItem}
             onItemSelect={this.handleChange}
             popoverProps={{
-              onClosed: () => this.toggleThemeSelectorOpened(),
+              onClosed: () => this.props.toggleHasPopoverOpen(),
             }}
             noResults={<MenuItem disabled={true} text="No results." />}
           >
             <Button
+              id="open-theme-selector"
               text={selectedName}
               icon="tint"
-              onClick={() => this.toggleThemeSelectorOpened()}
+              onClick={() => this.props.toggleHasPopoverOpen()}
             />
           </ThemeSelect>
         </FormGroup>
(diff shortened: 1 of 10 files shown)


erickzhao (changes requested):
Hey @mwoodhousemckenzie, thanks for the contribution! I think you're on the right track here,  and here's a proposal to generalize that involves moving some functionality around:

* In `src/renderer/components/settings.tsx`, set a local state variable in the component that's something like `hasPopoverOpen` (similar to the `appState.isThemeSelectorShowing` variable).

* Create a helper function `toggleHasPopoverOpen` in `src/renderer/components/settings.tsx` (similar to the `toggleThemeSelectorOpened()` function, but modifies the local state variable instead).

* Pass `toggleHasPopoverOpen` down as a prop to `src/renderer/components/settings-general-appearance.tsx` and use it in the same way as your previous function.

This should generalize it so that if any other settings page has a Popover with the same kind of interaction, we should be able to program the same state variable accordingly.

mwoodhousemckenzie:
@erickzhao Perfect. Thanks for the feedback! I was wondering if it should be more generic. I updated the code to be more generic now 😄

erickzhao (changes requested):
Hey @mwoodhousemckenzie, this looks great!

One thing to change that's completely not your fault: the listener for the `Esc` key actually never unmounts because I didn't pass it the correct (`true`) argument when initially adding this feature. This makes it so that your fix doesn't actually work for subsequent instances of the Settings pane.

Could you change line 56 of `src/renderer/components/settings.tsx` to this?

```diff
  public componentWillUnmount() {
-   window.removeEventListener('keyup', this.closeSettingsPanel);
+   window.removeEventListener('keyup', this.closeSettingsPanel, true);
  }
```

mwoodhousemckenzie:
Ah good catch @erickzhao, don't know why I didn't notice that. Thanks for pointing it out!

erickzhao (approved):

mwoodhousemckenzie:
> @mwoodhousemckenzie could you follow the instructions at #449 and force push until you get the Travis build to pass? I can't merge the PR in until it passes.
> 
> Sorry about this, it's an upstream issue that happens a bit too often. 🙇

Travis build has passed but the coveralls and circlici builds are failing, is there anything I can do to get those to pass?

erickzhao:
> Travis build has passed but the coveralls and circlici builds are failing, is there anything I can do to get those to pass?

You don't have to worry about those for the purposes of this PR:
* You could add unit tests for Coveralls to cover more lines of code, but this is very optional.
* Our CircleCI has been a WIP for a while now, and can safely be ignored.

I'll merge this in soon unless you feel like adding more tests!

mwoodhousemckenzie:
Unless there is any tests you see I should add I am good with this. Thanks for helping me on my first PR in this repo, greatly appreciated 😄

PR #360, opened 2020-03-23
build: Switch from AppVeyor/Travis CI to CircleCI

malept:
Attempt 2 for #246.

### TODO

* [ ] Figure out why Squirrel signing is failing

PR #504, opened 2020-10-19
build: Use GitHub Actions to build

felixrieseberg:
This PR moves our build infrastructure from Travis & AppVeyor over to GitHub Actions. I have also updated our repository secrets with a new three-year certificate. I'm pretty confident in the code, it's already running various of my personal projects well.

This shouldn't change anything, but we should check that everything got correctly signed the next time we release a new version.


erickzhao (approved):
Seems like all the checks are green ✅ 

Other TODO (could wait for later):
- Update badges in README
- Handle Coveralls for code coverage (used to be in TravisCI)
- Remove CircleCI/Travis/Appveyor checks (cc @malept does this PR overtake #360?)

malept:
>     * Remove CircleCI/Travis/Appveyor checks (cc @malept does this PR overtake #360?)

Yes, and also GitHub Actions is way better IMO

.github/workflows/build.yaml
```diff
@@ -0,0 +1,114 @@
...
+        continue-on-error: true
+        id: write_file
+        uses: timheuer/base64-to-file@v1
+        with:
+          fileName: 'win-certificate.pfx'
+          encodedString: ${{ secrets.WINDOWS_CODESIGN_P12 }}
+      - name: Install
+        run: yarn
+      - name: Git Submodules
+        run: git submodule update --init --recursive
```
malept, line 81: I didn't realize Fiddle used git submodules...

malept (commented):
For what it's worth, I have been working on Actions to make building Electron Forge projects in GitHub Actions way simpler, based on conversations I've had with Felix regarding one of his more recent projects. I can make a PR against this one to show what the changes would be, if y'all want. Or since I won't be able to get to it until late tonight, I can just make a normal PR after this is merged.

.github/workflows/build.yaml
felixrieseberg: Oh, it doesn't, that's a copy & paste fragment. Thanks for catching.

codebytere (approved):

Issue #282, opened 2019-12-13
Tick the open Fiddle in “Show Me > Electron APIs”


deadcoder0904:
I get confused every time which Fiddle I opened so it’d be great if a checkmark is provided beside the Electron APIs which I opened the last time

PR #507, opened 2020-10-21
fix: Add Tick to Show Me > Electron APIs

JordanTroutman:
Addressing Issue #282.  I added functionality to put a checkmark that keeps track of the current open Fiddle in `Show Me > Electron APIs`.  Each item within the `MenuItem` has the `radio` type.


erickzhao (changes requested):
Hey @JordanTroutman!

So this is a good start, but the `radio` MenuItem type unfortunately doesn't fulfill all of our requirements here. A few acceptance criteria that need to be done:

1) There should be no active checkbox when no Demo is selected.

Unfortunately, this is a limitation in the Chromium model for the `radio` model (see https://github.com/electron/electron/issues/21292#issuecomment-561923188), so this means that the checkboxes need to be managed programatically.

2) When loading another Fiddle, the checkboxes should be reset.

If we load a Fiddle from another source (remote Gist, New Fiddle, or local folder), the ticked MenuItem should be cleared.

3) When loading more than 1 window, the checkboxes should be adjusted accordingly.

Let's say we have two Electron Fiddle windows opened, one of them on the Debugger demo and one on the TouchBar demo, swapping between windows should tick the correct MenuItem for each window.

src/main/menu.ts
```diff
@@ -150,9 +150,12 @@ function getTasksMenu(): MenuItemConstructorOptions {
 function getShowMeMenuItem(
   key: string,
   item: string | Templates,
+  active_key: string | null,
```
vhashimotoo, line 153: Let's use camelCase

```suggestion
  activeKey: string | null,
```

src/main/menu.ts
```diff
@@ -150,9 +150,12 @@ function getTasksMenu(): MenuItemConstructorOptions {
...
 ): MenuItemConstructorOptions {
   if (typeof item === 'string') {
     return {
+      type: 'checkbox',
+      checked: key == active_key,
```
vhashimotoo, line 158: Why we can't use `===` here?

```suggestion
      checked: key == activeKey,
```

src/main/menu.ts
```diff
@@ -161,15 +164,15 @@ function getShowMeMenuItem(
   return {
     label: key,
     submenu: Object.keys(item).map((subkey) => {
-      return getShowMeMenuItem(subkey, item[subkey]);
+      return getShowMeMenuItem(subkey, item[subkey], active_key);
```
vhashimotoo, line 167:
```suggestion
      return getShowMeMenuItem(subkey, item[subkey], activeKey);
```

src/main/menu.ts
```diff
@@ -161,15 +164,15 @@ function getShowMeMenuItem(
...
     }),
   };
 }
 
-function getShowMeMenu(): MenuItemConstructorOptions {
+function getShowMeMenu(active_key: string | null): MenuItemConstructorOptions {
```
vhashimotoo, line 172:
```suggestion
function getShowMeMenu(activeKey: string | null): MenuItemConstructorOptions {
```

src/main/menu.ts
```diff
@@ -161,15 +164,15 @@ function getShowMeMenuItem(
...
   const showMeMenu: Array<MenuItemConstructorOptions> = Object.keys(
     SHOW_ME_TEMPLATES,
-  ).map((key) => getShowMeMenuItem(key, SHOW_ME_TEMPLATES[key]));
+  ).map((key) => getShowMeMenuItem(key, SHOW_ME_TEMPLATES[key], active_key));
```
vhashimotoo, line 175:
```suggestion
  ).map((key) => getShowMeMenuItem(key, SHOW_ME_TEMPLATES[key], activeKey));
```

src/main/menu.ts
```diff
@@ -248,7 +251,7 @@ function getFileMenu(): MenuItemConstructorOptions {
 /**
  * Creates the app's window menu.
  */
-export function setupMenu() {
+export function setupMenu(active_key: string | null = null) {
```
vhashimotoo, line 254:
```suggestion
export function setupMenu(activeKey: string | null = null) {
```

src/main/menu.ts
```diff
@@ -331,7 +334,7 @@ export function setupMenu() {
 
   menu.splice(process.platform === 'darwin' ? 1 : 0, 0, getFileMenu());
 
-  menu.splice(menu.length - 1, 0, getTasksMenu(), getShowMeMenu());
+  menu.splice(menu.length - 1, 0, getTasksMenu(), getShowMeMenu(active_key));
```
vhashimotoo, line 337:
```suggestion
  menu.splice(menu.length - 1, 0, getTasksMenu(), getShowMeMenu(activeKey));
```

src/main/menu.ts
vhashimotoo: Also since `activeKey` is unused under `index.ts`, why not use `activeKey?: string | null`

vhashimotoo (commented):
Since I'm not really known the history of the request so I'm just reporting some additions and bug report:

When you open two windows and choose on the first some another example, and on the second another example that not on window 1, you get that both windows have selected examples from 1 or 2.

JordanTroutman:
@vhashimotoo Can you explain a little more of exactly what you tried? I just opened two windows and clicked a different example for each window.  Focusing on a window ticked the appropriate example.

vhashimotoo:
This seems reproduced very weirdly, you can check the gif below. You need to have two windows and on the first window select example, second shouldn't take the value of the first, but if you select example on the second window the value is changed also in the first window.

recording

JordanTroutman:
electron-gif
This is what it looked like on my Macbook...is there a difference between OS when using menus?

erickzhao:
@JordanTroutman I can also repro on my Windows VM. I'll try to investigate a bit deeper soon.

src/main/windows.ts
```diff
@@ -54,6 +54,19 @@ export function createMainWindow(): Electron.BrowserWindow {
     browserWindow = null;
   });
 
+  // Focusing on new window loads the possibly clicked "Show Me" fiddle
+  browserWindow.on('focus', () => {
+    browserWindow?.webContents.send(IpcEvents.MENU_IPC_RENDER);
```
erickzhao, line 59: Can we use the `IpcMainManager` and `IpcRendererManager` modules to handle IPC events?

src/main/windows.ts
```diff
@@ -54,6 +54,19 @@ export function createMainWindow(): Electron.BrowserWindow {
...
+  });
+
+  // Show menu for respective window taking into consideration an entry in
+  // "Show Me" may be selected
+  ipcMain.on(IpcEvents.MENU_IPC_MAIN, (_, result) => {
```
erickzhao, line 64: This IPC event will duplicate on each window creation. Should probably go elsewhere.

erickzhao (changes requested):
A few general comments, unrelated to the Windows bug pointed above.

erickzhao:
superseded by #705

PR #512, opened 2020-10-22
fix: various automagic npm module installation fixes

### LINKED ISSUE #509: commented npm modules not ignored if comments are multi-line
Given this code in an editor pane:

```js
    // const cow = require('cow');
    /* const say = require('say'); */
    /**
     * const hello = require('hello'); 
     * const world = require('world'); 
    */
```

Fiddle will still attempt to install the `say`, `hello`, and `world` packages from NPM. This is because we only check if lines start with `//` to check if they're commented.

### LINKED ISSUE #510: npm modules installing multiple times if required in multiple editors
Given the below code:

```js
// main.js
const log = require('electron-log');
```

```js
// renderer.js
const log = require('electron-log');
```

Fiddle will attempt to `npm install` the `electron-log` package twice, once for each file it's in.

### LINKED ISSUE #511: npm modules don't load on preload editor
Given this code:

```js
// preload.js
const _ = require('lodash');
```

Fiddle will not attempt to install the `electron-log` module, leading to the following error in the renderer console:

```
Error: Cannot find module 'lodash'
```

erickzhao:
This PR has 3 bugfixes relating to the way we import `npm` modules. Each is in a separate commit.

* 5e3a588 Fixes #509 by using the `decomment` module inside a Web Worker thread.
* 57aa3d1 Fixes #510 by adding the list of modules detected into a `Set()`.
* 7ebbeac Fixes #511 by adding `preload` to the list of editors that we check modules for.
* 93c24dd adds tests to make sure everything works. :)

**Note**: this PR adds in a new dependency to the [`decomment`](https://github.com/vitaly-t/decomment) module, which uses the [Esprima](https://esprima.org/) JavaScript parser under the hood to strip any JS code fed in of its comments. I decided to use a library rather than play around with regexes for reliability's sake. `decomment` is licensed under MIT, so it should be compatible with Electron Fiddle for licensing purposes.


felixrieseberg:
Big fan, but I'm pretty unsure about including `decomment`, which bundles `esprima`. That's _ten Reacts_ in terms of size 😬

erickzhao:
@felixrieseberg I agree that it's a pretty big package to fix an edge case in our code, but if we want to provide a stable experience for users to `require` packages automatically, using a parser rather than regex to detect these module installs should be the way to go, even if it's to solve an edge case.

Are there big performance impacts at this bundle level for loading the local bundle? For reference, `decomment` seems to be [135kb minified](https://bundlephobia.com/result?p=decomment@0.9.3) while `react-dom` is [121kb](https://bundlephobia.com/result?p=react-dom@17.0.1).

As for alternatives, I'm not too familiar with using different parsers, but we could probably shave off 25kb here by using [`acorn`](https://bundlephobia.com/result?p=acorn@8.0.4) or [`meriyah`](https://bundlephobia.com/result?p=meriyah@3.0.3) directly instead. I'm not sure we can decrease that any further.

Maybe we can lazy load the module rather than import at the top of the file?

issacgerges:
Its unfortunate you can't lean on the tokenizer that's being used internally by monaco. I played around with this a bit but wasn't able to get access to the AST https://github.com/microsoft/monaco-typescript/pull/65

src/renderer/npm.ts
```diff
@@ -162,3 +160,16 @@ export function packageRun(
 ): Promise<string> {
   return exec(dir, `${packageManager} run ${command}`);
 }
+
+export function decommentWithWorker(input: string): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const worker = new Worker('../utils/decomment.ts');
```
felixrieseberg, line 166: My only concern is that our build process will somehow destroy this. Did you run `npm run build` and check that it still works post compilation?

felixrieseberg (approved):
Good stuff! I'd only check whether this survives compilation but it looks fantastic otherwise :clap:

src/renderer/npm.ts
erickzhao: `yarn package` seems to work fine for me. :)

PR #515, opened 2020-10-23
build: switch to use Electron Forge publish

vhashimotoo:
This PR switches to using just Electron Forge to publish command instead of action. Also moved to build ia32 to before the building x64, this should help by publishing the right `nupkg` files for updating x64.


vhashimotoo:
Replaced by #516

PR #524, opened 2020-11-05
fix: re-work output dom to remove whitespace when copying

### LINKED ISSUE #480: Copy/pasting lines from the console view adds extra whitespace
Fiddle version: 0.16.0
macOS 10.15.3

To repro:
1. Run any app
2. Select some lines in the console window and copy them
3. Paste them into another text field

Expected behavior:
Each line in the console window is pasted one after another as they're visible in the console window, without extra whitespace, eg.

```
Saved files to /var/folders/_7/jm5cr84j7jd4v7btf2d3bqwm0000gn/T/tmp-19301PgnXP2Cz02Iy
Electron v999.999.999 started.
[51298:0915/102615.051186:INFO:content_main_runner_impl.cc(993)] Chrome is running in full browser mode.
```

Actual behavior:
```
Saved files to /var/folders/_7/jm5cr84j7jd4v7btf2d3bqwm0000gn/T/tmp-19301PgnXP2Cz02Iy

Electron v999.999.999 started.

[51298:0915/102615.051186:INFO:content_main_runner_impl.cc(993)] Chrome is running in full browser mode.

```

issacgerges:
Fixes #480 

This PR moves the console from `<p>` elements per row, to `<span>`s wrapped in `<div>` to avoid additional newlines when copy pasting. I modeled this very loosely around the structure used by the chrome inspector console.

copy-paste

### DIFF
--- a/src/renderer/components/output.tsx
+++ b/src/renderer/components/output.tsx
@@ -88,10 +88,12 @@ export class Output extends React.Component<CommandsProps> {
       : {};
 
     return lines.map((text, lineIndex) => (
-      <p style={style} key={`${entry.timestamp}--${index}--${lineIndex}`}>
-        {timestamp}
-        {text}
-      </p>
+      <div key={`${entry.timestamp}--${index}--${lineIndex}`}>
+        <span style={style} className="output-message">
+          {timestamp}
+          {text}
+        </span>
+      </div>
     ));
   }
 
--- a/src/less/components/output.less
+++ b/src/less/components/output.less
@@ -1,4 +1,4 @@
-@import "../variables.less";
+@import '../variables.less';
 
 .output {
   font-size: 12px;
@@ -16,10 +16,7 @@
   overflow-y: scroll;
   -webkit-app-region: no-drag;
 
-  p {
-    margin-bottom: 0;
+  .output-message {
     white-space: pre;
-    padding-left: 5.6em;
-    text-indent: -5.6em;
   }
 }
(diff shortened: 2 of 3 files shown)


erickzhao (approved):
WFM, thanks @issacgerges!

src/renderer/components/output.tsx
```diff
@@ -85,13 +85,15 @@ export class Output extends React.Component<CommandsProps> {
     const lines = entry.text.split(/\r?\n/);
     const style: React.CSSProperties = entry.isNotPre
       ? { whiteSpace: 'initial' }
-      : {};
+      : { whiteSpace: 'pre' };
```
erickzhao, line 88: @issacgerges One comment: what does this line do? I tried manually switching the `initial` and `pre` values and didn't actually get any changes on my end.

screencast 2020-11-06 14-07-11

issacgerges: heya @erickzhao, `white-space: pre` tells the browser that line-breaking and whitespace are already "correct" in the output. With normal whitespace rules, the browser will break lines to fill the (layout) box and collapse repeated whitespace (`<span>     Hello</span>` would become `Hello`). With this rule it will only break lines when there are explicit linebreaks in the output and will preserve all whitespace.

In your example reducing the viewport size should show the text wrapping and not wrapping depending on that setting. Really glad you mentioned this though, helped me realized there was some CSS I had missed and could cleanup in this PR. Will send an update in a sec

codebytere (approved):
THANK YOU 🚀

erickzhao (approved):

PR #527, opened 2020-11-10
fix: spin while installing modules

codebytere:
Depends on https://github.com/electron/fiddle/pull/526 - should be rebased/merged only after that PR is merged.

Adds a spinner animation for Node.js module installs so it doesn't just freeze on run and give the user no UI indication of what's happening beyond console output.

cc @erickzhao @felixrieseberg


codebytere:
Rebased on https://github.com/electron/fiddle/pull/526 - should be good to go :)

felixrieseberg (approved):

PR #531, opened 2020-11-20
fix: repopulate URL bar on blur

### LINKED ISSUE #427: Add a reliable way to copy the current gist URL to clipboard
1. Open Fiddle
2. Publish
3. Delete the URL next to "Load Fiddle"
4. Make some changes and publish again
5. How do I copy the gist URL to the clipboard now?

When publishing/updating gists the url should be copied to the clipboard at the very least.

There should probably also be a menu item for copying the url to the current gist.

aryanshridhar:
Once the user manipulates the <code>address bar</code> value and publishes the fiddle again , the gist url is no longer available .

<strong>Fix</strong> :

Added a <code>onBlur</code> event to <code>address bar</code> , which loads back the original gist url if the user looses focus out of the bar .

Fixes #427

Would love an review @erickzhao :)

### DIFF
--- a/src/renderer/components/commands-address-bar.tsx
+++ b/src/renderer/components/commands-address-bar.tsx
@@ -31,6 +31,7 @@ export class AddressBar extends React.Component<
     super(props);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.handleChange = this.handleChange.bind(this);
+    this.handleBlur = this.handleBlur.bind(this);
     this.submit = this.submit.bind(this);
 
     const { gistId } = this.props.appState;
@@ -110,6 +111,15 @@ export class AddressBar extends React.Component<
     this.setState({ value: event.target.value });
   }
 
+  public handleBlur(_event: React.FocusEvent<HTMLInputElement>) {
+    const { gistId } = this.props.appState;
+    const url = urlFromId(gistId);
+
+    if (url) {
+      this.setState({ value: url });
+    }
+  }
+
   public renderLoadButton(isValueCorrect: boolean): JSX.Element {
     return (
       <Button
@@ -136,6 +146,7 @@ export class AddressBar extends React.Component<
             leftIcon="geosearch"
             intent={isCorrect || !value ? undefined : Intent.DANGER}
             onChange={this.handleChange}
+            onBlur={this.handleBlur}
             placeholder="https://gist.github.com/..."
             value={value}
             rightElement={this.renderLoadButton(isCorrect)}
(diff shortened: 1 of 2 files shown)


erickzhao (commented):
Hey @aryanshridhar, thanks for the contribution!

Seems like tests are failing right now because there's an outdated snapshot. Could you regenerate it?

erickzhao (changes requested):
Hey @aryanshridhar, thanks for the PR. UX-wise, I don't think this is the right approach to solving the problem.

Ideally, we would allow users to keep modifying the URL input, but add back the original gist URL if they edit the input but don't load anything new.

Haven't tried this code, but maybe a better solution would be to force the current Fiddle's `gist` value for the input's `onBlur` effect.

aryanshridhar:
Cool , Will fix this sooner

aryanshridhar:
Thanks for the review @erickzhao , I thought of similar idea before implementing this but was worried that it may create a sense of confusion (Since , minimizing the application also triggers the <code>onBlur</code> event ) . 
However , I have reflected the similar changes in the pr .

erickzhao (approved):
This code LGTM. :)

src/renderer/components/commands-address-bar.tsx
```diff
@@ -110,6 +111,15 @@ export class AddressBar extends React.Component<
     this.setState({ value: event.target.value });
   }
 
+  public handleBlur(event: React.FocusEvent<HTMLInputElement>) {
+    const { gistId } = this.props.appState;
+    const url = urlFromId(gistId);
+
+    if (event && url) {
```
erickzhao, line 118: @aryanshridhar:

Is `event` ever falsy here?

erickzhao (commented):
Actually, single comment

src/renderer/components/commands-address-bar.tsx
aryanshridhar: I don't think so , but since a event is being passed in the function parameter , I felt it better to pass it in the definition as well . (Will remove , if it doesn't feel right)

src/renderer/components/commands-address-bar.tsx
```diff
@@ -110,6 +111,15 @@ export class AddressBar extends React.Component<
...
+      this.setState({ value: url });
+    }
```
erickzhao, line 120:
```suggestion
  public handleBlur(_event: React.FocusEvent<HTMLInputElement>) {
    const { gistId } = this.props.appState;
    const url = urlFromId(gistId);

    if (url) {
      this.setState({ value: url });
    }
```

erickzhao (changes requested):
Yep, think it'd be best to remove the guard if `event` has no effect on the logic here. If the variable remains unused, prefix it with `_`.

erickzhao (approved):

PR #542, opened 2020-12-04
test: fix unhandled promise errors in state spec

erickzhao:
Seems like we were getting many `(node:2702) UnhandledPromiseRejectionWarning: TypeError: downloadingVersions is not iterable` errors when running `yarn test`, but these weren't picked up by Jest as test failures. I believe this snuck in PR #520.

See example: https://github.com/electron/fiddle/runs/1427245746?check_suite_focus=true

The root cause of this problem was that the Binary Manager's `getDownloadingVersions` and `getDownloadedVersions` were being used across the `state-spec.ts` file's `updateDownloadedVersionState` but weren't being mocked with each test.

For some reason, just added `mockReturnValue` in the initial mock doesn't work. Instead, I mock `updateDownloadedVersionState`and unmock it for the `updateDownloadedVersionState` tests.

### DIFF
--- a/tests/renderer/state-spec.ts
+++ b/tests/renderer/state-spec.ts
@@ -74,6 +74,7 @@ describe('AppState', () => {
 
   beforeEach(() => {
     appState = new AppState();
+    appState.updateDownloadedVersionState = jest.fn();
     ipcRendererManager.removeAllListeners();
   });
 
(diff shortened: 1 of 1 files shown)


felixrieseberg (approved):
Nice, good find!

PR #543, opened 2020-12-05
fix: Added 'OK' button to failed gist modal

### LINKED ISSUE #529: Failed Gist modal doesn't close .
If the user deletes the published gist and attempts to update it again through fiddle , A warning is shown in a dialog box with no close button .

<strong>Screenshot</strong> : 

Screenshot from 2020-11-19 15-34-16

aryanshridhar:
Added an <code>OK</code> button to <code><a href = 'https://www.electronjs.org/docs/api/dialog#dialogshowmessageboxbrowserwindow-options'>dialog.showMessageBox</a></code> in case of linux platform .

<strong>ScreenShot</strong> : 

Screenshot from 2020-12-05 15-13-52

Fixes #529  .


src/renderer/components/commands-action-button.tsx
```diff
@@ -210,6 +210,7 @@ export class GistActionButton extends React.Component<
         message:
           'Updating Fiddle Gist failed. Are you connected to the Internet and is this your Gist?',
         detail: `GitHub encountered the following error: ${error.message}`,
+        buttons: ['OK'],
```
aryanshridhar, line 213: Should this be adopted in other dialog boxes as well (Say, for example <a href = 'https://github.com/electron/fiddle/blob/84831af2b7095e2e967d2294a3d86e92f32b024f/src/renderer/components/commands-action-button.tsx#L209'>this</a>)?

J-P-S-O (approved):

src/renderer/components/commands-action-button.tsx
codebytere: This should probably be `Ok` not `OK` - additionally, we can perhaps be more specific:
```js
const isLinux = process.platform === 'linux'
...
...(isLinux && { buttons: ['Ok'] })
``` 

but i'm happy to defer to @erickzhao on this one!

aryanshridhar: Oh, is there any reason to be specific? Since adding `Ok` button in case of `Windows`/`MacOS`, just overrides the default value (and is similar to previous case).

codebytere (approved):

Issue #411, opened 2020-06-11
Better support for multiple major versions of Electron


erickzhao:
We currently only store one version of the default Fiddle editor content and the Show Me demos. However, since Fiddle allows you to use older versions of Electron, this code is not guaranteed to work well across major version API changes. As the Electron framework continues to evolve, this will only become more problematic.

We have in the past included stopgaps for API changes (that handle Electron 1 code), but nothing sophisticated enough to support the ever-changing API.

Ref #362 that we haven't merged because the API was only introduced in Electron 3.

ckerr:
Fixed by https://github.com/electron/fiddle/pull/594

ckerr:
...wait, only partially fixed. #594 solves the issue for fiddle templates, but _not_ for the show-me samples.

PR #544, opened 2020-12-08
chore: add `quick-start` folder to `static`

erickzhao:
Ref: #411 

This PR adds a separate folder for each major branch of `electron-quick-start`. The eventual goal is to change the content of a new Fiddle based on the currently-selected major version.

For now, these files are added with no changes to existing code paths. The goal is to add a GitHub Actions cron that will periodically update this content from the `electron-quick-start` source using Octokit.

### DIFF
--- a/static/quick-start/1-x-y/index.html
+++ b/static/quick-start/1-x-y/index.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <!-- https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP -->
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self'">
+    <meta http-equiv="X-Content-Security-Policy" content="default-src 'self'; script-src 'self'">
+    <title>Hello World!</title>
+  </head>
+  <body>
+    <h1>Hello World!</h1>
+    We are using Node.js <span id="node-version"></span>,
+    Chromium <span id="chrome-version"></span>,
+    and Electron <span id="electron-version"></span>.
+
+    <!-- You can also require other files to run in this process -->
+    <script src="./renderer.js"></script>
+  </body>
+</html>
--- a/static/quick-start/1-x-y/main.js
+++ b/static/quick-start/1-x-y/main.js
@@ -1,0 +1,47 @@
+// Modules to control application life and create native browser window
+const {app, BrowserWindow} = require('electron')
+const path = require('path')
+const url = require('url')
+
+function createWindow () {
+  // Create the browser window.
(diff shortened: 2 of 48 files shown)


malept:
I'm curious if you considered using git submodules instead.

erickzhao:
@malept I haven't! Would this involve running a script to `git submodule add` each major version branch into a separate submodule? I haven't worked much with submodules in the past.

Also, wondering what the major benefit of submoduling here would be. Reduced dependency on Octokit?

malept:
> I haven't! Would this involve running a script to `git submodule add` each major version branch into a separate submodule? I haven't worked much with submodules in the past.

Yes, I believe so.

> Also, wondering what the major benefit of submoduling here would be. Reduced dependency on Octokit?

Reduced dependency on Octokit, and we can leverage [Dependabot](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem) to update!

ckerr:
Hey, I'm revisiting this because whatever option we use to populate quick-start templates could also be used to populate test templates for #591.

Strong agree with @malept's dependabot comment, I like the idea of not needing manual upkeep PRs. About submodules though I don't think I see the big picture yet: if someone downloads a Fiddle release instead of cloning the Fiddle repo, how would the "New Fiddle" button work? At that point is there even a submodule to use?

* If we want to make things static, we could probably write a script to automatically populate templates in `static/` from electron-quick-start (the result would look a lot like the diffs in this PR) and then have a PR new updates whenever electron-quick-start gets new commits. Advantages: no new magic needed at runtime. Disadvantage: more scripts & bots to break, and old versions of Fiddle won't know about new content in electron-quick-start.

* One possible approach: at runtime when "New Fiddle" is clicked, do a `git clone` (if needed) + `git branch` of electron-quick-start in a scratch dir, then pull content from that directory. We'd probably also want to keep fallback static contents in case we're offline or if GitHub is down. Advantages: futureproof; doesn't need any dependabot upkeep; falls back gracefully to offline. Disadvantage: git needs to be installed?

* Variant of previous approach: instead of `git clone`, download and open a zip from github (e.g. `https://github.com/electron/electron-quick-start/archive/12-x-y.zip`) and unzip it in a cache directory (e.g. `${config_dir}/Electron Fiddle/static/electron-quick-start`). Advantages: same as previous. Disadvantages: ...none that I see? I kind of like this one.

@malept @erickzhao I think I'm going to try out that last approach and see how it works, but am happy to use other ideas e.g. the submodule suggestion if you think it would be better.

malept:
The zip approach seems fine to me as an alternative to an influx of Dependabot PRs. Which I wouldn't mind as much if we could allowlist, for example, the submodule PRs to automerge.

PR #545, opened 2020-12-09
build: Compile for arm64, upgrade to Electron 11

### LINKED ISSUE #537: Provide native arm64 build for macOS based on Electron 11
It works under Rosetta, but we would appreciate native performance.

felixrieseberg:
This PR allows us to build for Apple's new ARM processors and updates a few of our dependencies in service of that goal. No user-visible changes other than a new build though.

Closes #537


package.json
```diff
@@ -98,7 +98,7 @@
     "less": "^3.12.2",
     "lint-staged": "^10.4.2",
     "log-symbols": "^4.0.0",
-    "node-abi": "^2.19.1",
+    "node-abi": "^2.19.3",
```
malept, line 101: I don't know if we actually need this direct dependency anymore. I don't believe it's actually referenced anywhere, and I've [fixed `electron-rebuild`](https://github.com/electron/electron-rebuild/pull/501) to not call `node-abi` unless there is actually a native module to rebuild.

malept (commented):
At some point, do we want to build for win32/arm64?

Aside from my other question below, LGTM.

erickzhao:
Seems like the tests are failing? 🤔

PR #560, opened 2021-01-07
feat: system-synced dark mode (renderer-only)

erickzhao:
Supersedes #470, but doesn't require any use of the main process!


FAWLKTZ:
Simple simon would say let there be the next one to answer  it .  Win tip dir. defrag. Delete the fiddle add  popup instead

felixrieseberg (approved):

PR #565, opened 2021-01-21
chore: update @electron/get to 1.12.3

erickzhao:


malept:
For context this fixes using Very Old Versions of Electron inside fiddles.

malept (approved):

Issue #571, opened 2021-02-02
Replacing a previously loaded gist with a new one fails.


codebytere:
To reproduce:

1) Open Fiddle
2) Paste https://gist.github.com/codebytere/6ed9ee05659956e33c15343e49e26ed3 and click "Load Fiddle"
3) Replace gist link in command line bar with https://gist.github.com/eada1e5bdd1282db338971c4c42c671e and click "load fiddle" again
4) The screen will flash as it normally does when fiddles are replaced but the first gist will load again and the command line bar and title will once again contain the first gist

I've confirmed this is caused by https://github.com/electron/fiddle/commit/84831af2b7095e2e967d2294a3d86e92f32b024f by reverting it locally and testing the above.

cc @erickzhao

PR #572, opened 2021-02-02
fix: replacing an existing gist link

codebytere:
Closes https://github.com/electron/fiddle/issues/571.

Refs https://github.com/electron/fiddle/pull/531.

Fixes an issue where it was impossible to load a new gist via command bar when one was previously loaded. This was happening because the `onBlur` handler always pulled what it thought was the current gist from `appState`, which meant that if the gist a user was trying to load was not the same as the one in `appState`, it would be overwritten. This PR tries to fix that by asserting that the command bar is either empty or contains the last gist stored into `appState` before reloading the last stored Fiddle Gist.

Tested with the outlined steps in https://github.com/electron/fiddle/issues/571.

### DIFF
--- a/src/renderer/components/commands-address-bar.tsx
+++ b/src/renderer/components/commands-address-bar.tsx
@@ -111,11 +111,13 @@ export class AddressBar extends React.Component<
     this.setState({ value: event.target.value });
   }
 
-  public handleBlur(_event: React.FocusEvent<HTMLInputElement>) {
+  public handleBlur(event: React.FocusEvent<HTMLInputElement>) {
     const { gistId } = this.props.appState;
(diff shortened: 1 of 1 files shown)


malept (commented):
Seems reasonable, although my React/Event knowledge is a bit lacking.

Should we have tests for this?

codebytere:
@malept not sure 🤔 the original PR didn't, and i don't think you can super easily simulate mouse click focus changes

erickzhao (approved):
LGTM.

> i don't think you can super easily simulate mouse click focus changes

I think unit testing the behaviour of this component with Enzyme should be possible, but I'm okay with merging and adding a test later, since this behaviour is fairly annoying.

Issue #568, opened 2021-01-31
Package fiddle not working


jeanravenclaw:
![image](https://user-images.githubusercontent.com/75297066/106385808-21291a80-63e3-11eb-9f73-382cffe837c9.png)
why is this happening?

codebytere:
@jeanenchanted please provide the steps you took prior to encountering this issue. There is no version, no operating system - this issue is insufficient as it stands.

jeanenchanted:
> @jeanenchanted please provide the steps you took prior to encountering this issue. There is no version, no operating system - this issue is insufficient as it stands and we need all of the above to continue or this issue will be closed.

All I did was go to tasks > package fiddle. Version is Electron v11.1.1 on windows 10.  This issue also happens with the default fiddle

codebytere:
@jeanenchanted did you try to rerun when you saw the error? If i leave it it eventually succeeds:

<img width="695" alt="Screen Shot 2021-02-02 at 3 05 06 PM" src="https://user-images.githubusercontent.com/2036040/106674416-0e8c1e00-6568-11eb-95b2-445f09e2585a.png">

codebytere:
Closing as no response provided.

PR #575, opened 2021-02-02
chore: display better error when packaging

codebytere:
Refs https://github.com/electron/fiddle/issues/568.

The error was showing as `[object ErrorEvent]` and we want the user to see a more descriptive message. We should be pushing the message property of the [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) and not the object itself.

cc @felixrieseberg @erickzhao


malept (approved):
Seems reasonable to me.

PR #583, opened 2021-02-28
chore: Update dependencies

felixrieseberg:
This PR updates various dependencies, leaving out those that currently don't work.

### Packages not updated:
1) `mobx` still requires that we're a lot smarter about when and how we subscribe to changes, so I didn't upgrade to v7 just yet.
2) `monaco-editor`'s latest version doesn't load in our setup (see https://github.com/microsoft/monaco-editor/issues/2344). We could fix that by moving to `webpack`, which might eventually be a good idea anyway.
3) `enzyme` doesn't support React 17 just yet (https://github.com/enzymejs/enzyme/issues/2429)

### Other things to note:
* (Fixed) `husky` no longer self-installs, so it's now documented in CONTRIBUTING.md.
* `standard` has picked up on `/static` being in `.eslintignores` and errors with a "no files found" warning, so I've moved the ignores to `eslintrc.js`, which is really only used by eslint.


CONTRIBUTING.md
```diff
@@ -57,6 +57,13 @@ cd fiddle
 yarn
 ```
 
+Optional: Install [`husky`](https://typicode.github.io/husky) and enable
```
erickzhao, line 60: In the [`husky` migration guide](https://dev.to/typicode/what-s-new-in-husky-5-32g5#article-body:~:text=You%20can%20keep%20the%20previous%20behaviour%20(recommended)%20by%20adding%20a%20postinstall%20script%3A), they recommend keeping the old behaviour with a `postinstall` script.

Any reason we switched to the optional install?

felixrieseberg: I don't have strong feelings either way, let me add it back in.

bnb (approved):

erickzhao (approved):
LGTM

codebytere (approved):

MarshallOfSound (approved):

PR #602, opened 2021-03-09
refactor: add preload script

erickzhao:
At its core, this PR attempts to move our `remote.app.getPath()` calls into a preload script.

In order to do so, I had to change a bunch of things along the way:

* Move `window.ElectronFiddle` instantiation from `renderer/state.ts` into the new `preload/preload.ts` file.
* Add more specific conditional guards on `window.ElectronFiddle` checks in the renderer process.
* Add Invoke primitives to `ipcMainManager` and `ipcRendererManager`.
* Test everything.


felixrieseberg (approved):
Elegantly solved and super extendable in the future.

PR #606, opened 2021-03-16
fix: setupBinary() resolves when download completes

ckerr:
Right now there is an edge case where setupBinary() can resolve before a download is done: if called while a download for that version is already running, setupBinary() returns to avoid parallel downloads. Because of this return, the caller may find their RunnableVersion does not have the state that was expected.
    
This PR keeps track of pending downloads so that when a duplicate call occurs, all callers get a Promise that settles when the download does.
    
This is similar to how renderer/content [safeguards against parallel downloads](https://github.com/electron/fiddle/blob/858e4b1834832e0ea5ccf64825c2bd7d201a8a2b/src/renderer/content.ts#L57) of the same zipfile.

### DIFF
--- a/src/renderer/binary.ts
+++ b/src/renderer/binary.ts
@@ -11,11 +11,41 @@ import { AppState } from './state';
 
 const pendingDownloads: Record<string, Promise<void>> = {};
 
+/**
+ * General setup, called with a version. Is called during construction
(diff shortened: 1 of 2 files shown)


erickzhao (approved):
Code LGTM!

PR #608, opened 2021-03-18
fix: nullptr dereference in runner's cleanup function

ckerr:
Was previously erroring out upon passing undefined into `path.join()`.

Related to the appData -> appPaths improvements in https://github.com/electron/fiddle/pull/602

### DIFF
--- a/src/renderer/runner.ts
+++ b/src/renderer/runner.ts
@@ -317,7 +317,7 @@ export class Runner {
     }
 
     const name = await this.appState.getName();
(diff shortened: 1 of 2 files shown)


erickzhao (approved):

erickzhao (commented):
TODO (could be in a separate PR), remove this entirely:

https://github.com/electron/fiddle/blob/858e4b1834832e0ea5ccf64825c2bd7d201a8a2b/src/renderer/state.ts#L209-L211

erickzhao (approved):

PR #609, opened 2021-03-18
fix: accidental switch fall-through in remote-loader

ckerr:
remote-loader currently has an accidental switch fallthrough due to a missing `break;` statement [here](https://github.com/electron/fiddle/blob/255a81d7a7b827d9f290ef72f2924efc9d89ae86/src/renderer/remote-loader.ts#L126).

This PR removes the switch statement by promoting `FILENAME_KEYS` up from [read-fiddle.js](https://github.com/electron/fiddle/blob/255a81d7a7b827d9f290ef72f2924efc9d89ae86/src/utils/read-fiddle.ts#L14) into shared-constants.ts so that it can be used by remote-loader as well. By using this the switch statement can be avoided altogether.

Discovered while fixing up coverage leaks in the autobisect branch :smiley:

### DIFF
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -78,7 +78,7 @@ export class RemoteLoader {
         );
       }
 
-      const loaders: Array<Promise<any>> = [];
+      const loaders: Array<Promise<void>> = [];
 
       for (const child of folder.data) {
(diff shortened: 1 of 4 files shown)


erickzhao (approved):
Great change!

PR #612, opened 2021-03-19
feat: add "New Test" button

ckerr:
This PR adds a 'New Test' menuitem that loads a template from https://github.com/electron/electron-quick-start/tree/test-template .

This is third in a series of PRs (first two are https://github.com/electron/fiddle/pull/610 and https://github.com/electron/fiddle/pull/611) to bring the autobisect branch into `master` in smaller, reviewable chunks instead of one big huge omnibus PR. Next in line is the autobisect feature!


src/main/menu.ts
```diff
@@ -194,6 +194,14 @@ function getFileMenu(
       },
       accelerator: 'CmdOrCtrl+N',
     },
+    {
+      label: 'New Test',
+      click: () => {
+        ipcMainManager.send(IpcEvents.CLEAR_CONSOLE);
+        return ipcMainManager.send(IpcEvents.FS_NEW_TEST);
+      },
+      accelerator: 'CmdOrCtrl+T',
```
clavin, line 203: I'm not strongly against this, but does this need an accelerator?

src/renderer/content.ts
```diff
@@ -7,7 +7,8 @@ import { readFiddle } from '../utils/read-fiddle';
 import * as fsType from 'fs-extra';
 import * as path from 'path';
 import * as semver from 'semver';
-import * as decompress from 'decompress';
+
+const decompress = require('decompress');
```
clavin, line 11: nit: does `import =` work instead?

```suggestion
import decompress = require('decompress');
```

clavin (changes requested):

src/renderer/content.ts
ckerr: This is moot now that ba8a22f40e0a3900078112c4aab154547e2634ba is merged. This will do the Right Thing after rebasing.

src/main/menu.ts
ckerr: I lean toward 'yes' but don't feel strongly about it. Do you want it to be removed?

src/renderer/content.ts
ckerr: This is resolved now

src/main/menu.ts
clavin: Nah it's alright, like I said I'm not strongly against it. Just was curious because it wasn't something I would have done on my own. 👍

clavin (approved):

PR #613, opened 2021-03-20
refactor: remove unnecessary 'async' from getIsDownloaded()

ckerr:
A minor refactor to `src/renderer/binary getIsDownloaded()`, which is async only because it uses fancy-import. However this isn't necessary in order for testing, so the code can be simplified a little by removing `async`.


ckerr:
superceded by https://github.com/electron/fiddle/tree/refactor/use-esModuleInterop

erickzhao (approved):

PR #614, opened 2021-03-21
perf: faster sorting in sortedElectronMap()

ckerr:
Refactor sortedElectronMap() to parse the versions strings less often.

Calls to `semver.valid()`, `semver.prerelease()`, `semver.coerce()`, and `semver.gt()` keep causing the version string to be  reparsed. This refactor calls `semver.parse()` once per RunnableVersion and uses that object for the compare function. For our current startup, this drops the number of times `semver.parse()` is called on startup from 50k times to 4k times.


erickzhao (approved):
LGTM!

PR #615, opened 2021-03-23
feat: add Runner.autobisect()

ckerr:
Add `Runner.autobisect()` + full coverage testing.

Found a couple of edge cases, e.g. we want to make sure that the 'good' version is actually good and 'bad' version is actually bad -- the user might be wrong / the test doesn't work everywhere / etc. So we do now check that the 'good' version and 'bad' version didn't return the same RunResult.

Getting into the home stretch now; GUI + CLI are up next.

### DIFF
--- a/src/renderer/runner.ts
+++ b/src/renderer/runner.ts
@@ -1,11 +1,17 @@
 import { ChildProcess, spawn } from 'child_process';
 import * as path from 'path';
 
-import { EditorValues, FileTransform, RunResult } from '../interfaces';
+import {
+  EditorValues,
+  FileTransform,
+  RunResult,
+  RunnableVersion,
+} from '../interfaces';
 import { IpcEvents } from '../ipc-events';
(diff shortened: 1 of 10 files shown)


src/renderer/runner.ts
```diff
@@ -21,6 +27,12 @@ export enum ForgeCommands {
   MAKE = 'make',
 }
 
+const resultString: Record<RunResult, string> = Object.seal({
```
erickzhao, line 30: nit: Is there a preference for `Object.seal` over `Object.freeze` here?

erickzhao (approved):
LGTM!

src/renderer/runner.ts
ckerr: Good nit. Using `Object.freeze()` here is better.

ckerr: Changed in b72dffc2183adaa8cf9b1e323219a7505f2cb6c9

PR #617, opened 2021-03-24
perf: faster isReleasedMajor()

ckerr:
A small change to some code I wrote last week that keeps showing up with a larger-than-necessary footprint in the coverage tests. To get the max item from a `semver.SemVer[]` we don't need to sort the array; we can walk the array with reduce().

Brings this line from O(n log n) to O(n).


erickzhao (approved):
Logic seems sound to me.

PR #618, opened 2021-03-25
feat: autobisect cli

ckerr:
Implementation and tests for the CLI of autobisect / test.


clavin (approved):

src/renderer/task-runner.ts
```diff
@@ -0,0 +1,127 @@
...
+  private readonly done: (r: RunResult) => void;
+  private readonly hide: (channels: ElectronReleaseChannel[]) => Promise<void>;
+  private readonly log: (message: string) => void;
+  private readonly open: (o: SetFiddleOptions) => Promise<void>;
+  private readonly run: () => Promise<RunResult>;
+  private readonly setVersion: (ver: string) => Promise<void>;
+  private readonly show: (channels: ElectronReleaseChannel[]) => Promise<void>;
+
+  constructor(
+    app: App,
```
erickzhao, line 31: `app` is also exposed via `window.ElectronFiddle.app`, which is usually used for one-off calls elsewhere in the code.

src/renderer/task-runner.ts
```diff
@@ -0,0 +1,127 @@
...
+    private readonly appState: AppState,
+    runner: Runner,
+    ipc: IpcRendererManager,
```
erickzhao, line 34: How come we pass `ipcRendererManager` into the constructor here rather than importing it directly?

tests/renderer/task-runner-spec.tsx
```diff
@@ -0,0 +1,189 @@
...
+import { AppState } from '../../src/renderer/state';
+import { TaskRunner } from '../../src/renderer/task-runner';
+import { IpcRendererManager } from '../../src/renderer/ipc';
+
+describe('Task Runner component', () => {
+  let app: Partial<App>;
+  let appState: Partial<AppState>;
+  let runner: Partial<Runner>;
+  let ipc: IpcRendererManager;
+  // let instance: TaskRunner;
```
erickzhao, line 23: Is this comment intentional here?

src/renderer/task-runner.ts
```diff
@@ -0,0 +1,127 @@
...
+  private readonly log: (message: string) => void;
+  private readonly open: (o: SetFiddleOptions) => Promise<void>;
+  private readonly run: () => Promise<RunResult>;
+  private readonly setVersion: (ver: string) => Promise<void>;
+  private readonly show: (channels: ElectronReleaseChannel[]) => Promise<void>;
+
+  constructor(
+    app: App,
+    private readonly appState: AppState,
+    runner: Runner,
```
erickzhao, line 33: Could we call `app.runner` here rather than using a new parameter?

src/renderer/task-runner.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  private async test(req: TestRequest) {
+    let result = RunResult.INVALID;
+
+    try {
+      await this.setup(req.setup);
+      result = await this.run();
+    } catch (err) {
+      this.log(err);
```
erickzhao, line 96: nit: `const { log } = this;` for consistency as in the `bisect()` and `setup()` methods?

tests/utils/get-version-range-spec.ts
```diff
@@ -0,0 +1,70 @@
...
+import { getVersionRange } from '../../src/utils/get-version-range';
+
+describe('getVersionRange', () => {
+  const knownVersions = [...mockVersionsArray].reverse();
+  const midpoint = Math.floor(knownVersions.length / 2);
+
+  it('returns an empty array if the lower bound cannot be found', () => {
+    const lowerBound = '1.0.0';
+    const upperBound = knownVersions[midpoint].version;
+    // test setup: confirm that one bound is prsent and the other is not
```
erickzhao, line 14:
```suggestion
    // test setup: confirm that one bound is present and the other is not
```

tests/utils/get-version-range-spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const versions = getVersionRange(lowerBound, upperBound, knownVersions);
+
+    expect(versions).toEqual([]);
+  });
+
+  it('returns an empty array if the upper bound cannot be found', () => {
+    const lowerBound = knownVersions[midpoint].version;
+    const upperBound = '10000.0.0';
+    // test setup: confirm that one bound is prsent and the other is not
```
erickzhao, line 26:
```suggestion
    // test setup: confirm that one bound is present and the other is not
```

tests/setup.js
```diff
@@ -7,8 +7,12 @@ configure({ adapter: new Adapter() });
 
 global.confirm = jest.fn();
 
-jest.spyOn(global.console, 'log').mockImplementation(() => jest.fn());
-jest.spyOn(global.console, 'warn').mockImplementation(() => jest.fn());
+if (!process.env.hasOwnProperty('FIDDLE_VERBOSE_TESTS')) {
```
erickzhao, line 10: 👍 on this, but could we document it under `CONTRIBUTING.md`?

src/renderer/state.ts
```diff
@@ -353,6 +356,20 @@ export class AppState {
...
+  }
+
+  @action public showChannels(channels: Array<ElectronReleaseChannel>) {
+    const s = new Set<ElectronReleaseChannel>([
+      ...this.channelsToShow,
+      ...channels,
+    ]);
+    this.channelsToShow = [...s.values()];
+  }
+
```
erickzhao, line 372: Can we use these functions in the `settings-electron.tsx` logic?

src/renderer/task-runner.ts
ckerr: Same reason that constructor DI is used for app et al, to reduce unnecessary dependencies on globals.  You could also ask why all these classes take appState as a constructor argument when it's also a global. *shrug* :slightly_smiling_face: 

Still, it's not a hill that I want to die on. window.ElectronFiddle.app isn't going anywhere. I don't have a commit ready to push for the DI removal but I'll get that to you ASAP; I want to push the other commits to resolve the other conversations.

tests/renderer/task-runner-spec.tsx
ckerr: No. Removed in f58d1467da956ac11c1da63555b6bc9faa8f55e2

src/renderer/task-runner.ts
ckerr: I skipped it there since log() is only called once, so it didn't feel like it was saving anything.  But I don't mind changing it for consistency. de175b33b5903099db65b56d75526369cc6793bf

tests/utils/get-version-range-spec.ts
ckerr: fixed 9649b5c3c40581fce2428825e80b89ae1c4e8487

tests/utils/get-version-range-spec.ts
ckerr: fixed 9649b5c3c40581fce2428825e80b89ae1c4e8487

tests/setup.js
ckerr: Great idea. Added 63ecea476bcb72e323c44c471acfcc0c10edd97e

src/renderer/state.ts
ckerr: Sure! aff8b380fb573e1ac93300276142b4d0b1d1c289

src/renderer/task-runner.ts
ckerr: constructor DI removed in e89536e82ff48932d1b142fa35b897735c7c7190

src/renderer/task-runner.ts
ckerr: e89536e82ff48932d1b142fa35b897735c7c7190

src/renderer/task-runner.ts
ckerr: e89536e82ff48932d1b142fa35b897735c7c7190

src/renderer/task-runner.ts
ckerr: ...I spoke to soon. This `constructor()` function is called from App's `constructor()` function, meaning that the app instance isn't present in `window.ElectronFiddle.app`  yet. :roll_eyes:

ckerr: a607cac6a3079111430ef5412c260dc3f9ee0300 passes `App` in the constructor and then destructures that to get the state and runner objects. So the constructor is still relatively uncluttered but no longer relies on `window.ElectronFiddle.app`.

src/renderer/app.tsx
```diff
@@ -289,7 +289,6 @@ export class App {
 }
 
 window.ElectronFiddle = window.ElectronFiddle || {};
-window.ElectronFiddle.contentChangeListeners =
-  window.ElectronFiddle.contentChangeListeners || [];
-window.ElectronFiddle.app = window.ElectronFiddle.app || new App();
+window.ElectronFiddle.contentChangeListeners ||= [];
```
erickzhao, line 292: Was unaware of these new shorthands, this is really cool!

erickzhao (approved):

PR #619, opened 2021-03-25
feat: autobisect gui

ckerr:
Add an 'auto' button to the bisect dialog to invoke the autobisect feature.


erickzhao (changes requested):
To prevent weird interactions with the UI, I think it'd be good to prevent the version selector and bisect button to be interacted with during the auto-bisect process.

erickzhao (approved):

PR #620, opened 2021-03-25
refactor: remove fancy import

ckerr:
Removes fancyImport. Mostly this results in fs-extra and path being imported at the top level. There was discussion on the Slack channel that we might want to check the performance hit of top-loading vs deferred loading. The thing is, we were top-loading anyway to get the type so that we could cast the fancyImport to the right type. For example you would see

```js
import * as fsType from 'fs-extra';
...
async function doSomethingWithFsExtra() {
  const fs = await fancyImport<typeof fsType>('fs-extra');
  ...
}
```

... so we were already taking the top-level import hit in nearly every case and this PR doesn't change that. The one exception to this is top-importing namor, whose require/import tree is its own 4 files + a few more for its one crypto dependency, about 10 files in total.

PR as discussed with @felixrieseberg and @erickzhao.


src/renderer/fetch-types.ts
```diff
@@ -42,18 +41,10 @@ export async function fetchTypeDefinitions(version: string): Promise<string> {
...
   const typeDefsDir = path.dirname(getOfflineTypeDefinitionPath(_version));
-
-  if (fs.existsSync(typeDefsDir)) {
-    try {
-      await fs.remove(typeDefsDir);
-    } catch (error) {
-      throw error;
-    }
-  }
+  return fs.remove(typeDefsDir);
```
ckerr, line 47: Small explanation of this refactor: fs-extra's remove() is a no-op if the path doesn't exist, so fs.existsSync() is not needed. And since we rethrow any error we catch, the try-catch block is not useful.

erickzhao (approved):

PR #621, opened 2021-03-26
refactor: avoid hardcoding filenames, editor ids

ckerr:
This is a minor refactor to avoid hardcoding files and editor ids. Most cases can iterate through the `FILENAME_KEYS` object which lets the code avoid hardcoding files / editor is _and_ avoid making assumptions on what editors exist.

I started this cleanup while we were prototyping a new 'test.js' editor. It's less relevant since we discarded that idea, but it's still preferable to use symbolic names instead of magic strings.

No functional changes.


erickzhao (approved):
Love this change :D

PR #622, opened 2021-03-26
feat: add an icon for Electron DLs being unzipped

ckerr:
Right now there's a visual wart that during a download the icon goes from 'cloud' (unknown) -> 'cloud-download' (downloading) -> 'cloud' (unzipping) -> 'saved' (saved) and that second shortlived 'cloud' during unzip is a little confusing because for a moment it looks like the download has disappeared. 

This PR uses 'compressed' as the icon during unzip to avoid giving the user that mixed signal.


erickzhao (approved):
Good change, and also happens to be clearer than the nested ternary! :shipit:

ckerr:
Since it's a switch on the VersionState, the compiler should also warn us in the future if we add a new state and forget to handle it here :tada:

PR #628, opened 2021-03-29
fix: remove unexported member in task-runner-spec

erickzhao:
It seems like #623 makes the tests fail because `IpcRendererManager` is no longer exported. This PR fixes that.


ckerr (approved):

clavin (changes requested):
Why this instead of just exporting `IpcRendererManager` again?

Issue #574, opened 2021-02-02
Feature request: create additional files


nornagon:
It would be useful to be able to create files beyond the built-in 5 (main.js, index.html, renderer.js, preload.js and styles.css). e.g. for fiddles which use `<webview>`, it would be convenient to be able to have the webview src point at a 2nd html file. Or for preloads, you might want to have different preloads for two different windows.

jkleinsc:
I'm adding my +1 to this.  I am trying to test service workers in Electron and would love to be able to use fiddle to do so.

PR #636, opened 2021-04-01
feat: ability to add custom editors

codebytere:
Closes https://github.com/electron/fiddle/issues/574.

This implements a workflow for adding and removing custom editors, which may have either a `html`, `js` or `css` extension.

cc @nornagon @ckerr @erickzhao @felixrieseberg @jkleinsc 

Some screenshots:

Screen Shot 2021-04-01 at 8 46 30 PM

Screen Shot 2021-04-01 at 8 46 34 PM

Screen Shot 2021-04-01 at 8 46 42 PM

Screen Shot 2021-04-01 at 8 46 42 PM


erickzhao:
@codebytere Thanks for this!

I'm not sure if this should go into this PR, but I think a better user experience would be to have a sidebar because that dropdown is getting a bit crowded. BlueprintJS has a [File Tree](https://blueprintjs.com/docs/#core/components/tree) component that we could leverage to this end.

codebytere:
@erickzhao i don't disagree it's a bit crowded but personally i think the cost-benefit of this is perhaps simpler to add to the existing Editor palette 🤔  As a user I think i'd have a harder time seeing a whole new tree view with the explicit purpose of adding a new editor, at least without doing a lot more UI refactoring it may be better to push into a subsequent PR (or several)

codebytere:
I still need to add some tests but i'd like to some preliminary review before i do!

erickzhao:
> at least without doing a lot more UI refactoring it may be better to push into a subsequent PR (or several)
For sure, I agree! I'll add a review on the existing implementation today. :)

src/renderer/app.tsx
```diff
@@ -137,21 +162,29 @@ export class App {
...
-    const values: EditorValues = {
-      css: getEditorValue(EditorId.css),
-      html: getEditorValue(EditorId.html),
-      main: getEditorValue(EditorId.main),
-      preload: getEditorValue(EditorId.preload),
-      renderer: getEditorValue(EditorId.renderer),
+    const customEditorValues = Object.assign(
+      {},
+      ...customMosaics.map((m) => ({ [m]: getEditorValue(m) })),
+    );
```
ckerr, line 174:
```suggestion
    const customEditorValues = Object.fromEntries(
      customMosaics.map((m} => [m, getEditorValue(m)])
    );
```

src/renderer/content.ts
```diff
@@ -127,7 +127,13 @@ export async function getContent(
   name: EditorId,
   version: string,
 ): Promise<string> {
-  return (await getTemplate(version))[name];
+  const template = await getTemplate(version);
+  if (template?.[name]) {
+    return template[name];
+  } else {
```
ckerr, line 133:
```suggestion
  const content = await getContent(name, version);
  if (content) {
    return content;
  } else {
```

src/utils/read-fiddle.ts
```diff
@@ -39,9 +38,8 @@ export async function readFiddle(folder: string): Promise<EditorValues> {
     console.warn(`readFiddle(): "${folder}" does not exist`);
   } else {
     for (const file of fs.readdirSync(folder)) {
-      const key = FILENAME_KEYS[file];
-      if (key) {
-        ret[key] = tryRead(file);
+      if (Object.values(DefaultEditorId).includes(file as DefaultEditorId)) {
+        ret[file] = tryRead(file);
```
ckerr, line 42: This doesn't seem to support loading custom editors from saved fiddles?

ckerr:
> I'm not sure if this should go into this PR, but I think a better user experience would be to have a sidebar because that dropdown is getting a bit crowded. BlueprintJS has a [File Tree](https://blueprintjs.com/docs/#core/components/tree) component that we could leverage to this end.

TBH I also think > 4 editors gets pretty crowded and would prefer a File Tree for that use case.

That said, this PR is an improvement over the current state of not supporting custom windows :smile: and I'd be fine with adding this and seeing how well it wears.

src/utils/read-fiddle.ts
codebytere: ooh you're right, looks like i spaced that! i'll do the same permissions dance for non-default editors

src/interfaces.ts
```diff
@@ -114,31 +114,42 @@ export const enum GenericDialogType {
...
 
-// Editors
-export const enum EditorId {
-  'main' = 'main',
-  'renderer' = 'renderer',
-  'html' = 'html',
-  'preload' = 'preload',
-  'css' = 'css',
+// Default Editors
+export enum DefaultEditorId {
```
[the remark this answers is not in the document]
codebytere: i did! and for that precise linked reason :)

src/renderer/state.ts
```diff
@@ -789,6 +794,17 @@ export class AppState {
     this.setVisibleMosaics(currentlyVisible.filter((v) => v !== id));
   }
 
+  /**
+   * Hides the panel for a given MosaicId.
+   *
+   * @param {MosaicId} id
+   */
+  @action public removeMosaic(id: MosaicId) {
```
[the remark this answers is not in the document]
codebytere: @ckerr i'd say this is inaccurate as this method applies only to custom editors and doesn't hide them, it fully removes them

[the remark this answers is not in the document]
ckerr: No, terser is better. I agree, let's keep it as-is

erickzhao (commented):
Testing this locally, running the Fiddle seems to encounter issues with the good old `decomment` module. 🙃 
image

codebytere:
@erickzhao fixed in latest commit!

codebytere:
So far i've tested: 

- Creating a custom editor with a `css`, `js` or, `html` extension
- Removing a custom editor with a `css`, `js` or, `html` extension
- Saving a Fiddle with a custom editor
- Opening a saved Fiddle with a custom editor
- Publishing a Fiddle to gist with a custom editor
- Loading a Fiddle Gist with a custom editor

src/main/about-panel.ts
```diff
@@ -14,7 +14,9 @@ export function setupAboutPanel(): void {
   app.setAboutPanelOptions({
     applicationName: 'Electron Fiddle',
     applicationVersion: app.getVersion(),
-    authors: fs.readJSONSync(contribFile).map(({ name }) => name),
+    authors: fs
+      .readJSONSync(contribFile)
+      .map(({ name }: { name: string }) => name),
```
ckerr, line 19: :+1: I just hit this warning when doing test:ci

src/renderer/components/commands-editors.tsx
```diff
@@ -103,6 +150,72 @@ export class EditorDropdown extends React.Component<
...
+          });
+
+          appState.toggleGenericDialog();
+        } else {
+          appState.customMosaics.push(name);
+          appState.showMosaic(name);
+        }
+      }
+    }
+  }
```
ckerr, line 209: this nested if... if... else... if... else... logic is kind of difficult to read and is nested pretty deep. Could it be simplified with returns, e.g. `if (cancelled) return;` and similar for the other conditionals

src/renderer/components/editor.tsx
```diff
@@ -32,7 +32,8 @@ export class Editor extends React.Component<EditorProps> {
   constructor(props: EditorProps) {
     super(props);
 
-    switch (props.id) {
+    const fileExtension = props.id.split('.')[1];
```
ckerr, line 35: I think this fails on custom entries that have a dot in the filename?

Another part of the code uses

`const getSuffix = (name: string) => path.parse(name).ext.slice(1);`

Also depending on how you feel about removing the `switch` we could also use `if props.id.endsWith('.html')...else`

src/renderer/content.ts
```diff
@@ -127,7 +127,16 @@ export async function getContent(
   name: EditorId,
   version: string,
 ): Promise<string> {
-  return (await getTemplate(version))[name];
+  const { defaultMosaics, customMosaics } = await getTemplate(version);
+
+  if (defaultMosaics?.[name]) {
+    return defaultMosaics[name];
+  } else if (customMosaics?.[name]) {
```
ckerr, line 134: _nit_ The word `else`can be removed here and two lines down

src/renderer/npm.ts
```diff
@@ -75,11 +75,22 @@ export async function getIsPackageManagerInstalled(
...
+  // Filter out all editor values which aren't JavaScript files.
+  const files = [
+    ...Object.values({
+      ...Object.keys(values)
+        .filter((k) => k.endsWith('.js'))
+        .reduce((obj, v) => {
+          return { ...obj, [v]: values[v] };
+        }, {}),
+    }),
+  ] as Array<string>;
```
ckerr, line 89:
```suggestion
  const files = Object.entries(values)
    .filter(([filename]) => filename.endsWith('.js'));
    .map(([_, content]) => content);

```

ckerr (commented):
Do we need a conceptual difference between default ids and custom ids, or could Fiddle just say "hey here are editors for the js/html/css files we found"? It feels like the code for the latter would be much simpler.

I like this PR but the default + custom logic does make the editors/mosaics/content code more complex and I'm wondering if Fiddle would run equally well if we tore it all out and just used the new `isValidEditorName()` name everywhere for deciding what we should and shouldn't load

src/renderer/components/commands-editors.tsx
ckerr: As a followup: coverage of src/renderer/components/commands-editors.tsx drops 41% in this PR as of b776cb04532cdbece4a3889dceb347bf581a6c81

src/renderer/state.ts
```diff
@@ -759,6 +768,17 @@ export class AppState {
...
+  /**
+   * Removes the panel for a given custom MosaicId.
+   *
+   * @param {MosaicId} id
+   */
+  @action public removeCustomMosaic(id: MosaicId) {
+    this.hideAndBackupMosaic(id);
+    delete window.ElectronFiddle.editors[id];
+    this.customMosaics = this.customMosaics.filter((mosaic) => mosaic !== id);
+  }
```
ckerr, line 780: This block needs tests

src/renderer/file-manager.ts
```diff
@@ -62,11 +66,21 @@ export class FileManager {
...
     console.log(`FileManager: Asked to open`, filePath);
-    const editorValues = await readFiddle(filePath);
-    window.ElectronFiddle.app.replaceFiddle(editorValues, { filePath });
+    const { customMosaics, defaultMosaics } = await readFiddle(filePath);
+
+    const editorValues = defaultMosaics;
+    for (const custom of Object.keys(customMosaics)) {
+      const verified = await app.remoteLoader.verifyCreateCustomEditor(custom);
+      if (verified) {
+        editorValues[custom] = customMosaics[custom];
```
ckerr, line 79: Needs tests

src/renderer/file-manager.ts
```diff
@@ -123,14 +137,26 @@ export class FileManager {
...
-      output.set(filename, values[editorId]);
+
+    // Get values for default editors.
+    for (const filename of Object.values(DefaultEditorId)) {
+      output.set(filename, values[filename]);
+    }
+
+    // Get values for any custom editors that have been created.
+    for (const mosaic of customMosaics) {
+      output.set(mosaic, values[mosaic]);
```
ckerr, line 155: this line not reached in tests

src/utils/read-fiddle.ts
```diff
@@ -39,16 +46,18 @@ export async function readFiddle(folder: string): Promise<EditorValues> {
     console.warn(`readFiddle(): "${folder}" does not exist`);
   } else {
     for (const file of fs.readdirSync(folder)) {
-      const key = FILENAME_KEYS[file];
-      if (key) {
-        ret[key] = tryRead(file);
+      if (Object.values(DefaultEditorId).includes(file as DefaultEditorId)) {
+        defaultMosaics[file] = tryRead(file);
+      } else if (isValidEditorName(file)) {
+        customMosaics[file] = tryRead(file);
```
ckerr, line 52: Not reached in tests

src/renderer/remote-loader.ts
```diff
@@ -134,9 +138,26 @@ export class RemoteLoader {
...
+        (file) =>
+          !Object.values(DefaultEditorId).includes(file as DefaultEditorId),
+      );
+
+      for (const mosaic of maybeCustomEditors) {
+        const verified = await this.verifyCreateCustomEditor(mosaic);
+        if (verified) {
+          values[mosaic] = this.getContentOrEmpty(gist, mosaic);
+          this.appState.customMosaics.push(mosaic as CustomEditorId);
+        }
```
ckerr, line 158: Not reached in tests

src/renderer/remote-loader.ts
```diff
@@ -201,6 +222,19 @@ export class RemoteLoader {
...
+      type: GenericDialogType.confirm,
+      label: `Do you want to create a custom editor with name: ${editorName}?`,
+      cancel: 'Skip',
+    });
+
+    this.appState.isGenericDialogShowing = true;
+    await when(() => !this.appState.isGenericDialogShowing);
+
+    return !!this.appState.genericDialogLastResult;
+  }
```
ckerr, line 236: Not reached in tests

src/renderer/app.tsx
```diff
@@ -67,6 +70,16 @@ export class App {
...
     }
 
+    // Remove all previously created custom editors.
+    this.state.customMosaics = [];
+    const customEditors = Object.keys(editorValues).filter(
+      (v) => !Object.values(DefaultEditorId).includes(v as DefaultEditorId),
+    ) as CustomEditorId[];
+
+    for (const mosaic of customEditors) {
+      this.state.customMosaics.push(mosaic);
```
ckerr, line 80: Not reached in tests

src/utils/read-fiddle.ts
```diff
@@ -8,29 +11,33 @@ import * as path from 'path';
...
+    [DefaultEditorId.preload]: '',
+    [DefaultEditorId.renderer]: '',
   };
 
   const hits: string[] = [];
-  const misses = new Set(Object.keys(FILENAME_KEYS));
+  const misses = new Set(Object.values(DefaultEditorId));
+
+  const isValidEditorName = (name: string) =>
+    /^.*\.(js|html|css)$/gm.test(name);
```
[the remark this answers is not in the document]
ckerr: actually `^.*` not needed either; `/\.(css|html|js)$/` should work.

Might want to make case-insensitive because Windows?

`/\.(css|html|js)$/i`

src/interfaces.ts
```diff
@@ -108,37 +108,53 @@ export interface Templates {
   [index: string]: string | Templates;
 }
 
+export interface FiddleEditorTypes {
```
ckerr, line 111: _nit_ Maybe this could be called "EditorTypes"? The "Fiddle" prefix doesn't seem to add value

codebytere:
@ckerr primarily i think it's important to draw a distinction between trusted and untrusted editors - i do think it'd be simpler from a complexity standpoint to not consider that but personally speaking think it's important to explicitly draw a distinction for the user such that they are always explicitly opting into creation/loading of non-default editors 🤔

ckerr:
> @ckerr primarily i think it's important to draw a distinction between trusted and untrusted editors - i do think it'd be simpler from a complexity standpoint to not consider that but personally speaking think it's important to explicitly draw a distinction for the user such that they are always explicitly opting into creation/loading of non-default editors

What would the difference be to the user in practice, e.g. when loading a gist, or opening a show-me or a test or the default fiddle?

codebytere:
@ckerr yes - about loading from either disk or a gist!

ckerr:
If that is the only advantage, couldn't we just have something in interfaces like

```
export const KNOWN_FILES = new Set(
  'main.js',
  'index.html',
  ...
)
```

And then when loading files, it could check against `KNOWN_FILES` when deciding whether or not to prompt?

I'm :100: on supporting more files than the current hardcoded ones, but am hopeful that we can achieve the same goals with a much simpler codebase.

This PR helps show the limitations of the hardcoded list, but the complexity of that design existed before this PR.

src/renderer/state.ts
```diff
@@ -716,7 +723,9 @@ export class AppState {
 
     for (const id of ALL_MOSAICS) {
```
ckerr, line 724: This is the only place in `src/` where `ALL_ MOSAICS` is used and it could be replaced by

```
for (const id of currentlyVisible) {
  if (!visible.includes(id)) {
    ...
  }
}

for (const id of visible)) {
  if (!curentlyVisible.includes(id)) {
    ...
  }
}
```

IDK how this would play in the tests, but this looks like low-hanging fruit for removing one piece of cruft

ckerr (approved):

PR #637, opened 2021-04-01
feat: add About to Help menu

### LINKED ISSUE #631: Add menu item for About panel
There is code to [set up the about panel](https://github.com/electron/fiddle/blob/v0.22.0/src/main/about-panel.ts), but there doesn't seem to be any code to show it (i.e., `app.showAboutPanel()`). We should add a menu item to the bottom of the Help Menu, "About Electron Fiddle", that shows the about panel.

ckerr:
As @malept reported in #631, we have code to set up the About panel but never use it. This PR adds a Help menuitem to do so.

The previous code  had separate handling for different platforms -- setupAboutPanel() was a no-op on Windows, and then filled in vairous AboutPanelOptions based on platform. Since this setup is cheap, this PR just sets everything unconditionally and leaves it to Electron to decide what to use.

Doing it this way seems more straightforward, but I'm wondering what that `if (process.platform === 'win32') return` was for. It was introduced in https://github.com/electron/fiddle/commit/14279283d648ea73a1853fc972c453a9b40ee725.

Fixes #631.

### DIFF
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -76,6 +76,15 @@ function getHelpItems(): Array<MenuItemConstructorOptions> {
         shell.openExternal('https://github.com/electron/electron/issues');
       },
     },
+    {
+      type: 'separator',
+    },
+    {
+      label: 'About Electron Fiddle',
+      click() {
+        app.showAboutPanel();
+      },
+    },
   ];
 }
 
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -31,7 +31,9 @@ function isSubmenu(
  * @returns {Array<Electron.MenuItemConstructorOptions>}
  */
 function getHelpItems(): Array<MenuItemConstructorOptions> {
-  const items = [
+  const items: MenuItemConstructorOptions[] = [];
+
+  items.push(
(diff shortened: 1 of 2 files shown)


src/main/menu.ts
```diff
@@ -76,6 +76,15 @@ function getHelpItems(): Array<MenuItemConstructorOptions> {
         shell.openExternal('https://github.com/electron/electron/issues');
       },
     },
+    {
+      type: 'separator',
+    },
+    {
+      label: 'About Electron Fiddle',
```
erickzhao, line 83: Can we remove this menuItem for macOS? There's already the `About Electron Fiddle` menu item under the first submenu set by the `electron-default-menu` package: 
image

This positioning aligns with the behaviour of native macOS apps:
image

erickzhao (changes requested):

src/main/menu.ts
ckerr: 50231a12263a025b503aab357effbbc405fa53d0 omits on macOS now

erickzhao (approved):
LGTM once tests pass :)

Issue #392, opened 2020-05-12
"Make installers for Fiddle" task not working


0xtrzy:
clear instalation
download, run ok
package ok
task->make installers for fidler

------
0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node', '/usr/local/bin/npm', 'run', 'make' ]
2 info using npm@6.14.5
3 info using node@v12.16.1
4 verbose run-script [ 'premake', 'make', 'postmake' ]
5 info lifecycle educated-paste-juggle-tl6gg@1.0.0~premake: educated-paste-juggle-tl6gg@1.0.0
6 info lifecycle educated-paste-juggle-tl6gg@1.0.0~make: educated-paste-juggle-tl6gg@1.0.0
7 verbose lifecycle educated-paste-juggle-tl6gg@1.0.0~make: unsafe-perm in lifecycle true
8 verbose lifecycle educated-paste-juggle-tl6gg@1.0.0~make: PATH: /usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin:/tmp/tmp-77498s2aclx4nb184/node_modules/.bin:/home/user/.cargo/bin:/home/user/.local/bin:/home/user/bin:/home/user/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin
9 verbose lifecycle educated-paste-juggle-tl6gg@1.0.0~make: CWD: /tmp/tmp-77498s2aclx4nb184
10 silly lifecycle educated-paste-juggle-tl6gg@1.0.0~make: Args: [ '-c', 'electron-forge make' ]
11 silly lifecycle educated-paste-juggle-tl6gg@1.0.0~make: Returned: code: 1  signal: null
12 info lifecycle educated-paste-juggle-tl6gg@1.0.0~make: Failed to exec make script
13 verbose stack Error: educated-paste-juggle-tl6gg@1.0.0 make: `electron-forge make`
13 verbose stack Exit status 1
13 verbose stack     at EventEmitter.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:332:16)
13 verbose stack     at EventEmitter.emit (events.js:311:20)
13 verbose stack     at ChildProcess.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:55:14)
13 verbose stack     at ChildProcess.emit (events.js:311:20)
[...]

PR #643, opened 2021-04-08
fix: installers for nightlies

codebytere:
Closes https://github.com/electron/fiddle/issues/392.

Fixes packaging nightly builds. Also fixes fetching types for nightly builds.

Nightlies come from electron-nightly and not electron, so we need to change the name in package.json. We also need to force a new ABI number as none has been assigned for nightlies.

Tested on macOS.


ckerr (approved):

erickzhao (approved):

VChernomYA (commented):
safari

PR #654, opened 2021-04-12
refactor: only build `appState.versions` once

ckerr:
Right now, whenever we create/update/delete a version, all the proxies in the `appState.versions` array are `toJS`'ed back into normal JS objects, thrown into a temporary container where the update happens, and then re-converted back into proxies again in a rebuilt `appState.versions` object.

This was necessary when the code was written because mobx 4 did not have support for reacting to properties being added/removed from observed objects. However this feature was [added in mobx 5](https://github.com/mobxjs/mobx/blob/e17c47833d1812eee6d77914be890aa41e4b7908/CHANGELOG.md#500) so this should "just work" without our code needing to rebuild the object.

Building the `versions` object once and then updating it instead of rebuilding it has some nice side-effects in other parts of the code. For example `binary.ts` used to have to look up the RunnableVersion every time it was used:

```js
appState.versions[version].state = VersionState.downloading;
await download(...);
appState.versions[version].state = VersionState.unzipping;
```

Because a lookup every time was the only way to safely know that the _current_ proxy, instead of one discarded by a rebuild, was being modified. But once that churn has been removed, the code can now read:

```js
ver.state = VersionState.downloading;
await download(...);
ver.state = VersionState.unzipping;
```

... in fact, `binary.ts` no longer needs to know about AppState at all :tada:

### DIFF
--- a/src/renderer/binary.ts
+++ b/src/renderer/binary.ts
@@ -2,11 +2,15 @@ import * as fs from 'fs-extra';
 import * as path from 'path';
 import extract from 'extract-zip';
 
-import { Version, VersionSource, VersionState } from '../interfaces';
+import {
+  RunnableVersion,
+  Version,
+  VersionSource,
+  VersionState,
+} from '../interfaces';
 import { normalizeVersion } from '../utils/normalize-version';
 import { USER_DATA_PATH } from './constants';
 import { removeTypeDefsForVersion } from './fetch-types';
-import { AppState } from './state';
 import { download as electronDownload } from '@electron/get';
 
 // versions that are currently being downloaded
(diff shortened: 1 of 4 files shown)


src/renderer/binary.ts
```diff
@@ -204,28 +193,20 @@ export function getElectronBinaryPath(
...
+      console.debug(`Binary: Version ${version} download progress: ${percent}`);
     }
   };
 
-  if (!appState.versions[version]) {
-    throw new Error(
-      `Version ${version} does not exist in state, cannot download`,
-    );
-  }
-
```
ckerr, line 228: Note: test no longer needed

tests/renderer/binary-spec.ts
```diff
@@ -262,66 +233,22 @@ describe('binary', () => {
...
-        state: VersionState.unknown,
-        version,
-      };
-
-      await setupBinary(mockState, version);
-
-      expect(electronDownload).not.toHaveBeenCalled();
-      expect(mockState.versions[version].state).toBe('ready');
-    });
-
```
ckerr, line 280: Removed because this is effectively the same as `returns the same promise if called twice for the same version`

codebytere (approved):
cooooooool - nice change!

Issue #644, opened 2021-04-08
Have an option to keep versions of Electron in zipped rather than unzipped form


ckerr:
Inspired by https://github.com/electron/fiddle/issues/638 and shares the same goal of saving disk space, but they're two different ideas that could both be implemented.

A zipped Electron release is about 40% the size of an unzipped one, and we have a _lot_ of Electron versions. It would be nice if Fiddle had an option to keep the downloads in zipped form and then unzipped only as needed, e.g. when selecting a new version in the dropdown. Unzipping is fast.

IMO this would be best as a setting that users can opt out of. I know that @codebytere and I [both use](https://xkcd.com/1172/) Fiddle as a de facto Electron version manager even when running code from the command line

dragonDScript:
I suggest storing unzipped versions on /tmp as they would be removed on every boot.

aryanshridhar:
Closing it! (handled in #1134)

PR #656, opened 2021-04-13
feat: new version filters in File > Preferences > Electron

ckerr:
This PR revamps the version filters in File > Preferences > Electron:

1. The `statesToShow` toggles exist because those toggles fit the implementation, e.g. there's not really a use case for saying, e.g. "I only want to see versions that aren't downloaded." ... and the "Ready" toggle is unclickable with a tooltip that reads "Always enabled." This PR replaces that row with one toggle: "Show non-downloaded versions?"

2. The 'unsupported' filter has been removed. Back in 2019(?) the project made an Electron release tagged 'unsupported' because of reasons. It was a one-off thing that hasn't been done in years and doesn't need to be in the UI.

3. A new 'obsolete' filter has been added. This filter toggles whether or not end-of-lifed versions of Electron should be shown. This is a useful setting because disabling EOL versions by default (a) cuts down `versionsToShow` down from 580 versions to 111, and (b) signals to new users that it's preferable to write code on supported versions. Maintainers or users who need access to EOL versions can toggle the setting to get the same long list we know and love :heart: 

Removing `statesToShow` will make it possible to [add another state](https://github.com/electron/fiddle/issues/644) without complicating this UI :smile: 

## Before

old-filters

## After

new-filters

## Discussion

This PR currently disables EOL versions for the reasons listed above but it is a change that could affect users. I'm open to feedback if anyone has reasons that end-of-lifed versions should be shown by default.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -58,7 +58,6 @@ export const enum ElectronReleaseChannel {
   stable = 'Stable',
   beta = 'Beta',
   nightly = 'Nightly',
-  unsupported = 'Unsupported',
 }
 
 export interface SetFiddleOptions {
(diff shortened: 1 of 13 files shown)


codebytere (approved):

erickzhao:
This is great!

PR #659, opened 2021-04-15
refactor: partial simplification of Mosaic/Editor code

ckerr:
1. Remove the unused "Docs & Demo" code. There's scaffolding in place to use this feature, but it's been idle for a couple of years now and it makes the editor / mosaic code more complex by adding extra types and tests. This PR removes it. This means `PanelId` can be removed. Since `MosaicId` and `EditorId` are now equal,  the former has been removed for clarity.

2. Add `src/utils/editor-utils.ts` as a clearinghouse for some simple editor utils:
* `isKnownFile()` returns true if the file is one of the default files.
* `isSupportedFile()` returns true if the file is it's OK as a custom file.
* `getEditorTitle()` does what it says on the tin. Extracted from renderer/components/editors for easier reuse between components.
* `getEmptyContent()` does what `EMPTY_CONTENT` used to do, but is slightly easier easier to use.
* `compareEditors()` for sorting editors by ordered defaults first, then custom editors alphabetically.

3. Refactor a few cases where it was relatively easy to remove references to the mosaic/editor types in interfaces.ts without being too intrusive, e.g. `fetchGistAndLoad()`. There are a lot of these little changes, but so far they are just reducing the dependencies on the existing interfaces. Nothing fundamental has changed yet in terms of how the app handles default editors & custom editors.

There are one or two changes that might look counter-intuitive compared to the previous code. I've commented on them inline.


src/interfaces.ts
```diff
@@ -122,41 +122,22 @@ export enum DefaultEditorId {
   'css' = 'styles.css',
 }
 
-export type CustomEditorId = Exclude<
-  `${string}.${'js' | 'html' | 'css'}`,
-  DefaultEditorId
->;
+export type CustomEditorId = `${string}.${'js' | 'html' | 'css'}`;
```
ckerr, line 125: Note: I removed the `Exclude<>` because it wasn't working correctly. TS is smart enough to note files that end in the wrong suffix, but assigning `main.js` etc to a CustomEditorId compiled without warning. I was surprised too; it _looks_ like it should've worked.

src/renderer/remote-loader.ts
```diff
@@ -136,23 +115,9 @@ export class RemoteLoader {
...
-
-      // If it's a custom editor explicitly request permission before creation.
-      for (const mosaic of maybeCustomEditors) {
-        const verified = await this.verifyCreateCustomEditor(mosaic);
-        if (verified) {
-          values[mosaic] = this.getContentOrEmpty(gist, mosaic);
-          this.appState.customMosaics.push(mosaic as CustomEditorId);
+      for (const [id, data] of Object.entries(gist.data.files)) {
+        if (isKnownFile(id) || (await this.verifyCreateCustomEditor(id))) {
+          values[id] = data.content;
```
ckerr, line 120: Note: App.replaceFiddle(), which is called right after this via handleLoadingSuccess(), rebuilds the this.app.customMosaics array, so we no longer populate it here.

src/interfaces.ts
```diff
@@ -122,41 +122,22 @@ export enum DefaultEditorId {
...
+export type EditorId = CustomEditorId | DefaultEditorId;
 
 export const DEFAULT_EDITORS = [
   DefaultEditorId.main,
   DefaultEditorId.renderer,
   DefaultEditorId.preload,
   DefaultEditorId.html,
   DefaultEditorId.css,
-];
+] as const;
```
codebytere, line 135: can we consolidate this with `SORTED_EDITORS`?

src/renderer/remote-loader.ts
codebytere: The previous implementation iterates through all default editors, whereas this only iterates through gist files - I see above you say it gets rebuilt, so do you mean that the empty editors are appropriately populated there?

src/renderer/state.ts
```diff
@@ -725,25 +717,25 @@ export class AppState {
     // mount to ensure that we can load content into the editors as soon as they're
     // declared visible.
 
-    await waitForEditorsToMount(visible);
+    await waitForEditorsToMount(visible.filter(isKnownFile));
```
codebytere, line 720: why are we differentiating between custom/default editors here?

src/utils/editor-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function getEditorTitle(id: EditorId): string {
+  return TITLE_MAP[id] || `Custom Editor (${id})`;
+}
+
+const EMPTY_EDITOR_CONTENT = {
+  css: '/* Empty */',
+  html: '<!-- Empty -->',
+  js: '// Empty',
+} as const;
```
codebytere, line 24: curiosity question: why are we adding `as const` here and elsewhere?

ckerr: Mostly just cleanliness. Everything that can be const'ed should be, so that the compiler finds the humans' mistakes :robot:

src/interfaces.ts
ckerr: Oh I like that idea! Lemme try it out...

src/renderer/state.ts
ckerr: good catch. It shouldn't be. Fixed fce9f174e1bdce94ca8fa4ba569bd6d7a730ee2c

src/interfaces.ts
ckerr: added in 33586bc9ca6e23dc73bcfb38c9b46c6a37f36688

src/renderer/remote-loader.ts
ckerr: Missing defaults will be hidden when the fiddle is loaded. If the user then goes and toggles on a default via the toolbar, it will be loaded with the electron-quick-start default via https://github.com/electron/fiddle/blob/d2b7b61afd81f3a1c6cf6d3260e5745be3e1e831/src/renderer/components/editor.tsx#L164

codebytere: ah, ok! cool :D

ckerr:
Superceded by https://github.com/electron/fiddle/pull/664

Issue #660, opened 2021-04-15
Invalid npm registry modules cause endless install spin loop


codebytere:
To repro ad `const x = require('skdhufiwegbwkebcklwqebcklqeblqkwe')` to `main.js` in stock fiddle and see that no error is printed to console and the `Installing Modules` spinner never stops.

PR #661, opened 2021-04-15
fix: endless spin loop on module install failure

codebytere:
Closes https://github.com/electron/fiddle/issues/660.

After:

Screen Shot 2021-04-15 at 12 39 10 PM


erickzhao (approved):
LGTM, thanks!

PR #663, opened 2021-04-20
fix(linux): use shell path, same as macOS

malept:
## Problem

On Linux, when you run Fiddle from an app launcher (e.g., via the Activities Launcher in Gnome), it uses the `PATH` environment variable as defined when X11/Wayland is started, that is, usually without all of the PATH changes from shellrc (i.e., your Bash/ZSH/etc user config). On macOS, this is fixed with the `fix-path` Node module. Linux should be fixed similarly, but `fix-path` is [weirdly limited to macOS](https://github.com/sindresorhus/fix-path/blob/c7578359558831402413d5d8323057406bfdf40d/index.js#L5-L7).

## Solution

Drop `fix-path`, replace with `shell-path` (which is a direct dependency of `fix-path`), and self-implement a subset of `fix-path`.


src/utils/exec.ts
```diff
@@ -32,16 +32,19 @@ export async function exec(dir: string, cliArgs: string): Promise<string> {
...
  */
 export async function maybeFixPath(): Promise<void> {
-  if (!_fixPathCalled && process.platform !== 'win32') {
-    const fixPaths = require('fix-path');
-    fixPaths();
+  if (!_shellPathCalled && process.platform !== 'win32') {
+    const shellPaths = require('shell-path');
+    const paths = await shellPaths();
+    if (paths) {
+      process.env.PATH = paths;
```
erickzhao, line 45: Is overriding the existing `PATH` intentional here (as opposed to appending to it)? I see `fixPaths` did that anyways in its source code, though.

malept: Yes, it's intentional.

erickzhao (approved):

PR #664, opened 2021-04-20
refactor: merge default and custom editors

ckerr:
This is a large refactor with a few goals:
* Remove the distinction between 'custom' and 'default' editors. Happily this removes lots of logic paths, interfaces, enums, special cases, and so on. The only remaining special case is `main.js`, which is unavoidable because it's needed as a launch point when running and packaging the fiddle.
* Take most of the hairy mosaic & monaco management code from app.ts, state.ts, a half-dozen `utils/editor-*` files, and the command-editors / editor / and editor components, and move it into a single new class (`src/renderer/editor-mosaic`) where the ugliest stuff an be encapsulated and where the API visible to the rest of the app is simple, friendly, documented, and testable.
* Related to previous, improve separation of concerns / decoupling of modules. EditorMosaic _only_ knows about showing files in monaco editors placed on a mosaic. It doesn't know anything about where we got the files from (e.g. gist or file), about confirmation dialogs, etc. -- that is all managed elsewhere.

Review welcomed from any & all but especially @codebytere since this builds on her custom editor work. This keeps most of the 'custom' codepaths, e.g. in command-editors. In short, custom editors 'won' and everything is now a custom editor.

Supercedes https://github.com/electron/fiddle/pull/659 .


ckerr:
This is all handled in the other linked PRs now, so this draft PR can be closed :smile:

Issue #577, opened 2021-02-05
Intellisense not working


erickzhao:
This seems like a regression (was working in v0.15.0)

codebytere:
Pretty sure this is https://github.com/electron/electron/pull/27582 - that this is fixed when one disables `nodeIntegrationInWorker` and only shows up on macOS and Linux only seemingly confirms this

erickzhao:
Unfortunately, it seems like upgrading to a version of Electron that contains that fix (I tried v12.0.0) doesn't fix the issue. Intellisense still fails to load, and disabling `nodeIntegrationInWorker` fixes the problem.

codebytere:
@erickzhao i'll keep digging then - mind if i assign myself to this?

erickzhao:
@codebytere go for it!

Alternatively, I'm thinking of reverting https://github.com/electron/fiddle/commit/5e3a58896ebc771db0c7edbc10fecfbf30a1e48a and going back to the drawing board.

PR #667, opened 2021-04-22
fix: intellisense not working

codebytere:
Closes https://github.com/electron/fiddle/issues/577.

Fixes intellisense not working as a function of setting `nodeIntegrationInWorker: true` in the `BrowserWindow` options.

Ideally i'd like to narrow down the true fix but this shifts calls to `decomment` to `findModule` where they're invoked directly instead of via Worker thread - I think for now it's more important to regain this functionality while we search for a nicer long-term solution.


erickzhao (approved):
Seems legit

PR #669, opened 2021-04-22
test: improve `commands-action-button.tsx` coverage

ckerr:
no functional changes.

This is a refactor of `tests/renderer/components/commands-publish-button-spec.tsx` that fills in some of the edge cases, e.g. the description dialog and some error handling.

It reduces copy-paste code duplication by nesting 'describe' sections with beforeEach() functions, e.g. in `describe('publish mode'` there is a beforeEach() that creates a publish button ready to be tested.

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -263,7 +263,7 @@ export class GistActionButton extends React.Component<
       return this.handleUpdate();
     }
 
-    if (gistId && actionType == GistActionType.delete) {
+    if (gistId && actionType === GistActionType.delete) {
       return this.handleDelete();
     }
 
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -129,8 +129,8 @@ export class GistActionButton extends React.Component<
   private async publishGist(description: string) {
     const { appState } = this.props;
 
-    const octo = await getOctokit(this.props.appState);
-    const { gitHubPublishAsPublic } = this.props.appState;
+    const octo = await getOctokit(appState);
+    const { gitHubPublishAsPublic } = appState;
     const options = { includeDependencies: true, includeElectron: true };
     const values = await window.ElectronFiddle.app.getEditorValues(options);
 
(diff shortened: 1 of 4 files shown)


src/renderer/components/commands-action-button.tsx
```diff
@@ -259,21 +259,15 @@ export class GistActionButton extends React.Component<
...
-          await this.handleDelete();
-          break;
-      }
-    } else {
-      await this.handlePublish();
+    if (gistId && actionType === GistActionType.update) {
+      return this.handleUpdate();
+    }
+
+    if (gistId && actionType == GistActionType.delete) {
```
erickzhao, line 266: Is the `==` intentional here?

tests/renderer/components/commands-publish-button-spec.tsx
```diff
@@ -1,21 +1,45 @@
...
+    delete: jest.fn(),
+    update: jest.fn(),
+  };
+}
+
+describe('Action button component', () => {
+  const description = 'Electron Fiddle Gist';
+  const errorMessage = '💀';
+  let app: AppMock;
+  let mocktokit: OctokitMock;
```
erickzhao, line 38: hehe `mocktokit`

src/renderer/components/commands-action-button.tsx
ckerr: Nope. Fixed 9a1a55bc95fe4085fb28fc1df2220ae6cbd5de64

src/renderer/components/commands-action-button.tsx
```diff
@@ -259,21 +259,15 @@ export class GistActionButton extends React.Component<
...
-          break;
-      }
-    } else {
-      await this.handlePublish();
+    if (gistId && actionType === GistActionType.update) {
+      return this.handleUpdate();
+    }
+
+    if (gistId && actionType === GistActionType.delete) {
+      return this.handleDelete();
```
codebytere, line 267: this pattern is a little confusing since there's no return value to speak of here - what motivated this change?

ckerr: The main motivation for refactoring performGistAction() is to reduce the number of logic branches needed to reach the three possible outcomes of that function (publish, update, delete). However now that I reread it, I see three more branches can be removed while making the intent of the 'return' functions clearer. Revised in cff84ed.

The return values are the promises created by handlePublish(), handleDelete(), and handleDelete(). I don't feel super strongly about this, but my thinking was that by removing the unncessary await it avoids unnecessary microtasks.

codebytere (approved):

PR #670, opened 2021-04-24
test: reuse the Mock classes as much as possible

ckerr:
No functional changes. All changes are in `tests/`.

1. This PR tries to reuse our `tests/mocks/` classes as much as possible instead of reinventing the wheel whenever we need a temporary mock. There are many places in the tests that reinvent that wheel when there's a class in `tests/mocks/` that does the same thing and is already instantiated via the global `beforeEach()`.

2. Half of our mocks were named  MockFoo and the other half were named FooMock. This PR also picks one and uses it consistently so that you never have to guess.

3. Adds an umbrella `tests/mocks/mocks.ts` file from which you can import all the mocks you need instead of hitting each file in `mocks/` individually.

NB: **70% of the diffs are `__snapshots__` changes so don't let the diff size scare you! :smile:** The dialog-bisect snapshot alone makes up about 1/3 of this PR's diffs.

 --- 

Story time: the more-fully-fleshed-out mocks uncovered a few issues that grew `__snapshots__`. For example [the bisect dialog test used an array of 5 Electron versions](https://github.com/electron/fiddle/blob/master/tests/renderer/components/dialog-bisect-spec.tsx#L25) but [the production code starts at index #10](https://github.com/electron/fiddle/blob/master/src/renderer/components/dialog-bisect.tsx#L46). This never caused problems before because we never bounds-checked before. But since `tests/mocks/StateMock.versionsToShow` is a mobx observable (just as the production code's is), now the test would fail with:

> [mobx.array] Attempt to read an array index (10) that is out of bounds (5). Please check length first. Out of bound indices will not be tracked by MobX

We ought to address that 10 assumption in `src/` but this is a tests-only PR so it's out-of-scope for this PR.


codebytere (approved):

erickzhao (approved):
This is really clean! 🧼

PR #673, opened 2021-04-28
refactor: add editor-utils

ckerr:
Part 1 of breaking https://github.com/electron/fiddle/pull/664 into smaller, human-reviewable pieces.

This part lays a little bit of groundwork by adding editor-utils, which is a collection of helpers that can be used to decouple the rest of the code from `DefaultEditorId`s, `SORTED_EDITORS`, `DEFAULT_EDITORS`, and so on. But for now, their main use lies in making some existing pieces easier to use:

1. `getEmptyContent()` is a slightly friendlier form of `EMPTY_CONTENT` as clients no longer have to extract the file suffix out manually.
2. `getEditorTitle()` is a slightly friendlier form of `TITLE_MAP` as clients no longer have to handle the default editors & custom editors logic separately.
3. `isKnownFile()` returns true if it's one of the well-known files e.g. main.js or renderer.js. This is a slightly friendlier form of `Object.values(DefaultEditorId).includes()`
4. `isSupportedFile()` returns true if the file is js, html, or css. This isn't a new idea; it's just consolidated in once place so that there aren't duplicate implementations.
5. `compareEditors()` sorts the editors into mosaic order using SORTED_EDITORS order for the known editors, and then the rest are sorted lexicographically. Used by App when building the mosaic.

Also fixes a pre-existing bug in remote-loader's constructor that discarded its `bind(this)` functions after binding them.


src/renderer/remote-loader.ts
```diff
@@ -14,19 +10,25 @@ import { ELECTRON_ORG, ELECTRON_REPO } from './constants';
 import { getTemplate } from './content';
 import { AppState } from './state';
 import { getReleaseChannel } from './versions';
+import { isKnownFile, isSupportedFile } from '../utils/editor-utils';
 
 export class RemoteLoader {
   constructor(private readonly appState: AppState) {
-    this.loadFiddleFromElectronExample.bind(this);
```
codebytere, line 20: i'm totally ok with this change but i find it perhaps little less readable and less idiomatic/typical than what was there previously

codebytere (approved):

src/renderer/remote-loader.ts
ckerr: What we had previously didn't actually work did it? The new function with `this` bound was discarded instead of remembered. In order to work it would  need to be `this.loadFiddleFromElectronExample = this.loadFiddleFromElectronExample.bind(this)`

PR #674, opened 2021-04-29
refactor: remove unused Docs & Demos panels code

ckerr:
Part 2 of breaking https://github.com/electron/fiddle/pull/664 into smaller, human-reviewable pieces.

This part removes the code for the Docs & Demos experiment that was started but not finished a year or so ago. Some of the development code could be tried out by manually setting a `FIDDLE_DOCS_DEMOS` environment variable but it didn't have any working features.

Side-effects:
1. The `PanelId` type goes away, since its only value was `PanelId.docsDemo`.
2. `MosaicId` goes away, since it was defined as `EditorId + PanelId` and is now redundant since it equals `EditorId`.
3. `ALL_PANELS` goes away, since it only held `PanelId.docsDemo`.
4. `ALL_MOSAICS` goes away, since it was defined as `DEFAULT_EDITORS + ALL_PANELS` and is now redundant since it equals `DEFAULT_EDITORS`.
5. `isEditorId()` and `isPanelId()` go away, since they were only used to differentiate between editors and the now-nonexistent panels.
6. `closedPanels` was defined as a `Record<Id, EditorBackup|true>` where `EditorBackup` was used for editors and `true` was used for panels. Now that panels are gone, `closedPanels` is now `Record<id, EditorBackup>`.
7. `isBackupEditor()` goes away since it was only used to test whether a value in `closedPanels` was an `EditorBackup` or `true`. Since this was the last remaining function in `type-checks`, that file and its spec also go away.


codebytere (approved):
Screen Shot 2021-04-29 at 10 27 17 AM

satisfied-pacha

PR #675, opened 2021-04-29
feat: allow setting environment variables

codebytere:
Adds a new settings preference similar to that for command line switches that allows for custom environment variables to pass to the executable. 

Example use case - debugging drag regions. 

Screenshot:

Screen Shot 2021-04-29 at 1 47 42 PM


tests/renderer/components/settings-execution-spec.tsx
```diff
@@ -71,4 +74,29 @@ describe('ExecutionSettings component', () => {
       ]);
     });
   });
+
+  describe('handleEnvironmentVariableChange()', () => {
+    it('handles new environment variables', async () => {
+      const wrapper = shallow(<ExecutionSettings appState={store} />);
+      const instance = wrapper.instance() as any;
+      await instance.handleEnvironmentVariableChange({
+        currentTarget: { value: 'ELECTRON_DEBUG_DRAG_REGIONS=1' },
```
ckerr, line 83: _nit_ not a blocker, but a suggestion for cases like this: instead of using the same magic string in two places, declare `ELECTRON_DEBUG_DRAG_REGIONS=1` as a local const and use it in the invocation and in the expect.

And if you name the variable `value` then the invocation becomes `currentTarget: { value }` which reads kind of pleasingly.

ckerr (approved):

src/renderer/state.ts
```diff
@@ -109,6 +109,10 @@ export class AppState {
     (this.retrieve('executionFlags') as Array<string>) === null
       ? []
       : (this.retrieve('executionFlags') as Array<string>);
+  @observable public environmentVariables: Array<string> =
+    (this.retrieve('environmentVariables') as Array<string>) === null
+      ? []
+      : (this.retrieve('executionFlags') as Array<string>);
```
erickzhao, line 115:
```suggestion
      : (this.retrieve('environmentVariables') as Array<string>);
```

Was this a typo?

erickzhao (changes requested):

src/renderer/state.ts
codebytere: oops yes

erickzhao (approved):
This looks good!

I think one future improvement for this and the execution flag setting is that their UI could be more graphical than a bar-separated string.

codebytere:
@erickzhao i do too - ill probably work on that next week!

Issue #668, opened 2021-04-22
[Bug]: console.log output not shown


xmedeko:
When I put `console.log` into:
```js
app.on('window-all-closed', function () {
  console.log('window-all-closed')
  app.quit()
})
```
Then, the log is not visible in the Fiddle's output window. The log shows after I start the app again, see (some logs removed):
```
Electron v12.0.5 started.
Electron exited with code 0.
Electron v12.0.5 started.
window-all-closed
```

erickzhao:
I feel like I've ran into this before on Windows. Is that the OS you're using @xmedeko? Thanks for the report!

xmedeko:
Yes, Window 10, 64 bit.

codebytere:
I also see this fairly often - it looks like stdout is just flushing improperly?

ckerr:
@codebytere what is this also on Windows that you see it, or  elsewhere?

[this](https://github.com/electron/fiddle/blob/master/src/renderer/state.ts#L629) looks suspicious but I don't know if that is literally this bug report or if it's just adjacent?

codebytere:
@ckerr windows-only

PR #676, opened 2021-04-29
fix: console output being dropping on Windows

codebytere:
Closes https://github.com/electron/fiddle/issues/668.

Refs https://github.com/electron/fiddle/commit/5c98534cd89fc4cabbe4512ee9d758696b09882a.

This was previously dropping output if `parts.length` was 1. Now it doesn't.

### DIFF
--- a/src/renderer/runner.ts
+++ b/src/renderer/runner.ts
@@ -327,6 +327,7 @@ export class Runner {
     const {
       currentElectronVersion,
       isEnablingElectronLogging,
+      flushOutput,
       pushOutput,
       executionFlags,
       environmentVariables,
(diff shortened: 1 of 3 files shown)


ckerr (approved):

MarshallOfSound (changes requested):
I don't think this is fully correct, this has the side effect of fixing it if only a single chunk is dispatched at a time.  If multiple chunks are dispatched we'll still leave the buffer hanging without flushing it.

We should change this from using a buffer, to writing to the previous log line when required and always flush.

MarshallOfSound (approved):

PR #678, opened 2021-04-30
feat: better execution settings UI

codebytere:
Refs https://github.com/electron/fiddle/pull/675#pullrequestreview-648760711.

This reworks the execution settings UI for execution flags and environment variables to something a bit more user-friendly. 

Screen Shot 2021-05-03 at 10 41 21 AM

### DIFF
--- a/src/renderer/components/settings-execution.tsx
+++ b/src/renderer/components/settings-execution.tsx
@@ -1,4 +1,6 @@
 import {
+  Button,
+  ButtonGroup,
   Callout,
   Checkbox,
   FormGroup,
@@ -12,8 +14,18 @@ import * as React from 'react';
 import { IPackageManager } from '../npm';
 import { AppState } from '../state';
 
+export enum SettingItemType {
+  EnvVar,
+  Flag,
+}
+
 interface ExecutionSettingsProps {
   appState: AppState;
+}
+
+interface ExecutionSettingsState {
+  executionFlags: Record<string, string>;
+  environmentVariables: Record<string, string>;
 }
 
 /**
@@ -23,18 +35,48 @@ interface ExecutionSettingsProps {
  * @extends {React.Component<ExecutionSettingsProps, {}>}
  */
 @observer
-export class ExecutionSettings extends React.Component<ExecutionSettingsProps> {
+export class ExecutionSettings extends React.Component<
+  ExecutionSettingsProps,
+  ExecutionSettingsState
+> {
   constructor(props: ExecutionSettingsProps) {
     super(props);
+
+    this.state = {
+      executionFlags: Object.assign(
+        {},
+        ...props.appState.executionFlags.map((flag, idx) => {
+          return { [`flag-${idx}`]: flag };
+        }),
+      ),
+      environmentVariables: Object.assign(
+        {},
+        ...props.appState.environmentVariables.map((envVar, idx) => {
+          return { [`var-${idx}`]: envVar };
+        }),
+      ),
+    };
 
     this.handleDeleteDataChange = this.handleDeleteDataChange.bind(this);
     this.handleElectronLoggingChange = this.handleElectronLoggingChange.bind(
       this,
     );
-    this.handleExecutionFlagChange = this.handleExecutionFlagChange.bind(this);
-    this.handleEnvironmentVariableChange = this.handleEnvironmentVariableChange.bind(
-      this,
-    );
+
+    this.handleSettingsItemChange = this.handleSettingsItemChange.bind(this);
+    this.handleSettingsItemSave = this.handleSettingsItemSave.bind(this);
+    this.addNewSettingsItem = this.addNewSettingsItem.bind(this);
+  }
+
+  public componentDidMount() {
+    const { environmentVariables, executionFlags } = this.state;
+
+    if (Object.keys(executionFlags).length === 0) {
+      this.addNewSettingsItem(SettingItemType.Flag);
+    }
+
+    if (Object.keys(environmentVariables).length === 0) {
+      this.addNewSettingsItem(SettingItemType.EnvVar);
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


erickzhao (commented):
Looking at the screenshot, would it make sense that the save event happens in the `handleEnvironmentVariableChange` handler rather than in its own button?

src/renderer/components/settings-execution.tsx
```diff
@@ -59,36 +101,201 @@ export class ExecutionSettings extends React.Component<ExecutionSettingsProps> {
...
+
+  public renderDeleteItem(idx: string, type: SettingItemType): JSX.Element {
+    const clickFn = () => {
+      if (type === SettingItemType.Flag) {
+        const flags = this.state.executionFlags;
+        if (Object.keys(flags).length === 1) {
+          flags[idx] = '';
+        } else {
+          delete flags[idx];
+        }
```
ckerr, line 200: I don't understand why there's special logic for not deleting the last property? Do we need to avoid an empty object {} for some reason?

I'm not arguing pro or con on this; literally asking because I don't see why the special case is needed

src/renderer/components/settings-execution.tsx
```diff
@@ -59,36 +101,201 @@ export class ExecutionSettings extends React.Component<ExecutionSettingsProps> {
...
+        },
+      }));
+    } else {
+      this.setState((prevState) => ({
+        environmentVariables: {
+          ...prevState.environmentVariables,
+          [name]: value,
+        },
+      }));
+    }
```
ckerr, line 149: Nearly all the changes in settings-execution.tsx contain copypaste between handling flags and envvars. Would it make sense to extract out an abstract version of this and have both flags & envvars use that instead of repeating it?

src/renderer/components/settings-execution.tsx
```diff
@@ -59,36 +101,201 @@ export class ExecutionSettings extends React.Component<ExecutionSettingsProps> {
...
+      } else {
+        const vars = this.state.environmentVariables;
+        if (Object.keys(vars).length === 1) {
+          vars[idx] = '';
+        } else {
+          delete vars[idx];
+        }
+
+        this.setState({
+          environmentVariables: vars,
```
ckerr, line 214: It _looks_ like we're calling setState() with the same variable that's already in state (ie this.state.environmentVariables). Is this call to setState() redundant?

src/renderer/components/settings-execution.tsx
```diff
@@ -12,29 +14,69 @@ import * as React from 'react';
...
   constructor(props: ExecutionSettingsProps) {
     super(props);
 
+    this.state = {
+      executionFlags: Object.assign(
+        {},
+        ...props.appState.executionFlags.map((flag, idx) => {
+          return { [`flag-${idx}`]: flag };
+        }),
+      ),
```
ckerr, line 51: Is Object.assign needed?

```suggestion
      executionFlags: {
        ...props.appState.executionFlags.map((flag, idx) => {
          return { [`flag-${idx}`]: flag };
        }),
      },
```

src/renderer/components/settings-execution.tsx
```diff
@@ -12,29 +14,69 @@ import * as React from 'react';
...
+      environmentVariables: Object.assign(
+        {},
+        ...props.appState.environmentVariables.map((envVar, idx) => {
+          return { [`var-${idx}`]: envVar };
```
ckerr, line 55: Converting an array to an object with keys of `var-${idx}` is unintuitive. Is the object form necessary, or could we skip the intermediary and use the arrays here instead?

tests/renderer/components/settings-execution-spec.tsx
```diff
@@ -54,47 +57,104 @@ describe('ExecutionSettings component', () => {
...
   });
 
-  describe('handleExecutionFlagChange()', () => {
-    it('handles new flags', async () => {
-      const wrapper = shallow(<ExecutionSettings appState={store} />);
-      const instance = wrapper.instance() as any;
-      await instance.handleExecutionFlagChange({
-        currentTarget: { value: '--lang=es' },
+  describe('handleItemSettingsChange()', () => {
+    describe('with executionFlags', () => {
```
ckerr, line 61: This is more of a personal preference than a strong assertion, but these two tests repeat the same strings a lot e.g. `--lang=es` and `--js-flags=--expose-gc`. One alternative to this repetition would be to make these `const` describe-level variables that the different `it()` tests can use.

src/renderer/components/settings-execution.tsx
codebytere: ah yeah i could just do `{ [idx]: envVar }` - the limitation though is that we need to know when we're _updating_ a flag/var at a specific index and not adding a new flag/var; there might be a better way to do that but that's what motivated this

src/renderer/components/settings-execution.tsx
codebytere: hmm - not sure what you mean, since we're modifying that object directly above this 🤔

src/renderer/components/settings-execution.tsx
codebytere: yes - those code blocks aren't equivalent :)

Screen Shot 2021-05-01 at 10 50 22 AM

src/renderer/components/settings-execution.tsx
codebytere: cleaned this up a bit :)

src/renderer/components/settings-execution.tsx
codebytere: @ckerr if a user is removing the only item, i don't want to not show any input fields - open to other ways to do this though! instead  i could disable the remove button for example?

src/renderer/components/settings-execution.tsx
ckerr: Right we are, but when we modify `this.state.environmentVariable` isn't that an `@observable` such that the component already knows about the modifications we made even without calling setState()?

src/renderer/components/settings-execution.tsx
ckerr: That's a good idea; I like that better.

src/renderer/components/settings-execution.tsx
codebytere: @ckerr no - this class-level state is not an observable like that on `appState` :)

src/renderer/components/settings-execution.tsx
```diff
@@ -12,29 +14,76 @@ import * as React from 'react';
...
+      this.addNewSettingsItem(SettingItemType.Flags);
+    }
+
+    if (Object.keys(environmentVariables).length === 0) {
+      this.addNewSettingsItem(SettingItemType.EnvVars);
+    }
+  }
+
+  public componentDidUpdate() {
+    const { appState } = this.props;
```
codebytere, line 81: This needs to be done here because a call to `setState` queues a _pending_ state transition and as such isn't synchronous. So - if we update `appState` with `this.state` directly after a call to `setState`, the state is nontrivially likely to be the previous state.

ckerr (approved):

erickzhao (approved):
LGTM!

PR #679, opened 2021-04-30
refactor: add EditorMosaic class

ckerr:
Part 4 of breaking #664 into smaller, human-reviewable pieces.

This one is a very incremental change: replace `window.ElectronFiddle.editors` with `appState.editorMosaic`, which is a new class which just holds a `Map<EditorId, Monaco.editor.IStandaloneCodeEditor`. That's all! global -> non-global + Record -> Map. The important thing for future parts is that now there is a logical place to start consolidating mosaic code.

Structurally this is a _really_ simple PR. The diff is because snapshots feel every ripple in the mocks, and we've added a new class to the mocks. So don't be afraid of the diff size; this isn't 2K LOC being added :smile_cat: 

Biggest 10 file changes:

```
17	2	tests/mocks/state.ts
18	5	tests/utils/editor-backup-spec.ts
21	26	src/renderer/app.tsx
24	41	tests/renderer/app-spec.tsx
25	36	tests/utils/editor-value-spec.ts
27	23	tests/utils/focused-editor-spec.ts
100	0	tests/renderer/components/__snapshots__/commands-version-chooser-spec.tsx.snap
200	0	tests/renderer/components/__snapshots__/editors-spec.tsx.snap
500	0	tests/renderer/components/__snapshots__/commands-spec.tsx.snap
1000	0	tests/renderer/components/__snapshots__/dialog-bisect-spec.tsx.snap
```


src/ambient.d.ts
```diff
@@ -10,10 +10,6 @@ declare global {
       appPaths: Record<string, string>;
       app: AppType;
       contentChangeListeners: Array<any>;
-      editors: Record<
```
codebytere, line 13: Looks like the imports got left around - we should remove:

```ts
import * as MonacoType from 'monaco-editor';
```
and
```ts
import { EditorId } from './interfaces';
```

codebytere (commented):
nice cleanup :D these changes are moving things in a great direction imo!

src/ambient.d.ts
ckerr: Fixed in 306573e19585ce66bf824cd4ae75ae3024367d9d

codebytere (approved):

Issue #650, opened 2021-04-10
BrowserWindow with `closable: false` doesn't exit when stopping the Fiddle example


erickzhao:
I'm not sure if this is technically a bug because this seems like `closable: false` just doing what it's meant to do, but it's annoying to stop the Fiddle example and have to Force Quit the process.

PR #688, opened 2021-05-06
fix: properly kill dangling processes

codebytere:
Closes https://github.com/electron/fiddle/issues/650.

I'd love some feedback on potential better ways to do this if y'all have them!


src/renderer/runner.ts
```diff
@@ -379,14 +389,11 @@ export class Runner {
...
         if (typeof code !== 'number') {
-          pushOutput('Electron exited.');
+          pushOutput(`Electron exited with signal ${signal}`);
           resolve(RunResult.INVALID);
-        } else if (!code) {
-          pushOutput(`Electron exited with code ${code}.`);
-          resolve(RunResult.SUCCESS);
         } else {
           pushOutput(`Electron exited with code ${code}.`);
-          resolve(RunResult.FAILURE);
```
codebytere, line 389: just cleaned this up a bit and made the logging a bit better!

src/renderer/runner.ts
```diff
@@ -203,8 +204,17 @@ export class Runner {
...
-    return isIdle;
+
+    // If the child process is still alive 1 second after we've
+    // attempted to kill it by normal means, kill it forcefully.
+    setTimeout(() => {
+      const pid = this.child?.pid;
+      if (pid && isRunning(pid)) {
+        this.child?.kill('SIGKILL');
+      }
+    }, 1000);
```
ckerr, line 215: This makes the return value of this function kind of weird, since the true state could change right after this value is returned. Maybe this should return a promise so that users can await the stop iff they want, and just invoke without awaiting if that's all they need.

Also, it looks like the boolean being returned is not checked anywhere, so Maybe this should return `Promise<void>`

ckerr (commented):
I like this change!

Suggestion inline about the return value of Runner.stop() but other than that LGTM

src/renderer/runner.ts
codebytere: this is just used in the click handler for stop/start, so i think we can just......remove the return value maybe 🤔

ckerr: I've a slight preference for `Promise<void>` because that's what I would expect to see from a method named "stop()" but... **nobody currently calling this function cares** so YAGNI wins! :smile_cat:

ckerr (approved):

PR #697, opened 2021-05-28
feat: allow obsolete versions to be toggled in cli

ckerr:
When bisecting from the command line, it's possible that the last known good version is an obsolete release. Thus, it's necessary to be able to re-enable obsolete releases from the command line.

### DIFF
--- a/src/renderer/task-runner.ts
+++ b/src/renderer/task-runner.ts
@@ -27,6 +27,7 @@ export class TaskRunner {
   private readonly run: () => Promise<RunResult>;
   private readonly setVersion: (ver: string) => Promise<void>;
   private readonly show: (channels: ElectronReleaseChannel[]) => Promise<void>;
+  private readonly showObsoleteVersions: (show: boolean) => void;
 
   constructor(app: App) {
     const { runner, state } = app;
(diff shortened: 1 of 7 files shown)


src/renderer/task-runner.ts
```diff
@@ -102,7 +105,11 @@ export class TaskRunner {
 
   private async setup(req: SetupRequest) {
     const { log } = this;
-    const { fiddle, hideChannels, showChannels, version } = req;
+    const { fiddle, hideChannels, showChannels, useObsolete, version } = req;
+
+    if (useObsolete === true || useObsolete === false) {
```
clavin, line 110: Isn't this... always true?

clavin (changes requested):

src/renderer/task-runner.ts
ckerr: no. If the user doesn't specify it on the command line, its value is undefined.

I guess I could use `typeof useObsolete == 'boolean'` to be clearer?

clavin (approved):
👍

erickzhao (approved):

PR #698, opened 2021-05-28
fix: exit with err if the CLI is given a bad gist

ckerr:
Ensure that we error out quickly if the command line is fed an unparseable gist.


erickzhao (approved):
Code LGTM :)

Issue #316, opened 2020-02-03
Implement Recent Documents menu


erickzhao:
Electron has an API that allows users to open recently-opened documents directly from your dock menu or app menu.

See this API: https://www.electronjs.org/docs/api/app#appaddrecentdocumentpath-macos-windows

It would be good to get this hooked up for local Fiddle folders.

PR #700, opened 2021-05-31
feat: add recentDocuments to fiddle

georgexu99:
fixes: #316

This adds functionality to open and clear recent documents- either through the 'file' menu or the app context menu.

image

image

### DIFF
--- a/src/main/protocol.ts
+++ b/src/main/protocol.ts
@@ -92,6 +92,12 @@ export const listenForProtocolHandler = () => {
       handlePotentialProtocolLaunch(url);
     }
   });
+  app.on('open-file', (_, url) => {
+    if (!url || url.length < 1) {
+      return;
+    }
+    ipcMainManager.send(IpcEvents.FS_OPEN_FIDDLE, [url]);
+  });
 
   // pass protocol URL via npm start args in dev mode
   if (isDevMode() && process.env.npm_config_argv) {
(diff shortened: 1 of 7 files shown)


VerteDinde (approved):
Looks great! Working on multiple windows and working with multiple recent folders - excited to use this, this will make cycling through multiple bugs much easier 🙂

src/main/protocol.ts
```diff
@@ -92,6 +92,12 @@ export const listenForProtocolHandler = () => {
       handlePotentialProtocolLaunch(url);
     }
   });
+  app.on('open-file', (_, path) => {
+    if (!path || path.length < 1) {
+      return;
+    }
+    ipcMainManager.send(IpcEvents.FS_OPEN_FIDDLE, [path]);
+  });
```
erickzhao, line 100: Why was this handler necessary here? I'm not very familiar with the `open-file` event.

georgexu99: When you click on a file from either the menu or the context menu, a open-file event is [fired](https://www.electronjs.org/docs/tutorial/recent-documents). So we need this handler in order to do something with the event, in this case open a new fiddle w/ the specified path

otherwise nothing happens when you try to open a recent file

georgexu99: ^ the open-file tidbit is mentioned at the very bottom of the linked page

erickzhao (changes requested):
Seems like some menu tests are failing

erickzhao (approved):

PR #702, opened 2021-06-01
chore: release v0.23.2

erickzhao:


ckerr (changes requested):
It would be great if we could get https://github.com/electron/fiddle/pull/697 into the upcoming release since it's needed for bugbot. It's currently unmerged & needs review

Also it's semver minor, so a release including it should be called 0.24.0

ckerr (approved):

PR #705, opened 2021-06-03
feat: add checkmarks to "Show Me" menu to indicate which example is open

georgexu99:
issue: #282 

## Description
add checkmarks to indicate which fiddle is opened within the "Show Me" examples tab

## Acceptance Criteria
[x] Add tick that displays in menu per Fiddle Window
[x] tick removes itself if user saves or opens file (meaning they created/opened a local version)
[x] Display tick per Fiddle Window (if I have more than one Fiddle open, should change per Fiddle)
[x] Add tests

image


erickzhao (changes requested):
Looks like tests are failing here

codebytere (approved):

tests/main/menu-spec.ts
```diff
@@ -127,14 +127,47 @@ describe('menu', () => {
...
+
+        const result = (electron.Menu.buildFromTemplate as any).mock
+          .calls[0][0];
+        const showMeMenu = result[result.length - 2];
+
+        let showMeItemChecked = true;
+        if (showMeMenu.submenu[0].submenu[0]) {
+          showMeItemChecked = showMeMenu.submenu[0].submenu[0].checked;
+        }
+        expect(showMeItemChecked).toEqual(false);
```
erickzhao, line 144: This can be simplified to be less verbose:

```suggestion
        const showMeItemChecked = showMeMenu?.submenu[0]?.submenu[0]?.checked;
        expect(showMeItemChecked).toBe(false);
```

The biggest change here is the use of the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (`?.`) to drill into possibly undefined object properties.

It's also unnecessary to set the variable to `true` initially because `showMeItemChecked` will be `undefined` if the optional chaining fails, and `expect(undefined).toBe(false)` will throw.

tests/main/menu-spec.ts
```diff
@@ -127,14 +127,47 @@ describe('menu', () => {
...
+        setupMenu({ activeTemplate: 'App' });
+
+        const result = (electron.Menu.buildFromTemplate as any).mock
+          .calls[0][0];
+        const showMeMenu = result[result.length - 2];
+
+        let showMeItemChecked = false;
+        if (showMeMenu.submenu[0].submenu[0]) {
+          showMeItemChecked = showMeMenu.submenu[0].submenu[0].checked;
+        }
```
erickzhao, line 159: Same as above

src/main/main.ts
```diff
@@ -33,7 +33,7 @@ export async function onReady() {
   const { setupMenu } = await import('./menu');
   const { setupFileListeners } = await import('./files');
 
-  setupMenu();
+  setupMenu({});
```
erickzhao, line 36: We could probably make the options object optional here, and make `setupMenu()` behaviour the same as `setupMenu({})`.

erickzhao (approved):

PR #711, opened 2021-06-15
fix: Handle protocol links on Windows

felixrieseberg:
Turns out, we never handled protocol links on Windows. I KNOW. I can't quite believe it, either.

### DIFF
--- a/src/main/protocol.ts
+++ b/src/main/protocol.ts
@@ -92,6 +92,13 @@ export const listenForProtocolHandler = () => {
       handlePotentialProtocolLaunch(url);
     }
   });
+
+  app.removeAllListeners('second-instance');
+  app.on('second-instance', (_event, commandLine, _workingDirectory) => {
+    // Someone tried to run a second instance
+    scanArgv(commandLine);
+  });
+
   app.on('open-file', (_, path) => {
     if (!path || path.length < 1) {
       return;
--- a/tests/main/protocol-spec.ts
+++ b/tests/main/protocol-spec.ts
@@ -41,6 +41,20 @@ describe('protocol', () => {
   });
 
   describe('listenForProtocolHandler()', () => {
+    it('handles a Fiddle url (second-instance)', () => {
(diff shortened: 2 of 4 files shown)


src/main/protocol.ts
```diff
@@ -92,6 +92,13 @@ export const listenForProtocolHandler = () => {
       handlePotentialProtocolLaunch(url);
     }
   });
+
+  app.removeAllListeners('second-instance');
```
MarshallOfSound, line 96: Why do we need to remove all listeners here?

felixrieseberg: I actually have no idea and just followed "tradition". Maybe testing? That'd make sense to me.

erickzhao: "Testing" seems about right! I think we were registering _a lot_ of event listeners during tests and getting warnings from Node https://github.com/electron/fiddle/pull/329

tests/main/protocol-spec.ts
```diff
@@ -41,6 +41,20 @@ describe('protocol', () => {
   });
 
   describe('listenForProtocolHandler()', () => {
+    it('handles a Fiddle url (second-instance)', () => {
+      overridePlatform('win32');
+
+      listenForProtocolHandler();
+
+      const handler = (app.on as any).mock.calls[1][1];
```
erickzhao, line 49: nit: can we assert somewhere that `app.on` was also called with the `second-instance` event parameter?

erickzhao (approved):
Code LGTM

tests/main/protocol-spec.ts
STEMZANATOR:
```suggestion
      const handler = (app.on as any).mock.calls[1][1];
```

src/main/protocol.ts
ckerr: Agree, that's my understanding too about why we remove all listeners

Issue #709, opened 2021-06-10
Console Loses Next Message After Clear


dsanders11:
Seems like after clearing the console, the next message is lost. Seeing this on macOS.

Here's a test to highlight the behavior. Create a new fiddle, and in `main.js` set up an interval to output every 5 seconds. Clear the console right after a message, and then 10 seconds later you'll see another, showing that one of them was lost.

```javascript
setInterval(() => console.log('tick tock'), 5000);
```

PR #715, opened 2021-06-16
fix: don't exclude first line on clear

codebytere:
Closes https://github.com/electron/fiddle/issues/709.

```js
const lines = output
  .slice(Math.max(output.length - 1000, 1))
  .map(this.renderEntry);
```

excludes the first element, which we wouldn't want to do especially when clearing Fiddles as it would lead to single dropped lines. Correct that.

### DIFF
--- a/src/renderer/components/output.tsx
+++ b/src/renderer/components/output.tsx
@@ -137,9 +137,7 @@ export class Output extends React.Component<CommandsProps> {
     }
 
     // The last 1000 lines
-    const lines = output
-      .slice(Math.max(output.length - 1000, 1))
-      .map(this.renderEntry);
+    const lines = output.slice(-1000).map(this.renderEntry);
 
     return (
       <div className="output" ref={this.outputRef}>
--- a/tests/renderer/components/__snapshots__/output-spec.tsx.snap
+++ b/tests/renderer/components/__snapshots__/output-spec.tsx.snap
@@ -5,7 +5,22 @@ exports[`Output component renders with output 1`] = `
   className="output"
 >
   <div
-    key="1532704073130--0--0"
+    key="1532704072127--0--0"
+  >
+    <span
+      className="output-message"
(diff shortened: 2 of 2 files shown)


erickzhao (commented):
Seems like a snapshot test is failing with this change 😅

dsanders11:
@codebytere, I think this can be simplified to just `.slice(-1000)`? Should be functionally equivalent and a bit easier to read, IMO.

Probably better for a follow-up issue/PR, but is there a reason to do the 1000 line max in the display, rather than when storing it in the output array? Is there another use for the output array other than display? I have noticed before in cases where there's a lot of console output that Fiddle will get sluggish until I clear the console, presumably because that array is getting quite large and causing garbage collection to kick on constantly. If the array is only for display then the max lines logic could be moved to where lines of output are added, which would prevent that.

ckerr (approved):
:+1: on this line fixing the existing bug. I agree with @dsanders11 about `.slice(-1000)` as being a slightly nicer improvement

Also I like David's point about trimming the output array pre-display just to avoid excess memory use, but that seems like a bigger scope that might make more sense as a standalone issue/PR

src/renderer/components/output.tsx
```diff
@@ -138,7 +138,7 @@ export class Output extends React.Component<CommandsProps> {
 
     // The last 1000 lines
     const lines = output
-      .slice(Math.max(output.length - 1000, 1))
+      .slice(Math.max(output.length - 1000, 0))
```
ckerr, line 141:
```suggestion
      .slice(-1000)
```

Suggested by @dsanders11

ckerr, line 141:
```suggestion
      .slice(-1000)
```

erickzhao (approved):

erickzhao:
LGTM pending test fix

Issue #317, opened 2020-02-04
Make console searchable


erickzhao:
📑 As an Electron maintainer, the logs coming from Fiddle's console are useful for debugging purposes.

🔎 Being able to search through these with <kbd>⌘ F</kbd> would be great, especially when advanced Electron logging is enabled in the Chromium settings.

💡 One possible way to approach this would be to refactor the `<Output/>` class into a read-only Monaco editor and pushing new lines of the log through the Monaco Editor APIs.

cc @loc

dsanders11:
@erickzhao, I think this can be closed, since #717 got merged?

erickzhao:
Thank you!

PR #717, opened 2021-06-16
feat: make Fiddle console searchable

georgexu99:
#317 

This PR refactors our Output component to make it a read-only Monaco editor to allow for cmd/ctr-f functionality.

- [x] Refactored the Output component to include read-only Monaco editor
- [x] Refactored tests to expand coverage
- [x] Validated on Mac, (needs to be tested on Windows and Linux)


georgexu99:
Updated to be inline with #715

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
+      });
+
+      await this.editorDidMount(this.editor);
     }
   }
 
   /**
+   * Destroy Monaco.
+   */
+  public destroyMonaco() {
```
ckerr, line 110: _nit_ I'd have a slight preference to a name like `destroyMonacoEditor()` or `destroyEditor()` rather than `destroyMonaco()` because of the strange hoops we jump through to load the `monaco` object, so it's slightly clearer to not imply that we're destroying that `monaco` object

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
+    if (typeof this.editor !== 'undefined') {
+      console.log('Editor: Disposing');
+      this.editor.dispose();
```
ckerr, line 113: Is it OK to `delete this.editor;` here after calling dispose()? Or does the property get referenced elsewhere after this point s.t. we need to keep it?

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
+  /**
+   * Set up Monaco Language.
+   */
+  private setupMonacoLanguage(monaco: typeof MonacoType) {
+    // register a new language
+    monaco.languages.register({ id: 'consoleOutputLanguage' });
+    // Register a tokens provider for the language
+    monaco.languages.setMonarchTokensProvider('consoleOutputLanguage', {
+      tokenizer: {
+        root: [[/\d{1,2}:\d{2}:\d{2}\s(A|P)M/, 'custom-date']],
```
ckerr, line 158: This worked out pretty well!  This is cool!

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
+        } else if (mosaicTree.splitPercentage === 0) {
+          mosaicTree.splitPercentage = 25;
+        }
+        this.context.mosaicActions.replaceWith([], mosaicTree);
+      }
+    }
+  }
+
+  /**
+   * Set up Monaco Language.
```
ckerr, line 150: This comment can be removed; it literally repeats the function name :smiley_cat:

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
+   */
+  private setupMonacoLanguage(monaco: typeof MonacoType) {
```
ckerr, line 152: This could be more descriptive about what language we're setting up, e.g. explaining that we're tokenizing the date.

Hmm, maybe something like `setupDataDelimitedLanguage()`? That feels a little clumsy so don't feel compelled to use that verbatim, but something more descriptive than 'setupMonacoLanguage`

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
-            {text}
-          </span>
-        </div>
-      );
-
-    return lines.map(renderLine);
+  private getOutputLines(output: OutputEntry[]) {
+    const { appState } = this.props;
+    const lines: string[] = [];
+    const outputs = output.slice(appState.output.length - 1000);
```
ckerr, line 188:
```suggestion
    const outputs = output.slice(-1000);
```

src/renderer/components/output.tsx
```diff
@@ -30,121 +33,195 @@ interface CommandsProps {
...
 
-  public render(): JSX.Element | null {
-    const { output, isConsoleShowing } = this.props.appState;
+  /**
+   * Create a model and attach it to the editor
+   *
+   * @private
+   * @param {string} value
+   */
+  private createModel(value: string) {
```
ckerr, line 206: This function name doesn't reflect that the created model gets attached to `this.editor`, so it scans oddly to the eye when reading the code where it's called.

Maybe `setEditorText()`?

```suggestion
  private setEditorText(value: string) {
```

src/renderer/themes-defaults.ts
```diff
@@ -58,7 +58,7 @@ export const defaultDark: LoadedFiddleTheme = {
   editor: {
     base: 'vs-dark',
     inherit: true,
-    rules: [],
+    rules: [{ token: 'custom-date', foreground: '008800' }],
```
ckerr, line 61: Should 008800 be prefixed with a `#` here and below?

Asking because I didn't know it could be omitted! :slightly_smiling_face:  If this is legal, then cool

tests/mocks/app.ts
```diff
@@ -46,15 +46,19 @@ export class AppMock {
...
     },
     languages: {
       typescript: {
         javascriptDefaults: {
           addExtraLib: jest.fn(),
         },
       },
+      register: jest.fn(),
+      setMonarchTokensProvider: jest.fn(),
+      registerCompletionItemProvider: jest.fn(),
```
ckerr, line 61: At some point this custom object is going to be replaced with `new MonacoEditorMock()` so please add these 5 additions into `tests/mocks/monaco-editor` too :bowing_man:

ckerr (commented):
This is a pretty cool feature!

src/renderer/components/output.tsx
georgexu99: setupCustomOutputEditorLanguage? :P

src/renderer/components/outputs.tsx
```diff
@@ -0,0 +1,87 @@
...
+      this.setState({
+        monaco,
+        monacoOptions: defaultMonacoOptions,
+      });
+    } else {
+      this.setState({ monaco });
+    }
+
+    await activateTheme(monaco, undefined, this.props.appState.theme);
+  }
```
erickzhao, line 86: I see this bit is copy/pasted from the `editors.tsx` file. Can we re-use a helper function to avoid code duplication?

src/renderer/components/outputs.tsx
```diff
@@ -0,0 +1,87 @@
...
+@observer
+export class Outputs extends React.Component<OutputsProps, OutputsState> {
+  constructor(props: OutputsProps) {
+    super(props);
+
+    this.state = { monacoOptions: defaultMonacoOptions };
+  }
+
+  /**
+   * Executed right after the component mounts. We'll setup the IPC listeners here.
```
erickzhao, line 35: Are there IPC listeners here?

src/renderer/components/outputs.tsx
```diff
@@ -0,0 +1,87 @@
+import { observer } from 'mobx-react';
```
erickzhao, line 1: Question: What's the purpose of the `Outputs` wrapper component?

erickzhao (changes requested):

src/renderer/components/outputs.tsx
georgexu99: oops, this snuck in before I deleted some code :P

src/renderer/components/outputs.tsx
georgexu99: Outputs abstracts a lot of the monaco setup/loading behaviour- I made this class after needing a way to pass in monaco and monaco options to output.tsx

src/renderer/components/outputs.tsx
georgexu99: Tried tackling this- unfortunately since this.setState is dependant on each individual component we can't abstract it out- the rest of the function is boiler code

src/renderer/components/outputs.tsx
georgexu99: addressed in newest commit

src/renderer/components/outputs.tsx
erickzhao: That's fine for now, I think. Would you mind adding a `FIXME` comment here that clarifies that we should attempt to de-dupe this code anyways?

erickzhao (approved):
LGTM pending my last comment!

src/renderer/components/outputs.tsx
georgexu99: yes!

src/renderer/components/output.tsx
ckerr: Oops, I didn't realize this was inherited from the older code! :dizzy_face:

ckerr (commented):
Already merged, but for the record: :+1:

Issue #718, opened 2021-06-17
Fiddles cannot be launched via protocol from cli


codebytere:
#### Expected Behavior

Running 
```console
yarn start electron-fiddle://electron/main/docs/fiddles/screen/fit-screen
```

Should trigger a fiddle launch from documentation in Fiddle

#### Actual Behavior

Running 
```console
yarn start electron-fiddle://electron/main/docs/fiddles/screen/fit-screen
```

throws an error and Fiddle exits:

```console
fiddle on git:master ❯ yarn start electron-fiddle://electron/main/docs/fiddles/screen/fit-screen
yarn run v1.22.4
$ rimraf ./dist && electron-forge start -- electron-fiddle://electron/main/docs/fiddles/screen/fit-screen
✔ Checking your system
✔ Locating Application
Not in CI, skipping notarization
✔ Preparing native dependencies
✔ Contributors file on disk and recent.
✨  Built in 6.61s.

dist/app.15fac660.js.map           530.14 KB     66ms
dist/root.f39bf48d.css.map         436.97 KB     37ms
dist/root.f39bf48d.css                373 KB    3.91s
dist/app.15fac660.js               311.64 KB    5.72s
dist/src/main/main.js.map           94.05 KB     31ms
dist/src/main/main.js               59.64 KB    4.76s
dist/src/preload/preload.js.map     14.63 KB      6ms
dist/src/preload/preload.js         12.81 KB    2.84s
dist/static/index.html                 682 B    716ms
✔ Launching Application
Found protocol arg in argv: electron-fiddle://electron/main/docs/fiddles/screen/fit-screen
Creating main window
error: unknown command 'electron-fiddle://electron/main/docs/fiddles/screen/fit-screen'. See '. --help'.
CommanderError: error: unknown command 'electron-fiddle://electron/main/docs/fiddles/screen/fit-screen'. See '. --help'.
    at Command._exit (/Users/codebytere/Developer/fiddle/node_modules/commander/index.js:888:26)
    at Command._displayError (/Users/codebytere/Developer/fiddle/node_modules/commander/index.js:1729:10)
    at Command.unknownCommand (/Users/codebytere/Developer/fiddle/node_modules/commander/index.js:1812:10)
    at Command._parseCommand (/Users/codebytere/Developer/fiddle/node_modules/commander/index.js:1515:16)
    at Command.parse (/Users/codebytere/Developer/fiddle/node_modules/commander/index.js:1292:10)
[...]

PR #719, opened 2021-06-17
fix: loading fiddle protocol from cli

codebytere:
Closes https://github.com/electron/fiddle/issues/718.


ckerr (approved):
LGTM.

Optional, but this refactor is right on top of a pre-existing code smell of the `startsWith(${PROTOCOL}://)` test being repeated. If you extract the string test outside of the array like this:

```
const isProtocolString = (arg: string) => arg.startsWith(`${PROTOCOL}://`);

export const findProtocolArg = (argv: string[]) =>
  argv.find((arg) => isProtocolString(arg));
```

Then that other code repetition below could be rewritten as `if (isProtocolString(url))`

PR #720, opened 2021-06-17
feat: try downloading version if missing

codebytere:
Should ideally land after https://github.com/electron/fiddle/pull/719.

This PR makes it such that if a user tries to launch a fiddle via https://fiddle.electronjs.org/ and the package.json contains an Electron version the user doesn't have downloaded, Fiddle will ask if the user would like to download it before proceeding instead of just failing and exiting.

Tested by running 

```sh
yarn start electron-fiddle://electron/main/docs/fiddles/screen/fit-screen
```

### DIFF
--- a/src/renderer/binary.ts
+++ b/src/renderer/binary.ts
@@ -85,8 +85,8 @@ async function downloadBinary(ver: RunnableVersion): Promise<void> {
     // Ensure the target path is empty
     await fs.emptyDir(extractPath);
 
-    const electronFiles = await unzip(zipPath, extractPath);
-    console.log(`Binary: Unzipped ${version}`, electronFiles);
+    await unzip(zipPath, extractPath);
+    console.log(`Binary: Unzipped ${version}`);
     ver.state = VersionState.ready;
   } catch (error) {
     console.warn(`Binary: Failure while unzipping ${version}`, error);
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -4,6 +4,8 @@ import {
   ElectronReleaseChannel,
   GenericDialogType,
   PACKAGE_NAME,
+  VersionSource,
+  VersionState,
 } from '../interfaces';
 import { getOctokit } from '../utils/octokit';
 import { ELECTRON_ORG, ELECTRON_REPO } from './constants';
@@ -135,11 +137,27 @@ export class RemoteLoader {
   public async setElectronVersionWithRef(ref: string): Promise<boolean> {
     const version = await this.getPackageVersionFromRef(ref);
 
+    let downloadMissing = false;
     if (!this.appState.hasVersion(version)) {
-      this.handleLoadingFailed(
-        new Error('Version of Electron in example not supported'),
-      );
-      return false;
+      downloadMissing = await this.askToDownloadMissingVersion(version);
+      if (downloadMissing) {
+        const versionToDownload = {
+          source: VersionSource.remote,
+          state: VersionState.unknown,
+          version,
+        };
+
+        try {
+          this.appState.addNewVersions([versionToDownload]);
+          await this.appState.downloadVersion(versionToDownload);
+        } catch {
+          this.appState.removeVersion(versionToDownload);
+          this.handleLoadingFailed(
+            new Error(`Failed to download missing Electron version ${version}`),
+          );
+          return false;
+        }
+      }
     }
 
     // check if version is part of release channel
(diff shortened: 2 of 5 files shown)


src/renderer/binary.ts
```diff
@@ -85,8 +85,8 @@ async function downloadBinary(ver: RunnableVersion): Promise<void> {
     // Ensure the target path is empty
     await fs.emptyDir(extractPath);
 
-    const electronFiles = await unzip(zipPath, extractPath);
-    console.log(`Binary: Unzipped ${version}`, electronFiles);
+    await unzip(zipPath, extractPath);
```
codebytere, line 88: orthogonal change: `unzip` returns `Promise<void>` so this would always output

```sh
Binary: Unzipped v1.2.3 undefined
```

to console

tests/renderer/remote-loader-spec.ts
```diff
@@ -332,15 +352,6 @@ describe('RemoteLoader', () => {
       expect(instance.fetchGistAndLoad).toHaveBeenCalledWith<any>('gist');
     });
 
-    it('loads the example with confirmation', async () => {
```
codebytere, line 335: deleted duplicate test

src/renderer/remote-loader.ts
```diff
@@ -215,6 +235,25 @@ export class RemoteLoader {
...
   }
 
+  /**
+   * Asks the user if they would like to try to download a missing version
+   * of Electron.
+   *
+   * @param {string} missingVersion The version missing from the user's downloaded versions.
+   */
+  public async askToDownloadMissingVersion(
+    missingVersion: string,
```
ckerr, line 245: Doesn't look like this function gets reached in any of the tests. There's some pre-existing code in tests/renderer/state-spec.ts's `registerDialogHandler()` that can be copied into this spec to fix that gap

src/renderer/remote-loader.ts
```diff
@@ -215,6 +235,25 @@ export class RemoteLoader {
...
+  ): Promise<boolean> {
+    this.appState.setGenericDialogOptions({
+      type: GenericDialogType.confirm,
+      label: `It doesn't look like you've previously downloaded ${missingVersion} - Would you like to download it now?`,
+    });
+    this.appState.isGenericDialogShowing = true;
+    await when(() => !this.appState.isGenericDialogShowing);
+
+    return !!this.appState.genericDialogLastResult;
```
ckerr, line 254:
```suggestion
    return this.appState.runConfirmationDialog({
      label: `It doesn't look like you've previously downloaded ${missingVersion} - Would you like to download it now?`,
      ok: 'Download',
      type: GenericDialogType.confirm,
    });
```

tests/renderer/remote-loader-spec.ts
```diff
@@ -238,24 +239,43 @@ describe('RemoteLoader', () => {
...
+        {
+          source: VersionSource.remote,
+          state: VersionState.unknown,
+          version,
+        },
+      ]);
+      expect(store.setVersion).toBeCalledWith(version);
+    });
+
+    it('does not load unsupported versions of Electron', async () => {
```
ckerr, line 263: Could these two tests be renamed to better reflect what they're testing?

I think I'm reading this right that it's not checking to see if $version is supported, but whether or not the result of `askToDownload` is honored

src/renderer/remote-loader.ts
```diff
@@ -135,11 +137,29 @@ export class RemoteLoader {
   public async setElectronVersionWithRef(ref: string): Promise<boolean> {
     const version = await this.getPackageVersionFromRef(ref);
 
+    let downloadMissing = false;
     if (!this.appState.hasVersion(version)) {
-      this.handleLoadingFailed(
-        new Error('Version of Electron in example not supported'),
-      );
-      return false;
+      downloadMissing = await this.askToDownloadMissingVersion(version);
```
ckerr, line 142: ... does it make sense to prompt the user before downloading a version of Electron? We download them pretty freely everywhere else in the app, so I'm wondering why the app should handle it differently here.

ckerr:
> This PR makes it such that if a user tries to launch a fiddle via https://fiddle.electronjs.org/ and the package.json contains an Electron version the user doesn't have downloaded, Fiddle will ask if the user would like to download it before proceeding instead of just failing and exiting.

I love this idea.

src/renderer/remote-loader.ts
codebytere: That's a good point and i'm also generally happier to take the path of least resistance especially when it's already being taken elsewhere!

src/renderer/remote-loader.ts
```diff
@@ -136,10 +138,22 @@ export class RemoteLoader {
...
+        source: VersionSource.remote,
+        state: VersionState.unknown,
+        version,
+      };
+
+      try {
+        this.appState.addNewVersions([versionToDownload]);
+        await this.appState.downloadVersion(versionToDownload);
+      } catch {
+        this.appState.removeVersion(versionToDownload);
```
erickzhao, line 151: Should we `await` here?

codebytere: oops, yeah :)

erickzhao (approved):
LGTM!

PR #722, opened 2021-06-18
fix: title bar alightment on macOS

codebytere:
Before:

Screen Shot 2021-06-18 at 9 32 31 AM

After:

Screen Shot 2021-06-18 at 9 33 02 AM

Not totally sure if this is Big Sur only - if anyone else has thoughts please share!


ckerr:
LGTM. Anyone here _not_ on Big Sur who can confirm this looks alright on other versions?

erickzhao (commented):
Catalina comparison:

Live:
Live

Change:
Change

codebytere:
@erickzhao @bnb fixed this up a bit - the change only applied on Catalina or later now.

ckerr (approved):

Issue #723, opened 2021-06-18
If a local build is removed,  "Checking Status" spins indefinitely


codebytere:
As in title.

We should likely be checking local builds to ensure the file is still there, otherwise maybe pop up a dialog and say it can't be found and remove it.

PR #727, opened 2021-06-19
fix: show warning dialog for missing local versions

codebytere:
Closes https://github.com/electron/fiddle/issues/723.

if a local build is now selected but it's missing, Fiddle will now show a warning dialog indicating that the version can't be selected and that the user should verify it.

Screen Shot 2021-06-19 at 10 26 11 AM

### DIFF
--- a/src/renderer/components/commands-runner.tsx
+++ b/src/renderer/components/commands-runner.tsx
@@ -1,4 +1,4 @@
-import { Button, IButtonProps, Spinner } from '@blueprintjs/core';
+import { Button, ButtonProps, Spinner } from '@blueprintjs/core';
 import { observer } from 'mobx-react';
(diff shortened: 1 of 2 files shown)


src/renderer/components/commands-runner.tsx
```diff
@@ -26,7 +26,7 @@ export class Runner extends React.Component<RunnerProps> {
     } = this.props.appState;
 
     const state = currentElectronVersion && currentElectronVersion.state;
-    const props: IButtonProps = { className: 'button-run', disabled: true };
+    const props: ButtonProps = { className: 'button-run', disabled: true };
```
codebytere, line 29: saw a deprecated warning here and updated it.

ckerr (approved):
LGTM

PR #728, opened 2021-06-19
refactor: preload monaco

ckerr:
The way we load Monaco right now adds a handful of warts to the codebase:

* Two different pieces of the code (`components/outputs.tsx` and `components/editors.tsx`) try to load Monaco and initialize `app.monaco`, whichever one gets there first
* Loading Monaco in these components also couples them to the theme module, since it has code that needs to be invoked when Monaco is loaded
* Outputs and Editors have logic branches in their render code that must test whether Monaco is loaded
* fetch-types has extra logic branches to test whether Monaco is loaded and defer work until it is loaded.
* as a follow-up to the previous, there is also a `utils` helper whose only use is to help wait for Monaco to be loaded.

This PR's main goal is to preload monaco and make it a property of ElectronFiddle so that it is always available and accessible by the rest of the code. That fixes all the warts listed above!

It also introduces a new file, `src/renderer/main.tsx` that handles the bootstrapping of instantiating an App() that was previously done at the bottom of `src/renderer/app.tsx`. This is to ensure that an `App` is no longer automatically instantiated as a side-effect of importing from 'app', because that was problematic in ensuring the tests were in sync with the Monaco refactor.

Lastly, there is a related-but-slightly-tangential cleanup in preload + ambient.d.ts to remove a couple of unused window.ElectronFiddle properties, `editors` and `contentChangeListeners`. These were already unused; the only change here is to remove them.


src/renderer/components/editors.tsx
```diff
@@ -266,32 +261,6 @@ export class Editors extends React.Component<EditorsProps, EditorsState> {
     this.props.appState.editorMosaic.mosaicArrangement = currentNode;
   }
 
-  /**
-   * Loads monaco. If it's already loaded, it'll just set it on the current state.
-   * We're doing things a bit roundabout to ensure that we're not overloading the
```
codebytere, line 271: is there potential this happens with this refactor?

ckerr: answered in 1:1

codebytere (approved):

PR #731, opened 2021-06-21
fix: add dialog box for when user adds duplicate local electron builds

georgexu99:
Co-authored-by: VerteDinde <vertedinde@electronjs.org>

Users can be confused when adding a local electron build using the same folder path as a previously added (but not removed) build.

current behaviour checks that the path has been added before and ignores it, without feedback to the user.

This fix aims to resolve any user confusion as to why their local electron build "did not upload"

- [x] added check for existing local version
- [x] added snapshot test to ensure dialog box shows when expected

image

### DIFF
--- a/src/renderer/components/dialog-add-version.tsx
+++ b/src/renderer/components/dialog-add-version.tsx
@@ -17,7 +17,7 @@ import { getElectronNameForPlatform } from '../../utils/electron-name';
 import { getIsDownloaded } from '../binary';
 import { ipcRendererManager } from '../ipc';
 import { AppState } from '../state';
-import { hasExistingLocalVersion } from '../versions';
+import { getExistingLocalVersion } from '../versions';
 
 interface AddVersionDialogProps {
   appState: AppState;
@@ -74,7 +74,7 @@ export class AddVersionDialog extends React.Component<
    */
   public setFolderPath(folderPath: string) {
     const isValidElectron = getIsDownloaded('custom', folderPath);
-    const existingLocalVersion = hasExistingLocalVersion(folderPath);
+    const existingLocalVersion = getExistingLocalVersion(folderPath);
 
     this.setState({ existingLocalVersion, folderPath, isValidElectron });
   }
@@ -112,17 +112,18 @@ export class AddVersionDialog extends React.Component<
       name,
     };
 
-    // remove old local electron version if the user adds a new one with the same path
-    const shouldRemovePreviousVersion = isValidElectron && existingLocalVersion;
-    if (shouldRemovePreviousVersion) {
-      // get previous version
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
+    if (shouldSetToPreviousVersion) {
+      // set previous version as active version
       for (const ver of Object.values(versions)) {
         if (ver.localPath === existingLocalVersion?.localPath) {
-          this.props.appState.removeVersion(ver);
+          this.props.appState.setVersion(ver.version);
         }
       }
+    } else {
+      this.props.appState.addLocalVersion(toAdd);
     }
-    this.props.appState.addLocalVersion(toAdd);
     this.onClose();
   }
 
@@ -135,18 +136,22 @@ export class AddVersionDialog extends React.Component<
   }
 
   get buttons() {
-    const { isValidElectron, existingLocalVersion } = this.state;
-    const canSubmit = this.state.isValidElectron && this.state.isValidVersion;
-    // remove old local electron version if the user adds a new one with the same path
-    const shouldRemovePreviousVersion = isValidElectron && existingLocalVersion;
+    const {
+      isValidElectron,
+      isValidVersion,
+      existingLocalVersion,
+    } = this.state;
+    const canSubmit = isValidElectron && isValidVersion;
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
 
     return [
       <Button
         icon="add"
         key="submit"
-        disabled={shouldRemovePreviousVersion ? false : !canSubmit}
+        disabled={shouldSetToPreviousVersion ? false : !canSubmit}
         onClick={this.onSubmit}
-        text={shouldRemovePreviousVersion ? 'OK' : 'Add'}
+        text={shouldSetToPreviousVersion ? 'OK' : 'Add'}
       />,
       <Button icon="cross" key="cancel" onClick={this.onClose} text="Cancel" />,
     ];
@@ -190,44 +195,38 @@ export class AddVersionDialog extends React.Component<
   }
 
   private renderPath(): JSX.Element | null {
-    const { folderPath } = this.state;
+    const { isValidElectron, folderPath, existingLocalVersion } = this.state;
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
 
     if (!folderPath) return null;
     return (
       <Callout>
-        {this.renderDialog()}
-        {this.renderVersionInput()}
+        {this.getVersionFromPathDialog()}
+        {!shouldSetToPreviousVersion && this.renderVersionInput()}
       </Callout>
     );
   }
 
-  private renderDialog(): string {
+  private getVersionFromPathDialog(): string {
     const { existingLocalVersion, isValidElectron } = this.state;
 
     if (isValidElectron && existingLocalVersion) {
-      return `This folder path has already been added as version "${existingLocalVersion.version}". Would you like to set this path as a new version? \n \n`;
+      return `This folder path has already been added as version "${existingLocalVersion.version}".
+      Would you like to set your Electron version to "${existingLocalVersion.version}"? \n \n`;
     } else if (isValidElectron) {
       return `We found an ${getElectronNameForPlatform()} in this folder.`;
+    } else {
+      return `We did not find a ${getElectronNameForPlatform()} in this folder...`;
     }
-
-    return `We did not find a ${getElectronNameForPlatform()} in this folder...`;
   }
 
   private renderVersionInput(): JSX.Element | null {
-    const {
-      existingLocalVersion,
-      isValidElectron,
-      isValidVersion,
-      version,
-    } = this.state;
-    const shouldRemovePreviousVersion = isValidElectron && existingLocalVersion;
-
+    const { isValidElectron, isValidVersion, version } = this.state;
     if (!isValidElectron) return null;
 
     return (
       <>
         <p>
-          {shouldRemovePreviousVersion && <br />}
           Please specify a version, used for typings and the name. Must be{' '}
           <code>semver</code> compliant.
         </p>
--- a/src/renderer/components/dialog-add-version.tsx
+++ b/src/renderer/components/dialog-add-version.tsx
@@ -17,7 +17,7 @@ import { getElectronNameForPlatform } from '../../utils/electron-name';
 import { getIsDownloaded } from '../binary';
 import { ipcRendererManager } from '../ipc';
 import { AppState } from '../state';
-import { getExistingLocalVersion } from '../versions';
+import { getLocalVersionForPath } from '../versions';
 
 interface AddVersionDialogProps {
   appState: AppState;
@@ -74,7 +74,7 @@ export class AddVersionDialog extends React.Component<
    */
   public setFolderPath(folderPath: string) {
     const isValidElectron = getIsDownloaded('custom', folderPath);
-    const existingLocalVersion = getExistingLocalVersion(folderPath);
+    const existingLocalVersion = getLocalVersionForPath(folderPath);
 
     this.setState({ existingLocalVersion, folderPath, isValidElectron });
   }
@@ -101,7 +101,7 @@ export class AddVersionDialog extends React.Component<
       isValidElectron,
       existingLocalVersion,
     } = this.state;
-    const { versions } = this.props.appState;
+
     if (!folderPath) return;
 
     const name = folderPath.slice(-20).split(path.sep).slice(1).join(path.sep);
@@ -113,14 +113,9 @@ export class AddVersionDialog extends React.Component<
     };
 
     // swap to old local electron version if the user adds a new one with the same path
-    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
-    if (shouldSetToPreviousVersion) {
+    if (isValidElectron && existingLocalVersion?.localPath) {
       // set previous version as active version
-      for (const ver of Object.values(versions)) {
-        if (ver.localPath === existingLocalVersion?.localPath) {
-          this.props.appState.setVersion(ver.version);
-        }
-      }
+      this.props.appState.setVersion(existingLocalVersion.localPath);
     } else {
       this.props.appState.addLocalVersion(toAdd);
     }
@@ -149,7 +144,7 @@ export class AddVersionDialog extends React.Component<
       <Button
         icon="add"
         key="submit"
-        disabled={shouldSetToPreviousVersion ? false : !canSubmit}
+        disabled={!shouldSetToPreviousVersion && !canSubmit}
         onClick={this.onSubmit}
         text={shouldSetToPreviousVersion ? 'OK' : 'Add'}
       />,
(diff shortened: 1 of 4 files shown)


src/renderer/components/dialog-add-version.tsx
```diff
@@ -168,30 +190,44 @@ export class AddVersionDialog extends React.Component<
...
+
   private renderVersionInput(): JSX.Element | null {
-    const { isValidElectron, isValidVersion, version } = this.state;
+    const {
+      existingLocalVersion,
+      isValidElectron,
+      isValidVersion,
+      version,
+    } = this.state;
+    const shouldRemovePreviousVersion = isValidElectron && existingLocalVersion;
```
erickzhao, line 223: I don't think checking for `isValidElectron` is necessary here because if this variable is false, we're already returning `null` right below.

erickzhao (changes requested):
Seems like the renaming is a no-op for me when the local build is the active version. 🤔 

It might be easier to just select the existing version rather than performing the rename operation as well.

src/renderer/versions.ts
```diff
@@ -179,6 +179,26 @@ export function addLocalVersion(input: Version): Array<Version> {
...
+  folderPath: string,
+): Version | undefined {
+  const versions = getLocalVersions();
+  let duplicate;
+
+  versions.find((v) => {
+    if (v.localPath === folderPath) duplicate = v;
+  });
+
+  return duplicate ? (duplicate as Version) : undefined;
```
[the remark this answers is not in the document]
georgexu99: oh! Embarassing :P Thanks Erick

src/renderer/versions.ts
```diff
@@ -179,6 +179,22 @@ export function addLocalVersion(input: Version): Array<Version> {
...
+ * @param {Version} input
+ * @returns {Version | undefined}
+ */
+export function getExistingLocalVersion(
+  folderPath: string,
+): Version | undefined {
+  const versions = getLocalVersions();
+  const duplicate = versions.find((v) => v.localPath === folderPath);
+
+  return duplicate ? (duplicate as Version) : undefined;
```
ckerr, line 195: Array.prototype.find() returns the item or undefined, so I think we should be able to return it as-is without having to check the results:

```suggestion
  return getLocalVersions().find((v) => v.localPath === folderPath);
```

src/renderer/components/dialog-add-version.tsx
```diff
@@ -168,25 +195,33 @@ export class AddVersionDialog extends React.Component<
...
       <Callout>
-        {info}
-        {this.renderVersionInput()}
+        {this.getVersionFromPathDialog()}
+        {!shouldSetToPreviousVersion && this.renderVersionInput()}
       </Callout>
     );
   }
 
+  private getVersionFromPathDialog(): string {
```
ckerr, line 210: This function name is slightly misleading; it implies a version will be returned but it's a message.

Maybe something like `getConfirmationMessage()` or `buildDialogText()`?

src/renderer/versions.ts
```diff
@@ -179,6 +179,22 @@ export function addLocalVersion(input: Version): Array<Version> {
...
 }
 
+/**
+ * Check to see if local version has already been added
+ * for the current folder path
+ *
+ * @param {Version} input
+ * @returns {Version | undefined}
+ */
+export function getExistingLocalVersion(
```
ckerr, line 189: Another naming suggestion for consistency with other utils like `getElectronNameForPlatform()`, `getLocalTypePathForVersion()`, etc

```suggestion
export function getLocalVersionForPath(
```

src/renderer/versions.ts
```diff
@@ -179,6 +179,22 @@ export function addLocalVersion(input: Version): Array<Version> {
   return versions;
 }
 
+/**
+ * Check to see if local version has already been added
+ * for the current folder path
+ *
+ * @param {Version} input
```
ckerr, line 186:
```suggestion
 * @param {string} folderPath
```

src/renderer/versions.ts
```diff
@@ -179,6 +179,22 @@ export function addLocalVersion(input: Version): Array<Version> {
   return versions;
 }
 
+/**
+ * Check to see if local version has already been added
+ * for the current folder path
```
ckerr, line 184:
```suggestion
 * Get the Version (if any) that is located at localPath.
```

src/renderer/components/dialog-add-version.tsx
```diff
@@ -103,7 +112,18 @@ export class AddVersionDialog extends React.Component<
...
-    this.props.appState.addLocalVersion(toAdd);
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
+    if (shouldSetToPreviousVersion) {
+      // set previous version as active version
+      for (const ver of Object.values(versions)) {
+        if (ver.localPath === existingLocalVersion?.localPath) {
+          this.props.appState.setVersion(ver.version);
+        }
+      }
```
ckerr, line 123: Could this for loop be replaced with `this.props.appState.setVersion(existingLocalVersion.localPath)`?

It looks like there's something going on here that makes walking the `appState.versions` array necessary, but I'm not seeing why that's needed instead of using `existingLocalVersion.localPath` :thinking:

src/renderer/components/dialog-add-version.tsx
```diff
@@ -103,7 +112,18 @@ export class AddVersionDialog extends React.Component<
       name,
     };
 
-    this.props.appState.addLocalVersion(toAdd);
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
+    if (shouldSetToPreviousVersion) {
+      // set previous version as active version
+      for (const ver of Object.values(versions)) {
+        if (ver.localPath === existingLocalVersion?.localPath) {
```
ckerr, line 120: `?` is not needed; we know from line 116 that `existingLocalVersion` is truthy

src/renderer/components/dialog-add-version.tsx
```diff
@@ -116,15 +136,22 @@ export class AddVersionDialog extends React.Component<
...
+    const canSubmit = isValidElectron && isValidVersion;
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
 
     return [
       <Button
         icon="add"
         key="submit"
-        disabled={!canSubmit}
+        disabled={shouldSetToPreviousVersion ? false : !canSubmit}
```
ckerr, line 152:
```suggestion
        disabled={!shouldSetToPreviousVersion && !canSubmit}
```

src/renderer/components/dialog-add-version.tsx
georgexu99: Oops! You are totally right- an oversight for when line 121 used to delete the previous version in a previous commit that I ended up changing 😆

src/renderer/components/dialog-add-version.tsx
```diff
@@ -168,25 +190,33 @@ export class AddVersionDialog extends React.Component<
...
       </Callout>
     );
   }
 
+  private buildDialogText(): string {
+    const { existingLocalVersion, isValidElectron } = this.state;
+
+    if (isValidElectron && existingLocalVersion) {
+      return `This folder path has already been added as version "${existingLocalVersion.version}".
+      Would you like to set your Electron version to "${existingLocalVersion.version}"? \n \n`;
```
ckerr, line 210:
```suggestion
      return `This folder is already in use as version "${existingLocalVersion.version}". Would you like to switch to that version now?`;
```

src/renderer/components/dialog-add-version.tsx
```diff
@@ -116,15 +131,22 @@ export class AddVersionDialog extends React.Component<
...
   get buttons() {
-    const canSubmit = this.state.isValidElectron && this.state.isValidVersion;
+    const {
+      isValidElectron,
+      isValidVersion,
+      existingLocalVersion,
+    } = this.state;
+    const canSubmit = isValidElectron && isValidVersion;
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
```
ckerr, line 141: I know I had a run at reviewing this section before with the `disabled={foo}` line but this whole section is still hard for me to scan, so I'm going to try meddling again. :slightly_smiling_face: 

- I think part of it is that we can still submit even when `canSubmit` is false since this dialog now has a dual role.
- Another part is that `shouldSetToPreviousVersion` is a little unclear, we could rename it to draw a line between it and `existingLocalVersion` more clearly.
- Thirdly, the `disabled={foo}` logic, even revised, is still kind of hard to scan.

```suggestion
    const canAdd = isValidElectron && isValidVersion && !existingLocalVersion;
    const canSwitch = isValidElectron && existingLocalVersion;
```

To me this is a lot easier to read, e.g. `canAdd` and `canSwitch` are terse and their relationship to each other is easy to understand at-a-glance. And this also makes the `disabled={}` line clearer: `disabled={!canAdd && !canSwitch}`

src/renderer/components/dialog-add-version.tsx
```diff
@@ -116,15 +131,22 @@ export class AddVersionDialog extends React.Component<
...
 
     return [
       <Button
         icon="add"
         key="submit"
-        disabled={!canSubmit}
+        disabled={!shouldSetToPreviousVersion && !canSubmit}
```
ckerr, line 147:
```suggestion
        disabled={!canAdd && !canSwitch}
```

src/renderer/components/dialog-add-version.tsx
```diff
@@ -116,15 +131,22 @@ export class AddVersionDialog extends React.Component<
...
         onClick={this.onSubmit}
-        text="Add"
+        text={shouldSetToPreviousVersion ? 'OK' : 'Add'}
```
ckerr, line 149:
```suggestion
        text={canSwitch ? 'Switch' : 'Add'}
```

ckerr (commented):
I had a couple more suggestions but nothing structural. Overall, this is looking good!

src/renderer/components/dialog-add-version.tsx
```diff
@@ -168,25 +190,33 @@ export class AddVersionDialog extends React.Component<
   }
 
   private renderPath(): JSX.Element | null {
-    const { folderPath, isValidElectron } = this.state;
+    const { isValidElectron, folderPath, existingLocalVersion } = this.state;
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
```
ckerr, line 194:
```suggestion
    const canSwitch = isValidElectron && existingLocalVersion;
```

For consistency with other suggested change

src/renderer/components/dialog-add-version.tsx
```diff
@@ -168,25 +190,33 @@ export class AddVersionDialog extends React.Component<
...
+        {this.buildDialogText()}
+        {!shouldSetToPreviousVersion && this.renderVersionInput()}
       </Callout>
     );
   }
 
+  private buildDialogText(): string {
+    const { existingLocalVersion, isValidElectron } = this.state;
+
+    if (isValidElectron && existingLocalVersion) {
```
ckerr, line 208: Since we've got the same of (isValidElectron && existingLocalVesion) logic repeated here again, I'd probably use the same variables here too for clarity's sake, e.g.

```
const canAdd = isValidElectron && isValidVersion && !existingLocalVersion;
const canSwitch = isValidElectron && existingLocalVersion;

if (canSwitch) {
  return ...
}
if (canAdd) {
  return ...
}
return ...
```

georgexu99: I apologize if I'm mistaken, but I took a look at this with @VerteDinde for a bit on Friday and to my understanding, we were unable to make variable names work like you described here due to some typescript issues like so: 

image

ckerr: @georgexu99 this is more a nice-to-have and not a blocker.

However, in that screenshot I think at line 211 we know that `existingLocalVersion` is defined because of the test on line 209. So we just have to explain that to TypeScript by appending a `!` to the variable name e.g.

```diff
- ${existingLocalVersion.version}
+ ${existingLocalVersion!.version}
```

src/renderer/components/dialog-add-version.tsx
```diff
@@ -116,15 +131,23 @@ export class AddVersionDialog extends React.Component<
...
 
     return [
       <Button
         icon="add"
         key="submit"
-        disabled={!canSubmit}
+        disabled={!canAdd && !canSwitch}
         onClick={this.onSubmit}
-        text="Add"
+        text={shouldSetToPreviousVersion ? 'Switch' : 'Add'}
```
ckerr, line 150:
```suggestion
        text={canSwitch ? 'Switch' : 'Add'}
```

src/renderer/components/dialog-add-version.tsx
```diff
@@ -116,15 +131,23 @@ export class AddVersionDialog extends React.Component<
...
-    const canSubmit = this.state.isValidElectron && this.state.isValidVersion;
+    const {
+      isValidElectron,
+      isValidVersion,
+      existingLocalVersion,
+    } = this.state;
+    const canAdd = isValidElectron && isValidVersion && !existingLocalVersion;
+    const canSwitch = isValidElectron && existingLocalVersion;
+    // swap to old local electron version if the user adds a new one with the same path
+    const shouldSetToPreviousVersion = isValidElectron && existingLocalVersion;
```
ckerr, line 142: `shouldSetToPreviousVersion` is a little prolix and its relationship to the other fields isn't clear at-a-glance. In particular someone who hasn't read this PR might not be able to tell if 'previous version' is the same thing as 'existing local version'.

It's also identical to the value of `canSwitch` so we probably don't need both.

There's real value in being able to understand code at-a-glance:

> `{text=canSwitch ? 'Switch' : 'Add'}`

Is something a developer can read and understand as soon as their eyes see it. It's short, it's direct, and it follows the hints the rest of the code has with use of the word 'switch', e.g. it's easy to understand the relationship between the two uses of 'switch' in that line.

> `shouldSetToPreviousVersion`

is slower to read because it takes longer to parse the words in the name and understand how they relate to the rest of the code.

```suggestion
```

ckerr (commented):
Thanks for the changes!

src/renderer/components/dialog-add-version.tsx
```diff
@@ -168,25 +191,32 @@ export class AddVersionDialog extends React.Component<
...
+    const { existingLocalVersion, isValidElectron } = this.state;
+
+    if (isValidElectron && existingLocalVersion) {
+      return `This folder is already in use as version "${existingLocalVersion.version}". Would you like to switch to that version now?`;
+    } else if (isValidElectron) {
+      return `We found an ${getElectronNameForPlatform()} in this folder.`;
+    } else {
+      return `We did not find a ${getElectronNameForPlatform()} in this folder...`;
+    }
+  }
```
ckerr, line 216:
```suggestion
  private buildDialogText(): string { 
    const { isValidElectron, isValidVersion, existingLocalVersion } = this.state;
    const canAdd = isValidElectron && isValidVersion && !existingLocalVersion;
    const canSwitch = isValidElectron && existingLocalVersion;

    if (canSwitch)
      return `This folder is already in use as version "${existingLocalVersion!.version}". Would you like to switch to that version now?`;

    if (canAdd)
      return `We found an ${getElectronNameForPlatform()} in this folder.`;

    return `We did not find a ${getElectronNameForPlatform()} in this folder...`;
  } 

```

src/renderer/components/dialog-add-version.tsx
georgexu99: oh Thank you!

src/renderer/components/dialog-add-version.tsx
georgexu99: Thanks for this feedback, I'll keep this in mind for the future <3

src/renderer/components/dialog-add-version.tsx
georgexu99: The second block on line 214, i think the conditional must stay as Felix's old implementation 'isValidElectron'. If we change it as 'canAdd' and add 'isValidVersion' and '!existingLocalVersion' , the logic changes and we don't get the text we expect in certain test cases

I will however change the if-else styling to remove the curly brackets (is this a style choice or is it objectively better?)

ckerr: Yes, you're right about the logic. Good catch.

No, the if-else braces are subjective and you can do whatever you like. Now that you mention it, the style I used in my suggested change is probably subjectively worse since it's not as consistent with the rest of the code around it, so probably better to keep the braces.

It's arguably better to not use 'else' after 'return' since it's an unnecessary code instruction that the compiler will have to parse and then discard. How much difference that overhead makes in the real world is probably negligible, but that's the closest I know of to an _objective_ answer of what would be the better way to write that function.

tldr; probably keep the braces in, and keep the elses out.

Anyway, thanks for your patience with all the variable & function naming. :slightly_smiling_face:

ckerr (approved):

MobarakeGH (changes requested):
#744

erickzhao (approved):

Issue #726, opened 2021-06-18
Enhancement: better sorting logic in the version selector


nornagon:
When I type "5" in the version selector box, the first suggestion is 15.0.0-nightly.\*, followed by 8.5.\*, followed by 5.*.

Ideally, the 5.* versions would come first.

erickzhao:
I think there's support for better filtering through the `itemPredicate` prop in https://blueprintjs.com/docs/#select/select-component. Will have to do more investigating!

PR #739, opened 2021-06-24
feat: improve version filtration logic

codebytere:
Closes https://github.com/electron/fiddle/issues/726.

Something to note here is that it does slow down rendering perf a bit - i'd love yall to poke at it a bit and see if it's something you'd be bothered by.

### DIFF
--- a/src/renderer/components/version-select.tsx
+++ b/src/renderer/components/version-select.tsx
@@ -1,5 +1,5 @@
-import { Button, IButtonGroupProps, MenuItem } from '@blueprintjs/core';
-import { ItemPredicate, ItemRenderer, Select } from '@blueprintjs/select';
+import { Button, ButtonGroupProps, MenuItem } from '@blueprintjs/core';
+import { ItemListPredicate, ItemRenderer, Select } from '@blueprintjs/select';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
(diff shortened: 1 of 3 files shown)


src/renderer/components/version-select.tsx
```diff
@@ -1,5 +1,5 @@
-import { Button, IButtonGroupProps, MenuItem } from '@blueprintjs/core';
```
codebytere, line 1: `IButtonGroupProps` is deprecated - noticed when looking through this file.

src/renderer/components/version-select.tsx
```diff
@@ -58,15 +58,32 @@ export function getItemIcon({ state }: RunnableVersion) {
...
+  const q = query.toLowerCase();
+
+  return versions
+    .filter(({ version }) => version.toLowerCase().includes(q))
+    .sort((a, b) => {
+      if (a.version.indexOf(q) > b.version.indexOf(q)) return 1;
+      if (b.version.indexOf(q) > a.version.indexOf(q)) return -1;
+
+      return 0;
+    });
```
ckerr, line 86: We can eliminate half of these searches

```suggestion
    .sort((a, b) => a.version.indexOf(q) - b.version.indexOf(q)));
```

src/renderer/components/version-select.tsx
ckerr: I'm having deja vu. Didn't we aleady remove this? :dizzy:

src/renderer/components/version-select.tsx
ckerr, line 86: Since you commented on perf -- we can bring the number of searches down from O(n^2) to O(n) with a temporary, and that lets us perform the sort on a lowercase set as well

```suggestion
  return versions
    .map((version: RunnableVersion) => ({
      index: version.version.toLowerCase().indexOf(q),
      version,
    }))
    .filter((item) => item.index !== -1)
    .sort((a, b) => a.index - b.index)
    .map((item) => item.version);
```

src/renderer/components/version-select.tsx
codebytere: we removed another instance of `IButtonProps` i think :D

src/renderer/components/version-select.tsx
codebytere: ooh i didn't consider the temporary! good shout, ty!

ckerr (approved):

PR #740, opened 2021-06-24
refactor: make EditorMosaic implementation details private

ckerr:
OK we're in the home stretch of the [EditorMosaic refactor saga](https://github.com/electron/fiddle/pull/664)! This PR is the last stop before the next & final PR gets to the goal of merging custom editors and default editors together. :tada:

This PR looks big but it's actually simple because all the changes do the same thing: **If code needs something from EditorMosaic, it should be doable with one or two simple API calls.** Don't force client code to look at the class' internals and guess an answer. Make state private whenever possible and make the public API as simple as possible. For example:

1. Add a new computed public property `EditorMosaic.numVisible` so that `commands-editors.tsx` and `editors.tsx` don't have to look at `typeof editorMosaic.mosaicArrangement === 'string'` anymore to figure out if there's only one editor showing

2. Add a new function `EditorMosaic.addNew(id)` that throws on unsupported/duplicate IDs so that `commands-editors.tsx` doesn't have to walk through the mosaic's contents to look for duplicates

3. Add a new helper function `EditorMosaic.toggle(id)` to toggle an editor's visibility so that `commands-editors.tsx` doesn't have to look up a file's visibility when the 'toggle' command button is pressed

4. Add a new computed property `EditorMosaic.files` that is a Map of files to visibility state. This is intended as stable public API for tests & client code so that other EditorMosaic implementation details can be made private and refactored:
    - `EditorMosaic.editors`, which was only public for tests, can be removed from public API
    - `EditorMosaic.getVisibleMosaics()` can be removed from public API. It wasn't as useful as `EditorMosaic.files` since if a file was missing from the former, you had no way of knowing whether it was because the file was hidden or if it didn't exist.

5. `EditorMosaic.addEditor()` now handles populating the editor with the file's contents (and the MonacoEditor's ViewState data, if available). This gives us a slew of encapsulation wins:
    - `components/editor.tsx` no longer needs access to the backup information
    - `components/editor.tsx` no longer needs to care about monaco models
    - the `EditorBackup` interface in `editor-mosaic.ts` is no longer exported
    - `EditorBackup` now only has two states instead of three
    - `EditorMosaic.getEditorBackup()` can be removed from public API
    - `EditorMosaic.getEditorViewState()` can be removed from public API
    - `EditorMosaic.getEditorModel()` can be removed from public API
    - `EditorMosaic.getAndRemoveEditorValueBackup()` can be removed from public API
    - `EditorMosaic.closedPanels` property has been made private (and renamed `EditorMosaic.backups`)
    - `getContent()` was no longer used & has been removed

Other changes:

- **Lots** of new tests, especially in `editor-mosaic-spec.ts`

- AppState can now use `editorMosaic.isEdited` instead of `isContentUnchanged()`, so the latter has been removed.

- `DEFAULT_MOSAIC_ARRANGEMENT` and `SORTED_EDITORS` are no longer used and have been removed.

- Some of EditorMosaic's public API has been renamed for terseness / simplicity, e.g.
    - `EditorMosaic.resetEditorLayout()` -> `EditorMosaic.resetLayout()`,
    - `EditorMosaic.hideAndBackupMosaic()` -> `EditorMosaic.hide()`


src/renderer/editor-mosaic.ts
```diff
@@ -1,85 +1,68 @@
...
-  value?: string;
-  model?: MonacoType.editor.ITextModel | null;
+interface EditorBackup {
+  model: MonacoType.editor.ITextModel;
   viewState?: MonacoType.editor.ICodeEditorViewState | null;
 }
 
 export class EditorMosaic {
-  @observable public readonly editors: Map<EditorId, Editor> = new Map();
   @observable public customMosaics: EditorId[] = [];
```
erickzhao, line 36: Does it make sense to turn `customMosaics` into a Set given that we don't allow duplicate file names?

src/renderer/editor-mosaic.ts
```diff
@@ -150,173 +139,110 @@ export class EditorMosaic {
...
-  /**
-   * Shows the editor value for a given editor.
-   *
-   * @param {EditorId} id
-   */
-  @action public showMosaic(id: EditorId) {
-    this.setVisibleMosaics([...this.getVisibleMosaics(), id]);
+  /** Reset the mosaic's layout back to the default */
+  @action public resetLayout() {
+    this.set(this.values());
```
erickzhao, line 148: Optional: I'm wondering if I'm overthinking this, but I had trouble understanding the magic behind the `resetLayout()` function.

`this.set` does all the heavy lifting, but you have to go a couple layers deep in the abstraction to understand how exactly the default layout is set.

src/renderer/components/commands-action-button.tsx
```diff
@@ -420,9 +418,8 @@ export class GistActionButton extends React.Component<
       };
     }
 
-    // Add files for any custom editors created by the user.
-    for (const mosaic of customMosaics) {
-      filesList[mosaic] = { content: values[mosaic] };
+    for (const id of Object.keys(values)) {
+      filesList[id] = { content: values[id] };
     }
```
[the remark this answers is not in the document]
ckerr: You're right! It's not needed here. I left it alone for logistical reasons -- this PR is already pretty big and changing the `DEFAULT_EDITORS` code is more on-topic for the next PR which finally merges default & custom editors together.

src/renderer/editor-mosaic.ts
ckerr: Probably doesn't matter -- It's going away altogether in the next PR

As of this PR I think it's only read by tests. Another case of "would this change make more sense in `refactor/merge-custom-and-default-editors`

src/renderer/editor-mosaic.ts
ckerr: This is a good point. I'll try to make it more intuitive.

ckerr: I've tried to improve this readability in 203c6ae:

- rename addOrReplace() to addFile() for clarity
- add some comments for nonobvious parts esp. in set(), setVisible()
- inline-method refactor: move createModel() into addFile() as a minor readability tweak; one fewer function to have to follow the control-flow through when reading.
- reorganized the source to put related functions together, esp. putting resetLayout() before set() and moving show() to be right after addFile() so that one can begin reading at resetLayout(), the functions that follow in the file are the same as the call chain: resetLayout() -> set() -> addFile() -> show() -> setVisible()

erickzhao (approved):

src/renderer/editor-mosaic.ts
erickzhao: I think the comments make it a lot clearer!

PR #745, opened 2021-06-26
refactor: remove unused output.editorDidMount

ckerr:
Just a small piece of dead code removal.

`CommandsProps.editorDidMount` exists because the code was cloned from the sibling `EditorProps` class; however, since `editorDidMount` is never using for outputs, it's dead code there and can be removed.

_edit:_ Coveralls is just angry because tested code is being removed. :stuck_out_tongue:


erickzhao (approved):
LGTM! Ignoring the Coveralls temper tantrum 😂

PR #746, opened 2021-06-28
chore: remove dead code

ckerr:
Remove some code that's not being called anywhere:

- function invertPosition()
- function sortButtons()
- function getDocsUrlForModule()
- type AppState.Editor

Move to tests/ because it's used by tests but not in production:

- function waitFor()


erickzhao (approved):
LGTM!

PR #750, opened 2021-06-30
refactor: unify custom and default editors

ckerr:
The last PR in the [EditorMosaic refactor saga](https://github.com/electron/fiddle/pull/664)! Finally gets to the destination of unifying 'default editors' and 'custom editors' together.

## The Big Picture

- The distinction between 'custom editors' and 'default editors' has been removed. For example, adding an editor via the toolbar now says "Enter a name for your new file" instead of "Enter a filename for your custom editor".

## The Details

- The most minimal Fiddle now needs only one file, `main.js`. If you have an EditorValues and want to ensure this file is present, e.g. in readFiddle(), editor-utils' [`ensureRequiredFiles()`](https://github.com/electron/fiddle/blob/80df98d8aa5f8ed77701e2a787b14eab23294ec1/src/utils/editor-utils.ts#L68) will fill in a placeholder if needed.

- There is one use case left where the app needs to know if a file is one of the 'default editors' (e.g. `preload.js` and `renderer.js`): when deciding whether or not to show a confirm dialog when loading a remote gist. editor-utils' [`isKnownFile()`](https://github.com/electron/fiddle/blob/80df98d8aa5f8ed77701e2a787b14eab23294ec1/src/utils/editor-utils.ts#L14) has been kept for this use case. *(Do we actually need this use case?)*

- `DEFAULT_EDITORS` and `DefaultEditor` have been removed from `src/interfaces.ts`.

- `customMosaics` has been removed from EditorMosaic.

## The Diff

* The changes in `src/` are mostly straightforward and are a smaller part of the diff.

* There are more changes in `tests/`. There were a lot of uses of `DefaultEditorId` or `DEFAULT_EDITORS` that had to be updated. Most of these were procedural changes but they still increase the diff size.

* A few tests were rewritten because it no longer makes sense to test defaults & customs separately.

* I linted my code with `typescript-eslint/recommended-requiring-type-checking` and hit a few warnings. Some secondary changes were needed to fix these errors:
    - Most of these were about unsafe use of `any`. These were fixed by adding type information or using `unknown`.
    - A few were about ambiguous 'this' references when calling unbound class members. The [typescript docs](https://hamednourhani.gitbooks.io/typescript-book/content/docs/tips/bind.html) say "If you have a class member function that you expect to pass around, use an arrow function in the first place" so this has been done, e.g. `AppState.showConfirmDialog()` and `EditorDropdown.addNewFile()` are now arrow functions for this reason.

* EditorMosaicMock has been removed; the tests now use a live EditorMosaic instead. This was a really easy change. The Mock class wasn't adding any value and now the tests are using more production code.


ckerr:
Marking as a draft for now because we've got so many fixes queued up in the main branch that we should do a 0.25.3 release with those fixes _before_ this lands.

Note, **This code is still ready for review**. I don't plan on making any more changes except syncing to main & addressing code review comments.

src/main/files.ts
```diff
@@ -51,31 +50,23 @@ export async function showSaveDialog(event?: IpcEvents, as?: string) {
...
  * @param {string} filePath
  * @returns {Promise<boolean>}
  */
-async function ensureSaveTargetEmpty(filePath: string): Promise<boolean> {
-  const targetPaths = Object.values(DefaultEditorId).map((filename) =>
-    path.join(filePath, filename),
+async function isOkToSaveAt(filePath: string): Promise<boolean> {
+  return (
+    !fs.existsSync(filePath) ||
+    fs.readdirSync(filePath).filter(isSupportedFile).length === 0 ||
```
erickzhao, line 67: nit(non-blocking): If the function is `async`, is there value to using the synchronous `fs` APIs here? Is the performance cost negligible either way?

src/renderer/components/commands-action-button.tsx
```diff
@@ -409,9 +408,11 @@ export class GistActionButton extends React.Component<
   };
 
   private gistFilesList = (values: EditorValues) => {
-    const ids = [...DEFAULT_EDITORS, ...Object.keys(values)];
+    values = ensureRequiredFiles(values);
```
erickzhao, line 411: nit(non-blocking): any reason for re-assigning `values` here rather than creating a new `const` variable?

src/renderer/components/commands-editors.tsx
```diff
@@ -143,24 +140,22 @@ export class EditorDropdown extends React.Component<
...
 
-  public async removeCustomEditor(event: React.MouseEvent) {
+  public removeFile = (event: React.MouseEvent) => {
     const { id } = event.currentTarget;
     const { editorMosaic } = this.props.appState;
 
-    console.log(`EditorDropdown: Removing custom editor ${id}`);
-    editorMosaic.removeCustomMosaic(id as EditorId);
-  }
+    editorMosaic.hide(id as EditorId);
```
erickzhao, line 152: thought: the previous implementation of `removeCustomMosaic` also entirely removed the editor from the `customEditors` mosaic. Here, we're instead only hiding it. Is there a further action we can take to ensure that the editor is entirely erased?

src/renderer/components/commands-editors.tsx
```diff
@@ -143,24 +140,22 @@ export class EditorDropdown extends React.Component<
...
+  };
 
-  public onItemClick(event: React.MouseEvent) {
+  public onItemClick = (event: React.MouseEvent) => {
```
erickzhao, line 155: thought: maybe we should go ahead and refactor our old class functions into arrow functions to avoid the `bind(this)` boilerplate everywhere.

erickzhao (changes requested):

src/renderer/components/commands-editors.tsx
ckerr: Agreed. Arrow functions are an overall better way of doing the same thing.

src/main/files.ts
ckerr: I don't think there's any harm in using async here. Updated in 41260f3fa427794ab13c8fc1455e68f3f32e8932 .

src/renderer/components/commands-editors.tsx
ckerr: Previous implementation:

ckerr: Previous implementation of removeCustomMosaic():

```ts
  @action public removeCustomMosaic(id: EditorId) {
    this.hide(id);
    this.customMosaics = this.customMosaics.filter((mosaic) => mosaic !== id);
  }
```

But the thing is, `EditorMosaic.customMosaics` [isn't being used anywhere](https://github.com/electron/fiddle/search?q=customMosaics), so the function is equivalent to calling `hide()`. No further action is needed.

Slightly more info: EditorMosaic.hide() makes a backup of the monaco editor's state and then removes it from the mosaic property. That mosaic property is an observable and _that_ is what triggers react-mosaic-component to actually initiate creating / deleting monaco editors. EditorMosaic doesn't actually create or destroy the directly; it's all done indirectly via the mosaic component. This is the same approach that Fiddle's been using forever.

erickzhao: I see. This seems to be a wider problem with the original "remove" functionality because removing a custom editor just hides it but it remains in the `Editors` dropdown regardless.

I filed a separate issue describing my qualms with this UX: https://github.com/electron/fiddle/issues/863

erickzhao: @ckerr I think removing the Popover `Remove` menu altogether might be worth it because:

1) If the editor is visible, `editor.hide()` is the same as `editor.toggle()` and we can already toggle by clicking on the editor name directly.
2) If the editor is not visible, `editor.hide()` is a no-op.

ckerr: This is a good point. We need a way to actually remove a editor _and_ its backed up state if the user doesn't want that file anymore, and I don't think we have that right now.

src/renderer/components/commands-editors.tsx
ckerr: Added in af73e65b544f71d1ee398f7e35e3a83c552c0cbb. The 'Remove' menuitems now fully remove the file, not just hide it.

erickzhao (commented):
Digging the new `remove` function.

One bug I noticed is that removing an editor and updating the published gist URL will not remove the file from the gist.

ckerr:
> One bug I noticed is that removing an editor and updating the published gist URL will not remove the file from the gist.

This has been the case ever since the 'remove' menubutton was introduced. For example you can see the same behavior in 0.23.0. Is it a blocker for this PR?

The root cause is we update the gist [here](https://github.com/electron/fiddle/blob/master/src/renderer/components/commands-action-button.tsx#L179) with [octokit.rest.gists.update](https://octokit.github.io/rest.js/v18#gists-update) where "Files from the previous version of the gist that aren't explicitly changed during an edit are unchanged."

Since the files have been removed and Fiddle's internal modified EditorValues don't know what's been removed, we probably need to fetch the old copy, diff them, and explicitly remove the files that way. (Also, it's not completely clear how to remove a file from a gist using github's API...?

erickzhao (approved):
Change can be left for a separate PR. LGTM I think.

PR #760, opened 2021-07-11
chore: bump electron dep to 12.0.14

ckerr:
Startup time (as measured by reload) is faster than in 11.

For reasons I haven't tracked down yet, bumping to 13 causes the main window to flash white for a moment, so using 12 in this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,6 @@
   "description": "The easiest way to get started with Electron",
   "repository": "https://github.com/electron/fiddle",
   "main": "./dist/src/main/main",
-  "engines": {
-    "node": ">=14.16.0"
-  },
   "scripts": {
     "contributors": "node ./tools/contributors.js",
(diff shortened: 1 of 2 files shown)


erickzhao (approved):

package.json
```diff
@@ -93,7 +93,7 @@
     "@typescript-eslint/eslint-plugin": "^4.15.2",
     "@typescript-eslint/parser": "^4.15.2",
     "cross-fetch": "^3.1.0",
-    "electron": "11.3.0",
+    "electron": "^12.0.14",
```
erickzhao, line 96: Can we keep this as an exact dep?

erickzhao (approved):

ckerr:
Literally only changes package.json and yarn.lock

> coverage/coveralls — Coverage decreased (-0.002%) to 94.582%

ಠ_ಠ

PR #775, opened 2021-07-14
fix: editors sometimes disappear when loading new fiddle

### LINKED ISSUE #774: Editors sometimes disappear when loading new Fiddle
Screenshot:

image

Repro:

1. Load a Fiddle with empty files (e.g. https://gist.github.com/e27800f409182981352b68a90690c4a6)
2. Load the new default Fiddle (with <kbd>⌘</kbd> + <kbd>N</kbd>)

I bisected a bit and found out that #740 was where this bug regressed. I've also confirmed that #750 doesn't fix the issue.

ckerr:
Fixes #774 reported by @erickzhao.

When loading a new fiddle, if it has a filename that was in the previous fiddle too, EditorMosaic tries to recycle the MonacoEditor allocated to that file. Unfortunately, does this even if the new fiddle's file is going to be hidden. This corrupts the internal state.

In the repro case described in 744, the state was corrupted when _loading_ https://gist.github.com/e27800f409182981352b68a90690c4a6 because its preload & renderer files were hidden, causing the inconsistent internal state described above. The error that appears when loading another fiddle _after_ that gist is a symptom of that already-inconsistent state.

This PR fixes that bug and adds Yet Another Regression Test for this state too. :slightly_smiling_face:

### DIFF
--- a/src/renderer/editor-mosaic.ts
+++ b/src/renderer/editor-mosaic.ts
@@ -125,8 +125,12 @@ export class EditorMosaic {
       this.backups.set(id, backup);
     }
 
-    // if the file has nontrivial content, put it in the mosaic to show it.
-    if (value.length && value !== getEmptyContent(id)) this.show(id);
+    // only show the file if it has nontrivial content
+    if (value.length && value !== getEmptyContent(id)) {
+      this.show(id);
+    } else {
+      this.hide(id);
+    }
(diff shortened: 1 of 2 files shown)


erickzhao (approved):
LGTM! Great description of the problem. 👍

PR #777, opened 2021-07-14
fix: publishing gists with empty files to Github

### LINKED ISSUE #776: bug: "Publishing Fiddle to GitHub failed. Are you connected to the internet?"
Reported by both @nornagon and @miniak in #proj-fiddle. This is a 0.25.2 regression.

cannot-publish

ckerr:
Fixes #776 reported by both @nornagon and @miniak. The problem was that we were trying to submit empty files, which fails.

Add a regression test. The tests already checked for missing files; now also test for existing-but-empty files.

*edit:* no newly-untested LOC or logic branches. Coveralls' negative report is because the code size decreased by a few lines.

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -409,19 +409,9 @@ export class GistActionButton extends React.Component<
   };
 
   private gistFilesList = (values: EditorValues) => {
-    const filesList: EditorValues = {};
-
-    // Add files for default editors.
-    for (const editor of DEFAULT_EDITORS) {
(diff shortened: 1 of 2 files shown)


nornagon (approved):
I'd still prefer that the gists were published without the empty files present at all, but I guess that's coming in the future refactor.

codebytere (approved):

ckerr:
> I'd still prefer that the gists were published without the empty files present at all, but I guess that's coming in the future refactor.

Yes that's in https://github.com/electron/fiddle/pull/750.

Issue #753, opened 2021-07-02
output console is rebuilt frequently


ckerr:
It looks like the MonacoEditor in components/output is being regenerated pretty frequently when there's new console output pushed to appState.output. During a session of `yarn start bisect v11.2.0 v10.0.0 --fiddle https://gist.github.com/ckerr/8c5fc0c6a5153d49b5a4a56d3ed9da8f --log-config --obsolete` the console is rebuilt 60 times.

Possibly introduced in #717?

Sample trace:

```
===> Error: trace
    at Output.destroyMonacoEditor (main.99d03ea8.js:9026)
    at main.99d03ea8.js:8974
    at reactionRunner (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:2106)
    at trackDerivedFunction (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:776)
    at Reaction.track (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1806)
    at Reaction.onInvalidate (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:2087)
    at Reaction.runReaction (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1772)
    at runReactionsHelper (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1909)
    at reactionScheduler (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1887)
    at /home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1916
    at batchedUpdates$1 (/home/charles/electron/fiddle/node_modules/react-dom/cjs/react-dom.development.js:21856)
    at reactionScheduler (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1916)
    at runReactions (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1892)
    at endBatch (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:1592)
    at _endAction (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:980)
    at executeAction (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:934)
    at AppState.pushOutput (/home/charles/electron/fiddle/node_modules/mobx/lib/mobx.js:914)
    at ChildProcess.<anonymous> (main.99d03ea8.js:3818)
instrument.ts:129 Editor: Disposing
[...]

PR #780, opened 2021-07-14
chore: fix output console  regenerating monaco editor upon output update

georgexu99:
This PR addresses #753

 - [x] Refactored the Output component to update a single monaco editor instead of generating a new one
 - [x] Validated on Mac, (needs to be tested on Windows and Linux)


georgexu99:
I put up this draft PR for a visual reference of a quick fix for #753. I don't like how I am manually passing in an `init` parameter to indicate whether we need to generate or update a monaco editor- Will attempt to utilize `editor.getModel()` to determine whether the model exists instead tomorrow, hopefully that will yield a cleaner solution

ckerr:
I really like your idea of having reusing the original model each time; I think you're on the right track. But I don't think this current iteration will work -- the default uri is an empty value, so `outputUri.toString()` is probably `''` or `://` or something and not a great lookup key.

That's OK though, because this class creates the model that it's searching for! When creating it, it can keep a handle to it as a `private` property so that no searching necessary. Moreover, if it's built in the constructor it can be a `private readonly` property so there's no searching necessary _and_ never any ambiguity about whether it's been created yet!

Something like

```diff
-  private outputUri: MonacoType.Uri;
+  private readonly model: MonacoType.editor.ITextModel;
```
and

```diff
   constructor(props: CommandsProps) {
     super(props);
 
     const { monaco } = this.props;
-    this.outputUri = new monaco.Uri();
+    this.model = monaco.editor.createModel('', this.language);
+    this.model.updateOptions({ tabSize: 2 });
+
     this.language = 'consoleOutputLanguage';
```

and then in initMonaco(), the model would be fed into the MonacoEditor being constructed:

```diff
@@ -87,11 +84,9 @@ export class Output extends React.Component<CommandsProps> {
         readOnly: true,
         contextmenu: false,
         automaticLayout: true,
-        model: null,
+        model: this.model,
         ...monacoOptions,
       });
```

one nice side-effect of this is that all of `Output.setEditorText(value: string)` can basically be removed and the places that call it can call `this.model.setValue(value);` instead.

This model will work even if the editor is closed and then a new one is opened later; the same existing model can be fed into the new MonacoEditor exactly the same way in initMonaco()

ckerr:
@georgexu99 this looks really good, I'm happy that creating a model in the constructor worked out so well!

Happy to review whenever this gets out of draft :+1:

ckerr (approved):

PR #783, opened 2021-07-16
refactor: default to latest stable version

ckerr:
On first start, we have to pick which version of Electron to first use. This PR changes behavior to choose the latest stable release.

Sibling PR to https://github.com/electron/fiddle/pull/782 . They're related but can land separately.

### DIFF
--- a/src/renderer/versions.ts
+++ b/src/renderer/versions.ts
@@ -15,37 +15,19 @@ import { normalizeVersion } from '../utils/normalize-version';
  * @param {Array<RunnableVersion>} knownVersions
  * @returns {string}
  */
-export function getDefaultVersion(
-  knownVersions: Array<RunnableVersion> = [],
-): string {
-  const ls = localStorage.getItem('version');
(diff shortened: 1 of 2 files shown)


src/renderer/versions.ts
```diff
@@ -15,37 +15,19 @@ import { normalizeVersion } from '../utils/normalize-version';
...
-  // Self-heal: Version not formated correctly
-  const normalized = ls && normalizeVersion(ls);
-  if (normalized) {
-    if (
-      knownVersions &&
-      knownVersions.find(({ version }) => version === normalized)
-    ) {
-      return normalized;
-    }
-  }
```
ckerr, line 40: I've removed this 'self-heal' code because the only way this code path would get exercise if someone last used Electron Fiddle before we started saving normalized versions in LocalStorage, and is now starting the app for the first time since then. In this case their previously-selected version will have fallen off the main page due to being obsolete, so choosing `latestStable` is actually a safer choice.

codebytere (approved):

PR #792, opened 2021-07-18
feat: add os version + arch info to --log-config

ckerr:
When Electron Fiddle is launched with `--log-config`,  include OS version and architecture in the log.

Pull in [getos](https://www.npmjs.com/package/getos) ([2.2m downloads per week](https://www.npmjs.com/package/getos), [9.1 kB minified + gzipped](https://bundlephobia.com/package/getos@3.2.1)) to get better platform info on Linux.


package.json
```diff
@@ -83,7 +85,7 @@
     "@types/jest": "^26.0.20",
     "@types/json-stable-stringify": "^1.0.32",
     "@types/log-symbols": "^3.0.0",
-    "@types/node": "^12.18.0",
+    "@types/node": "^14.16.0",
```
erickzhao, line 88: Good catch!

erickzhao (approved):

package.json
ckerr: I found it because otherwise I got warnings trying to log `os.version()`. Apparently that's new in 14 :slightly_smiling_face:

PR #795, opened 2021-07-19
feat: allow copying current version number

codebytere:
Addresses a use case i've found myself running into frequently and just got around to figuring out how to address. When doing a bisection, especially if it involves nightlies, it's a pain to re-type out the whole version number again.

This PR seeks to address that by adapting the context menu shown when hovering over the version select component. Now, when you right click the currently selected version, you'll see a single item in the context menu which will copy the version to the user's clipboard.

Before (not very useful!):

Screen Shot 2021-07-19 at 2 18 04 PM

After (much more useful!):

Screen Shot 2021-07-20 at 12 36 11 PM


ckerr:
This is really cool!

erickzhao (commented):
question: Would it make sense to avoid using IPC and use Blueprint's https://blueprintjs.com/docs/#popover2-package/context-menu2 component?

codebytere:
@erickzhao potentially! happy to take a stab at it today

codebytere:
@erickzhao @ckerr done!

ckerr:
This reminds me, we should probably start the Blueprint 4 migration chores listed @ https://github.com/palantir/blueprint/wiki/Blueprint-4.0

ckerr (approved):
Short & sweet.

Issue #747, opened 2021-06-28
New logging window makes it impossible to copy/paste logs separately from timestamps


nornagon:
Previously, it was possible to copy/paste just the log messages out of the log window, without the timestamps. Now, it seems the log message and the timestamp are copied together, resulting in pastes like:

```
10:50:23 AM Saving files to temp directory...
10:50:23 AM Saved files to /var/folders/_7/jm5cr84j7jd4v7btf2d3bqwm0000gn/T/tmp-41327-rDvL38ixZjEw
10:50:23 AM Electron v999.999.999 started.
10:50:23 AM [41354:0628/105023.607715:INFO:content_main_runner_impl.cc(1083)] Chrome is running in full browser mode.
10:50:24 AM [41356:0628/105024.165210:WARNING:ca_layer_tree_coordinator.mm(54)] Blank frame: No overlays or CALayers
10:50:25 AM [41354:0628/105025.306726:WARNING:discardable_shared_memory_manager.cc(432)] Some MojoDiscardableSharedMemoryManagerImpls are still alive. They will be leaked.
10:50:25 AM [41354:0628/105025.315777:ERROR:process_posix.cc(330)] Unable to terminate process 41358: No such process (3)
10:50:25 AM 
10:50:25 AM Electron exited with code 0.
```

I'm not sure that the timestamp is all that useful to be copy/pasted? Log messages from `LOG()` already have a timestamp in them, so this is repeated information there. It might be more useful if it were 'time since app launch' and included ms? But I'd also be happy if we could just copy the raw log messages :)

This is especially relevant when a stack trace is printed out, because horizontal real estate is quite limited on github issues.

codebytere:
Closed by https://github.com/electron/fiddle/pull/796

PR #796, opened 2021-07-19
feat: add option to disable output timestamps

georgexu99:
this PR addresses: #747

- [ ] Validated on Mac, (needs to be tested on Windows and Linux)

image

image

image

**EDIT: ** I have added a separate implementation suggested by @nornagon in the latest commit- I wanted to show off both of them to gather opinions on which way to go forward.

The second implementation places the timestamp in the gutter, where the line numbers used to be! This implementation would not need a toggle-menu in the preferences so it may reduce menu clutter (that being said there are currently 2 items in the menu)

image

### DIFF
--- a/src/renderer/components/output.tsx
+++ b/src/renderer/components/output.tsx
@@ -35,6 +35,7 @@ export class Output extends React.Component<CommandsProps> {
   public language = 'consoleOutputLanguage';
 
   private outputRef = React.createRef<HTMLDivElement>();
+  private outputTimeStamps: string[] = [];
   private readonly model: MonacoType.editor.ITextModel;
 
   constructor(props: CommandsProps) {
@@ -71,6 +72,7 @@ export class Output extends React.Component<CommandsProps> {
   public async initMonaco() {
     const { monaco, monacoOptions: monacoOptions } = this.props;
     const ref = this.outputRef.current;
+    this.getLineNumber = this.getLineNumber.bind(this);
     if (ref) {
       this.setupCustomOutputEditorLanguage(monaco);
       this.editor = monaco.editor.create(ref, {
@@ -81,6 +83,9 @@ export class Output extends React.Component<CommandsProps> {
         automaticLayout: true,
         model: this.model,
         ...monacoOptions,
+        lineNumbers: this.getLineNumber,
+        lineNumbersMinChars: 12,
+        wordWrap: 'on',
       });
     }
   }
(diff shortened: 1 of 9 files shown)


georgexu99:
Currently looking into adding the timestamp into the gutter (line numbers) of the output monaco editor to reduce settings clutter

src/renderer/state.ts
```diff
@@ -82,6 +82,7 @@ export class AppState {
   @observable public isClearingConsoleOnRun = !!this.retrieve(
     'isClearingConsoleOnRun',
   );
+  @observable public isShowingTimestamp = !!this.retrieve('isShowingTimestamp');
```
ckerr, line 85: This needs to be followed by the default value, similar to the `?? true` in `isUsingSystemTheme` below.

src/renderer/components/output.tsx
```diff
@@ -170,11 +170,15 @@ export class Output extends React.Component<CommandsProps> {
...
 
+    const { isShowingTimestamp } = this.props.appState;
+
     for (const output of outputs) {
       const segments = output.text.split(/\r?\n/);
+
       const date = new Date(output.timestamp).toLocaleTimeString();
       for (const segment of segments) {
-        lines.push(date + ' ' + segment);
+        const line = isShowingTimestamp ? date + ' ' + segment : segment;
```
codebytere, line 180: nit: probably a little cleaner here to do template string: `${date}  ${segment}`

src/renderer/state.ts
codebytere: I'd also say we should have timestamps enabled by default - e.g. the option in settings is to _disable_ it. So the option might instead be something like `isDisablingTimestamps`

ckerr: Agree that the default value should be to show the timestamps.

*bikeshed*: I prefer the current name. All other things being equal, code should avoid negative logic because it leads to double-negative expressions e.g. expressing "are timestamps enabled?" as `!isDisablingTimestamps`.

georgexu99: Thanks for the reviews! I'll take action on both of these suggestions 🙇

nornagon:
FWIW I prefer the gutter solution!

src/renderer/components/output.tsx
```diff
@@ -96,6 +101,10 @@ export class Output extends React.Component<CommandsProps> {
     }
   }
 
+  public getLineNumber(originalLineNumber: number) {
+    return this.outputTimeStamps[originalLineNumber - 1];
```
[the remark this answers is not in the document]
georgexu99: Right! I assumed since we initialize our output with "Output ready 🔬 " we would always have something there..just realized that that was a horrible choice from a maintainable standpoint 😔

codebytere (approved):

ckerr (approved):

ckerr:
CI failure appears to be unrelated

ckerr (approved):

Issue #790, opened 2021-07-17
bug: bisect can reverse endpoints when good or bad version is a nightly


ckerr:
- Electron's release pattern is nightly -> [alpha, beta] -> stable.
- semver.compare() compares prerelease tags lexicographically, e.g. alpha and beta would come before nightly
- utils/get-version-range uses semver.compare() to ensure that the endpoints are in the "correct" order but thanks to that wart, would place nightlies _after_ alphas & betas

This would be good to fix since the results will be shown in the wrong order, but it's not critical since (I think?) the bisection will still work.

PR #797, opened 2021-07-21
fix: proper release sorting in get-version-range

codebytere:
Closes https://github.com/electron/fiddle/issues/790.

Updates our logic in `get-version-range` to augment `semver.compare`. The new logic takes into account Electron's prerelease tag hierarchy, whereby `nightly -> alpha -> beta`.

### DIFF
--- a/src/utils/get-version-range.ts
+++ b/src/utils/get-version-range.ts
@@ -2,6 +2,56 @@ import * as semver from 'semver';
 
 import { RunnableVersion } from '../interfaces';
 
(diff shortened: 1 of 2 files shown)


ckerr (approved):
This works and is good.

It's out-of-scope for this PR but just to make it explicit -- we now have three thematically-adjacent pieces of code here, in `utils/sort-versions`, and in https://github.com/electron/fiddle/pull/786. IMO the next logical step is to refactor these into One Reusable Sorting Utility to Rule Them All

PR #798, opened 2021-07-21
feat: before exiting, log the exit code.

ckerr:
Does what it says on the tin.

When running in batch mode -- e.g. launched on the command line to do an autotest or autobisect operation and then exit -- be more explicit about the exit code in the log so that anyone reading the session logs via BugBot can see it.


ckerr:
oop, the tests are checking what gets sent to console.log and in my haste I forgot to update them. Fixing now.

codebytere (approved):

PR #802, opened 2021-07-22
refactor: merge version sorting logic

codebytere:
Follow-up to https://github.com/electron/fiddle/pull/797#pullrequestreview-711926906 - merges and cleans up similar sorting logic in `src/utils/get-version-range.ts` and `src/utils/sort-versions.ts`

Notes:
 * In looking at the logic in https://github.com/electron/fiddle/pull/786 i think it's semantically different enough to stay as-is for now 🤔
 * Added an extra filter to the logic in `sortVersions` to remove versions that don't coerce to a valid semver version - there's no use case I see where we'd want to keep them around so imo it's better just to get rid of them.
     * Tests updated accordingly

### DIFF
--- a/src/utils/sort-versions.ts
+++ b/src/utils/sort-versions.ts
@@ -11,20 +11,17 @@ const preTags = ['nightly', 'alpha', 'beta'];
  * @param b Prerelease tag data for the new version.
  * @returns 0 | 1 | -1
  */
-const preCompare = (
-  a: readonly (string | number)[],
-  b: readonly (string | number)[],
-) => {
-  const one = preTags.indexOf(a[0] as string);
-  const two = preTags.indexOf(b[0] as string);
-  if (one === two) {
+const preCompare = (a: string[], b: string[]) => {
+  const first = preTags.indexOf(a[0]);
+  const second = preTags.indexOf(b[0]);
+  if (first === second) {
     // Whether the prerelease tag number is the same
     // e.g. alpha.1 & alpha.1.
     if (a[1] === b[1]) return 0;
     return a[1] > b[1] ? 1 : -1;
   }
 
-  return one > two ? 1 : -1;
+  return first > second ? 1 : -1;
 };
 
 /**
@@ -37,16 +34,25 @@ const preCompare = (
  * @param b The new Electron version.
  * @returns 0 | 1 | -1
  */
-export function semverCompare(a: string, b: string) {
-  const hasPre = (v: string) => preTags.some((tag) => v.includes(tag));
+export function semverCompare(
+  a: string | semver.SemVer,
+  b: string | semver.SemVer,
+) {
+  const pA = typeof a === 'string' ? semver.parse(a) : a;
+  const pB = typeof b === 'string' ? semver.parse(b) : b;
+
+  const sameMain = (a: semver.SemVer | null, b: semver.SemVer | null) =>
+    a !== null && b !== null && a.compareMain(b) === 0;
 
   // Check that major.minor.patch are the same for a and b.
-  if (semver.coerce(a)?.raw === semver.coerce(b)?.raw) {
-    if (hasPre(a) && hasPre(b)) {
-      const { prerelease: aPre } = semver.parse(a) as semver.SemVer;
-      const { prerelease: bPre } = semver.parse(b) as semver.SemVer;
-      return preCompare(aPre, bPre);
-    }
+  if (a === 'v3.0.0' || b === 'v2.0.0') throw new Error('hey');
+  if (
+    sameMain(pA, pB) &&
+    pA?.prerelease.length !== 0 &&
+    pB?.prerelease.length !== 0
+  ) {
+    if (a === 'v3.0.0' || b === 'v3.0.0') throw new Error('hey');
+    return preCompare(pA?.prerelease as string[], pB?.prerelease as string[]);
   }
 
   return semver.compare(a, b);
@@ -59,15 +65,18 @@ export function semverCompare(a: string, b: string) {
  * @returns {RunnableVersion[]}
  */
 export function sortVersions(versions: RunnableVersion[]): RunnableVersion[] {
-  type VerSemRun = [ver: string, run: RunnableVersion];
+  type VerSemRun = [
+    ver: string,
+    sem: semver.SemVer | null,
+    run: RunnableVersion,
+  ];
 
-  const filtered = versions.filter((runnable) =>
-    semver.valid(runnable.version),
-  );
-  const sorted = filtered
-    .map((run): VerSemRun => [run.version, run])
-    .sort(([vera], [verb]) => -semverCompare(vera, verb));
-  sorted.forEach(([_1, run], idx) => (filtered[idx] = run));
-
-  return filtered;
+  const sorted = versions
+    .map((run): VerSemRun => [run.version, semver.parse(run.version), run])
+    .sort(
+      ([vera, sema], [verb, semb]) =>
+        -semverCompare(sema || vera, semb || verb),
+    );
+  sorted.forEach(([_1, _2, run], idx) => (versions[idx] = run));
+  return versions;
 }
--- a/src/utils/sort-versions.ts
+++ b/src/utils/sort-versions.ts
@@ -2,28 +2,54 @@ import * as semver from 'semver';
 
 import { RunnableVersion } from '../interfaces';
 
-function electronSemVerCompare(a: semver.SemVer, b: semver.SemVer) {
-  const l = a.compareMain(b);
-  if (l) return l;
-  // Electron's approach is nightly -> other prerelease tags -> stable,
-  // so force `nightly` to sort before other prerelease tags.
-  const [prea] = a.prerelease;
-  const [preb] = b.prerelease;
-  if (prea === 'nightly' && preb !== 'nightly') return -1;
-  if (prea !== 'nightly' && preb === 'nightly') return 1;
-  return a.comparePre(b);
-}
+const preTags = ['nightly', 'alpha', 'beta'];
 
-function compare(a: semver.SemVer | string, b: semver.SemVer | string) {
-  // non-semver goes first, sorted lexicographically
-  const astr = typeof a === 'string';
-  const bstr = typeof b === 'string';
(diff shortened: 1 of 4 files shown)


src/utils/sort-versions.ts
```diff
@@ -2,28 +2,54 @@ import * as semver from 'semver';
...
+ * @param a Prerelease tag data for the old version.
+ * @param b Prerelease tag data for the new version.
+ * @returns 0 | 1 | -1
+ */
+const preCompare = (
+  a: readonly (string | number)[],
+  b: readonly (string | number)[],
+) => {
+  const one = preTags.indexOf(a[0] as string);
+  const two = preTags.indexOf(b[0] as string);
```
ckerr, line 19: It seems unsafe to array of strings *or* an array of numbers, but then to unconditionally treat it as an array of strings

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,54 @@ import * as semver from 'semver';
...
+  if (one === two) {
+    // Whether the prerelease tag number is the same
+    // e.g. alpha.1 & alpha.1.
+    if (a[1] === b[1]) return 0;
+    return a[1] > b[1] ? 1 : -1;
```
ckerr, line 24: These two comparisons can be folded into a single compare by replacing both lines with a.localeCompare(b)

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,54 @@ import * as semver from 'semver';
...
+ * tag hierarchy.
+ *
+ * Sorts in ascending order when passed to Array.sort().
+ *
+ * @param a The old Electron version.
+ * @param b The new Electron version.
+ * @returns 0 | 1 | -1
+ */
+export function semverCompare(a: string, b: string) {
+  const hasPre = (v: string) => preTags.some((tag) => v.includes(tag));
```
ckerr, line 41: faster [and (I think) correct](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions):

`const hasPre = v.includes('-')`

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,54 @@ import * as semver from 'semver';
...
+
+  // Check that major.minor.patch are the same for a and b.
+  if (semver.coerce(a)?.raw === semver.coerce(b)?.raw) {
+    if (hasPre(a) && hasPre(b)) {
+      const { prerelease: aPre } = semver.parse(a) as semver.SemVer;
+      const { prerelease: bPre } = semver.parse(b) as semver.SemVer;
+      return preCompare(aPre, bPre);
+    }
```
ckerr, line 49: - Is coerce() needed here, or will parse() do just as well? I don't think I understand the need for coercion here.  I *think* we already ensure semver correctness when Fiddle loads up the version array to begin with.

- the values of coerce() are discarded, only to turn around and re-parse the same string all over again 2-3 lines below

- And all of this is done in the compare func, meaning that we're back to doing something like 4 * O(n^2) parses again. This discards the "parse in O(N), not O(N^2)" behavior that was in the implementation of sortVersions()

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,54 @@ import * as semver from 'semver';
...
-  if (l) return l;
-  // Electron's approach is nightly -> other prerelease tags -> stable,
-  // so force `nightly` to sort before other prerelease tags.
-  const [prea] = a.prerelease;
-  const [preb] = b.prerelease;
-  if (prea === 'nightly' && preb !== 'nightly') return -1;
-  if (prea !== 'nightly' && preb === 'nightly') return 1;
-  return a.comparePre(b);
-}
+const preTags = ['nightly', 'alpha', 'beta'];
```
ckerr, line 5: This is a nice way of doing it. I think this is better than my previous approach in `electronSemVerCompare()`

src/utils/sort-versions.ts
```diff
@@ -33,14 +59,15 @@ function compare(a: semver.SemVer | string, b: semver.SemVer | string) {
...
-  const sorted = versions
-    .map((run): VerSemRun => [run.version, semver.parse(run.version), run])
-    .sort(([vera, sema], [verb, semb]) => compare(sema || vera, semb || verb));
-  sorted.forEach(([_1, _2, run], idx) => (versions[idx] = run));
-  return versions;
+  type VerSemRun = [ver: string, run: RunnableVersion];
+
+  const filtered = versions.filter((runnable) =>
+    semver.valid(runnable.version),
+  );
```
ckerr, line 66: `semver.valid()` is literally just a boolean wrapper around semver.parse() so this is another single-time-use parse

ckerr (commented):
I love the idea of merging these together, thank you for doing this!

I think this needs another iteration before it's ready though.

ckerr:
> * Added an extra filter to the logic in `sortVersions` to remove versions that don't coerce to a valid semver version - there's no use case I see where we'd want to keep them around so imo it's better just to get rid of them.

I don't understand where the non-coercing versions come from?

src/utils/sort-versions.ts
codebytere: we can work around this if we'd like, but this was done because it's the required type for `Array.sort()`

src/utils/sort-versions.ts
codebytere: this is a workaround for checking that `a.major === b.major && a.minor === b.minor && a.patch === b.patch` - i'll rework it.

src/utils/sort-versions.ts
codebytere: you mean you'd prefer to coerce them to a string? all indices after 0 in a pre array are numbers

codebytere:
@ckerr this function previously didn't remove garbage versions which were specifically tested for - like the calls to `makeVersion('moreGarbage')` was previously sorted and would remain in the sorted version list whereas after this change it's simply removed.

src/utils/sort-versions.ts
codebytere: technically this is an array where the first index is a string and the rest are numbers - not sure the best way to type this tbh

src/utils/sort-versions.ts
ckerr: lol! No, I didn't realize that the first was a string and the remainder were numbers :smiley:

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,56 @@ import * as semver from 'semver';
...
+ *
+ * Sorts in ascending order when passed to Array.sort().
+ *
+ * @param a The old Electron version.
+ * @param b The new Electron version.
+ * @returns 0 | 1 | -1
+ */
+export function semverCompare(a: string, b: string) {
+  const pA = semver.parse(a);
+  const pB = semver.parse(b);
```
ckerr, line 39:
```suggestion
export function semverCompare(a: string | semver.SemVer, b: string | semver.SemVer) {
  const pA = typeof a === 'string' ? semver.parse(a) : a;
  const pB = typeof b === 'string' ? semver.parse(b) : b;
```

src/utils/sort-versions.ts
```diff
@@ -33,14 +61,15 @@ function compare(a: semver.SemVer | string, b: semver.SemVer | string) {
...
+
+  const filtered = versions.filter((runnable) =>
+    semver.parse(runnable.version),
+  );
+  const sorted = filtered
+    .map((run): VerSemRun => [run.version, run])
+    .sort(([vera], [verb]) => -semverCompare(vera, verb));
+  sorted.forEach(([_1, run], idx) => (filtered[idx] = run));
+
+  return filtered;
```
ckerr, line 74: keep old implementation but replace

```diff
-    .sort(([vera, sema], [verb, semb]) => compare(sema || vera, semb || verb));
+    .sort(([vera, sema], [verb, semb]) => -semverCompare(sema || vera, semb || verb));
```

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,56 @@ import * as semver from 'semver';
...
+ */
+export function semverCompare(a: string, b: string) {
+  const pA = semver.parse(a);
+  const pB = semver.parse(b);
+
+  const sameMMP = (a: semver.SemVer | null, b: semver.SemVer | null) => {
+    if (!a || !b) return false;
+
+    return a.major === b.major && a.minor === b.minor && a.patch === b.patch;
+  };
```
ckerr, line 45:
```suggestion
  const sameMain = (a: semver.SemVer | null, b: semver.SemVer | null) =>
    a && b && a.compareMain(b) === 0;
```

src/utils/sort-versions.ts
```diff
@@ -2,28 +2,56 @@ import * as semver from 'semver';
...
+
+  // Check that major.minor.patch are the same for a and b.
+  if (sameMMP(pA, pB)) {
+    if (a.includes('-') && b.includes('-')) {
+      return preCompare(pA?.prerelease as string[], pB?.prerelease as string[]);
+    }
+  }
```
ckerr, line 52:
```suggestion
  if (sameMMP(pA, pB) && a.includes('-') && b.includes('-') {
    return preCompare(pA?.prerelease as string[], pB?.prerelease as string[]);
  }
```

ckerr:
> @ckerr this function previously didn't remove garbage versions (this was tested) - like the calls to `makeVersion('moreGarbage')` was previously sorted and would remain in the sorted version list whereas after this change it's simply removed.

OK I hear you that this is in the tests, but is there actually a path in the code to where we can have a version number named 'moreGarbage' when users are actually running the program?

ckerr:
(Whoops, hit wrong button; didn't mean to close)

src/utils/sort-versions.ts
codebytere: hmm - isn't this what it is now? not sure what i need to replace since i'm already using `-semverCompare` - would you mind clarifying?

src/utils/sort-versions.ts
codebytere: TIL `compareMain`!! that's exactly what i was after 🌟

codebytere:
@ckerr probably not (or at least not that i could discern) - would you prefer i do something else?

ckerr (approved):

PR #810, opened 2021-07-30
refactor: use union types

ckerr:
Prefer union types over enums in `src/interfaces.ts`.


clavin:
@ckerr What is your rationale for this PR?

ckerr:
You know, my initial motive was to reduce potential diffs with electron-fiddle-runner in the future, but there are easier ways to do that which cause less churn. I'd be OK with this landing but I think it's less a serious PR and more an experiment.

PR #816, opened 2021-07-30
test: reduce copy paste in get-package-spec and remote-loader-spec

ckerr:
Tests only.  No functional changes.

This PR reduces some test copy-paste that's built up over time.

In the interest of keeping PRs reasonably small and easy to review, this only works through two test suites:  get-package and remote-loader.


ckerr:
hmm, shouldn't've changed coverage at all. I'll take a look.

PR #817, opened 2021-07-30
test: reduce copy-paste in app-spec

ckerr:
Tests only. No functional changes.

Part of a series of small refactors to reduce copy-paste that's built up over time in tests/.

This PR is for app-spec.


erickzhao (approved):
Seems legit to me!

PR #825, opened 2021-08-05
feat: more descriptive local version names

ckerr:
Looking for feedback on this as well as code review -- this PR is better for _my_ workflow but I'd like another opinion not just on the code but on the idea, since other people may have other workflows.

One small annoyance is that Electron Fiddle names all my local builds something like `src/out/release` or `src/out/testing`. Since `src/out/` is literally in every local build, it's not the most useful fragment to identify a version. Usually I keep checkouts in something like `.../main/src/out/...` or `.../13/src/out/...` so that parent directory would be more helpful.

When given a path like `/home/username/electron/gn/main/src/out/testing`, instead of returning `src/out/testing`, this PR returns `gn/main - testing`.

So, like I say -- this is helpful _to me_, but I'd like a sanity check about whether or not this works as a general-purpose change.

What do your folder names look like?

### DIFF
--- a/src/renderer/components/dialog-add-version.tsx
+++ b/src/renderer/components/dialog-add-version.tsx
@@ -32,6 +32,31 @@ interface AddVersionDialogState {
 }
 
 /**
+ * Build a default name for a local Electron vresion
+ * from its dirname.
+ *
+ * @param {string} dirname
+ * @return {string} human-readable local build name
+ */
+function makeLocalName(folderPath: string) {
+  // take a dirname like '/home/username/electron/gn/main/src/out/testing'
+  // and return something like 'gn/main - testing'
+  const tokens = folderPath.split(path.sep);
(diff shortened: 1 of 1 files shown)


codebytere (commented):
@ckerr where exactly is this happening in the UI? I just see my semver names and that's all

ckerr:
Here's a screenshot.

999.0.0-main's name ('src/out/testing') comes from the name generator in master.

The two _incredibly_ cleverly-named temporary builds, scratch and scratch2, were in ~/electron/gn/scratch and ~/electron/gn/scratch2 with version numbers 999.0.0-scratch and 999.0.0-scratch2. Their names ('gn/scratch - testing' and 'gn/scratch2 - testing') came from the name generator in this PR.

screenshot

codebytere:
ahh gotcha! ty.

codebytere (approved):
i'm good with this for now - it's an incremental improvement at the very least and we can always later decide to tweak it!

PR #831, opened 2021-08-12
fix: console css styles

codebytere:
Refs https://github.com/electron/fiddle/pull/717.

After refactor:

Screen Shot 2021-08-12 at 9 48 07 AM

Pre-refactor and post this PR:

Screen Shot 2021-08-12 at 9 58 16 AM


erickzhao (approved):
Good catch! LGTM

PR #813, opened 2021-07-30
build: bump forge dependency

ckerr:
This bumps Electron Fiddle's forge version to `6.0.0-beta.59`.

In package.json, it had been `-beta.55`.

In `src/renderer/transforms/forge.ts`, it had been `-beta.52`, so a followup task might be to refactor this code to keep them in sync...

PR #835, opened 2021-08-12
build(deps): yarn.lock update

erickzhao:
Running the `yarn install` command on the current `trunk` gives me a bunch of lockfile updates. Seems like a follow-up to #813.


malept (approved):
:shrug: seems legit.

andersk:
Running `yarn --frozen-lockfile` in CI would have caught this mistake: #848.

PR #839, opened 2021-08-13
chore: make sure static JS files use standard

malept:
The lint-staged pre-commit rules ran both ESLint and standard on JS files in `static/show-me`, with ESLint (+ prettier) always winning (and contradicting `yarn lint`). This PR makes ESLint ignore the `static/` folder and fixes up any files that were formatted with prettier due to lint-staged.


ckerr (approved):
:+1:

PR #840, opened 2021-08-13
ci: refactor GitHub Actions for caching / network timeouts

malept:
* Adds Dependabot config for GitHub Actions
  * I deliberately didn't want to update JS dependencies here. If we want to do that we should tackle that in a separate PR (and also discuss whether those PRs should get auto-merged when tests pass)
*  Refactor the setup-node/cache config since setup-node now comes with caching built-into it
* Add yarn network timeout from #833 to the build step (it failed [here](https://github.com/electron/fiddle/runs/3326328222?check_suite_focus=true))

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,7 +12,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2.3.4
       - name: Setup Node.js
         uses: actions/setup-node@v2.4.0
         with:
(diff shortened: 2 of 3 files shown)


malept:
```
Run yarn --network-timeout 100000
yarn install v1.22.5
[1/4] Resolving packages...
[2/4] Fetching packages...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/monaco-editor/-/monaco-editor-0.21.3.tgz: unexpected end of file".
info If you think this is a bug, please open a bug report with the information provided in "/Users/runner/work/fiddle/fiddle/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Process completed with exit code 1.
```

Well, that didn't work :sob:

ckerr (commented):
Looks reasonable on the face of it.

Is there any magic going on with hardcoding the action version numbers? I see we bumped from setup-node@v1 to setup-node@v2.4.0 but mostly I'm wondering about the rationale for pinning to a specific patch number?

malept:
> Is there any magic going on with hardcoding the action version numbers? I see we bumped from setup-node@v1 to setup-node@v2.4.0 but mostly I'm wondering about the rationale for pinning to a specific patch number?

It's for a couple of reasons:

1. It's more obvious when a build fails, that it's more likely to be the app code and not that an action upgraded itself between runs.
2. When Dependabot upgrades GitHub Actions, they use a specific version number instead of a version range/branch name.

ckerr (approved):

PR #843, opened 2021-08-19
fix: removing a missing local version from UI

codebytere:
Fixes an issue where missing local versions wouldn't be removed from the UI because we branched on a conditional (version state being ready) that wouldn't apply to local versions (they'd be unknown).


ckerr (approved):
LGTM

PR #848, opened 2021-09-02
ci: Run yarn with --frozen-lockfile

andersk:
This is the [recommended](https://classic.yarnpkg.com/en/docs/cli/install/) way to run Yarn in CI, in order to catch pull requests like #813 that leave package.json and yarn.lock out of sync.


erickzhao (approved):
👍

PR #850, opened 2021-09-03
fix(linux): enable protocol handler for deb/rpm

malept:
I was in the middle of fixing up the protocol handler guide in the Electron docs and I realized that Fiddle's URLs don't work on Linux. It turns out that you need to [add an `x-scheme-handler` to the `MimeType`](https://specifications.freedesktop.org/shared-mime-info-spec/0.21/ar01s02.html#idm45917071429120) in the Electron app's `.desktop` file (which is generated by the various `electron-installer-*` modules for Linux).

While I was in there I synced the common config for the Linux makers (the rpm was missing the icon).

I tested this on Ubuntu 18.04 by running:

```shell
yarn make --targets=@electron-forge/maker-deb
sudo apt install ./out/make/deb/x64/electron_fiddle_0.26.0_amd64.deb
```

...and then loading an Electron Fiddle URL.

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -14,6 +14,13 @@ if (process.env['WINDOWS_CODESIGN_FILE']) {
   }
 }
 
+const commonLinuxConfig = {
+  icon: {
+    scalable: path.resolve(iconDir, 'fiddle.svg'),
+  },
+  mimeType: ['x-scheme-handler/electron-fiddle'],
+};
+
 const config = {
   hooks: {
     generateAssets: require('./tools/generateAssets'),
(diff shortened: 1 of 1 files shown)


erickzhao (approved):

solsticedhiver:
Does not work for me. Well, to be precise, on archlinux, adding the mimetype in the .desktop file, make electron-fiddle shows up, but it always open the "Hello World" fiddle. Not the one requested.

What is the command to launch a fiddle from cli ?
I have tried:
```
electron-fiddle electron-fiddle://electron/v15.3.1/docs/fiddles/quick-start
electron-fiddle test --fiddle electron-fiddle://electron/v15.3.1/docs/fiddles/quick-start
electron-fiddle start --fiddle electron-fiddle://electron/v15.3.1/docs/fiddles/quick-start
```
None of them work as expected. Always opening the default "Hello World" fiddle

Issue #822, opened 2021-08-02
 "white screen of death" during startup when using electron >= v13.0.0-beta.14


ckerr:
I saw this before and mentioned it in https://github.com/electron/fiddle/pull/760, and came back to look at it again. I still don't understand it, so I'm filing a tracking issue :slightly_smiling_face: 

I still don't know the cause of it, e.g. whether it's an Electron issue, or something that EF is doing in its own code, but the behavior is that if EF's electron dependency in package.json is bumped to v13.0.0-beta.14 or higher, EF's main window is a frozen white screen for 1-2 seconds before the current startup behavior kicks in. By contrast, v13.0.0-beta.13 behaves for me the same as the current 12.0.14 dependency.

https://github.com/electron/electron/compare/v13.0.0-beta.13...v13.0.0-beta.14 indicates the main change between those betas was a bump from Chromium 90 to 91.

nornagon:
@ckerr is there any upstream issue in electron/electron tracking this? If EF stayed unchanged and the behavior appeared with an Electron version bump, it seems pretty likely to me that it could be an Electron issue.

Would be great to get a reduced repro, too.

PR #858, opened 2021-09-22
chore: upgrade to Electron 15

erickzhao:
As a side effect, this PR hides the window by default (see #822). We already had a handler to `show()` the window when the `webContents` were ready, but it seems like the `show: false` by default was inverted in https://github.com/electron/fiddle/commit/a6e6fad4d901a16c7d76569c627137bb72ef7a11.

This sets it back to avoid the flash of white content described in the aforementioned issue.


malept (approved):
Seems legit?

ckerr (approved):

PR #865, opened 2021-09-29
feat: editor tree in sidebar

erickzhao:
This PR mainly replaces the `commands-editor.tsx` component in the commands bar with a new sidebar next to the editors. I think this is an improvement since it aligns with other editors' UX, and makes the editor list visible with 0 clicks.

Screenshot of change

The only orthogonal change is a code deletion in ddfc727, which allows the user to have 0 editors open and exposes the non-ideal state when that happens:

non-ideal state

I think giving users more flexibility like this is more intuitive.

cc @codebytere


src/less/components/sidebar.less
```diff
@@ -0,0 +1,14 @@
...
+  font-size: 14px;
+}
+
+.add-file-input .bp3-tree-node-label {
+  overflow: visible;
+}
+
+.bp3-tree-node-caret-none {
+  min-width: 8px;
+}
```
ckerr, line 14: On a new startup, the sidebar takes up enough space that a lot of the editor text is wrapped.

What would you think of some combo of narrowing the sidebar / making the whole window wider?

src/renderer/components/sidebar-file-tree.tsx
```diff
@@ -0,0 +1,158 @@
...
+          id: index,
+          hasCaret: false,
+          icon: 'document',
+          label: (
+            <ContextMenu2
+              content={
+                <Menu>
+                  <MenuItem
+                    icon="remove"
+                    text="Delete"
```
ckerr, line 63: This button should be disabled if the file is required (see `isRequiredFile()`).

If we try to run Electron on a fiddle where `main.js` has been removed, we're going to have a bad time :slightly_smiling_face:

ckerr (changes requested):
I had a couple of requests, but this looks really cool!

src/less/components/sidebar.less
erickzhao: Addressed in c39632f

made the size `1400` and the sidebar tile % smaller.

src/renderer/components/sidebar-file-tree.tsx
erickzhao: Addressed in d4d32a6

ckerr (approved):

PR #868, opened 2021-10-04
chore: add tooltips to file tree actions

codebytere:
In the new file tree layout, it's not always clear to a new user what each button does. This adds some tooltips after a brief delay to make clearer what the function of each button is.

Screen Shot 2021-10-04 at 11 29 06 AM
Screen Shot 2021-10-04 at 11 29 03 AM

### DIFF
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -251,7 +251,7 @@ export class RemoteLoader {
   private handleLoadingFailed(error: Error): false {
     const failedLabel = `Loading the fiddle failed: ${error.message}`;
     this.appState.showErrorDialog(
-      navigator.onLine
+      this.appState.isOnline
         ? failedLabel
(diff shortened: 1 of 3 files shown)


src/renderer/remote-loader.ts
```diff
@@ -251,7 +251,7 @@ export class RemoteLoader {
   private handleLoadingFailed(error: Error): false {
     const failedLabel = `Loading the fiddle failed: ${error.message}`;
     this.appState.showErrorDialog(
-      navigator.onLine
+      this.appState.isOnline
```
codebytere, line 254: Small consistency cleanup from https://github.com/electron/fiddle/pull/862

ckerr (commented):
Looks good!

Only request is to update the snapshots to make the tests green again

erickzhao (approved):
LGTM

Issue #866, opened 2021-09-29
Updating gist with removed file won't delete it from gist


erickzhao:
> The root cause is we update the gist here with octokit.rest.gists.update where "Files from the previous version of the gist that aren't explicitly changed during an edit are unchanged."

> Since the files have been removed and Fiddle's internal modified EditorValues don't know what's been removed, we probably need to fetch the old copy, diff them, and explicitly remove the files that way. (Also, it's not completely clear how to remove a file from a gist using github's API... 🤔)

https://github.com/electron/fiddle/pull/750#issuecomment-929168919

PR #869, opened 2021-10-04
fix: delete gist files on update

codebytere:
Closes https://github.com/electron/fiddle/issues/866.

Gist files are deleted by setting content to an empty string, so when we go to update a gist we fetch the old gist first and set any files that exist in the old gist and not the new gist to empty string in order to delete them.


src/renderer/components/commands-action-button.tsx
```diff
@@ -175,9 +175,19 @@ export class GistActionButton extends React.Component<
...
 
     try {
+      const {
+        data: { files: oldFiles },
+      } = await octo.gists.get({ gist_id: appState.gistId! });
+
+      const files = this.gistFilesList(values);
+      for (const id of Object.keys(oldFiles)) {
+        // Gist files are deleted by setting content to an empty string.
+        if (!Object.keys(files).includes(id)) files[id] = { content: '' };
```
ckerr, line 185: This seems inefficient -- one, we're rebuilding the array each time we iterate the loop, and two, we trade away any v8 optimizations for object property lookups and use an array linear search instead

```suggestion
        if (!(id in files)) files[id] = { content: '' };
```

ckerr (changes requested):
This is great! Thank you for finding out that an empty string deletes the file.

src/renderer/components/commands-action-button.tsx
erickzhao: TIL the `in` operator exists 🙈

ckerr (dismissed):

ckerr (commented):
Tests need to be synced up to this change but otherwise LGTM!

erickzhao (approved):
LGTM

Issue #725, opened 2021-06-18
Update Show Me demos to support new `contextIsolation` defaults


erickzhao:

thewheat:
Having been bitten by some outdated Electron code in the past, I've fixed the Notification and WebFrame demos. Hope that's ok 😅

akgupta0777:
is this still an issue ?

PR #888, opened 2021-10-28
fix(show-me): update clipboard and shell demos with preload

deepak-sreekumar:
Part of [Update Show Me demos to support new contextIsolation defaults](https://github.com/electron/fiddle/issues/725) #725 
Updated show me demos for `clipboard` and `shell` examples with `preload`

PR #889, opened 2021-10-30
fix(show-me/notification): refactor to use current Electron security defaults

thewheat:
- Part of #725 
- The current 'Show Me' > Notification example does not work
- Updating to use `nodeIntegration: false` and `contextIsolation: true` similar to https://github.com/electron/fiddle/pull/837

### DIFF
--- a/static/show-me/notification/index.html
+++ b/static/show-me/notification/index.html
@@ -12,8 +12,5 @@
     notification APIs to display it.
   </p>
   <button>Send Notification</button>
-  <script>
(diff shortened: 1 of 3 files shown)


deepak-sreekumar:
@thewheat For some reason, I was having trouble with this when I tried fixing it as part of #888 . Thanks a lot for taking it up!

ckerr (approved):

erickzhao (approved):

PR #895, opened 2021-11-09
fix(show-me/screen): refactor to use current Electron security defaults

thewheat:
- Part of https://github.com/electron/fiddle/issues/725
- The current 'Show Me' > Screen example does not work
- Used IPC to get `screen.getAllDisplays()` values from `main.js`values
   - `screen` was showing as null in the `preload.js` page
   - If there is a better way to achieve this? 

**Preview of code change working**
image

### DIFF
--- a/static/show-me/screen/preload.js
+++ b/static/show-me/screen/preload.js
@@ -1,6 +1,9 @@
 const { ipcRenderer } = require('electron')
 
 window.addEventListener('DOMContentLoaded', () => {
-  const screens = ipcRenderer.sendSync('get-displays')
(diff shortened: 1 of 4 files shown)


static/show-me/screen/preload.js
```diff
@@ -0,0 +1,6 @@
+const { ipcRenderer } = require('electron')
+
+window.addEventListener('DOMContentLoaded', () => {
+  const screens = ipcRenderer.sendSync('get-displays')
```
erickzhao, line 4: Can we use `invoke/handle` instead of `sendSync/on` here?

erickzhao (approved):

PR #904, opened 2021-11-16
chore: release v0.27.0

erickzhao:
Note: this removes the unsupported Linux ia32 release binary.


malept (approved):

malept:
For possibly a future PR: it looks like each CI job is running for both pushes and PRs, which means that for PRs, it gets run twice. I think it should only be run on the default branch and PRs.

PR #912, opened 2021-11-27
chore: improve macOS 12 support

### LINKED ISSUE #905: Loading the fiddle failed: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
Electron 16.0.0 (alternatively 13.1.7)
Electron Fiddle 0.26.0 (12.0.14)

Tested with several of these: https://www.electronjs.org/docs/latest/tutorial/examples
image

dsanders11:
Fixes #905.

It looks like in certain situations, there can be an error on macOS 12 due to an older version of `macos-release` (transitive dependency) which doesn't include macOS 12 in the list of known macOS versions.

This is a band-aid fix using ["resolutions" in `yarn`](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) to bump the version of `macos-release` (still within the supported range) despite it being a transitive dependency. A real fix would be to upgrade Fiddle's direct dependency, `@octokit/rest` to a newer version, probably v18, since `macos-release` is pulled in transitively through it.

However, upgrading `@octokit/rest` may be a larger task (I didn't attempt it), where as this PR could fix the issue on macOS 12 right away.

@ckerr, @erickzhao

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,9 @@
     "ts-jest": "^26.5.2",
     "typescript": "^4.2.2"
   },
+  "resolutions": {
+    "macos-release": "^2.5.0"
+  },
   "lint-staged": {
     "./**/*.{js,ts,tsx}": [
       "eslint --fix"
--- a/yarn.lock
+++ b/yarn.lock
@@ -8046,10 +8046,10 @@ lzma-native@^6.0.1:
     readable-stream "^2.3.5"
     rimraf "^2.7.1"
 
-macos-release@^2.2.0:
-  version "2.4.1"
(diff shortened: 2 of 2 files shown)


malept:
I don't think there needs to be a resolution since  the upgrade is within the supported range.

dsanders11:
@malept, is there another way to update `yarn.lock` without manually editing it? I could drop the resolution, it just felt like a better way to update `yarn.lock` since with the resolution a simple `yarn install` updates it for you, without manually editing it. It also makes it more explicit that something has been changed with `macos-release`.

I can do either way, just explaining my thought process here.

MarshallOfSound:
@dsanders11 `npx uuaw macos-release` will update it to the highest it can go within current ranges

PR #913, opened 2021-11-27
fix: set version correctly when switching to a local version

### LINKED ISSUE #778: [bug]: TypeError: Cannot read property 'version' of undefined
Seen on Sentry.

0.25.2.

Breadcrumbs:

```
State: setVersion() got an unknown version D:Study Folder\electron\electron-v1.3.15-win32-x64
{
arguments: [
State: setVersion() got an unknown version D:\Study Folder\electron\electron-v1.3.15-win32-x64
], 
logger: console
}
exception
TypeError: Cannot read property 'version' of undefined
```

The warning and error are here at https://github.com/electron/fiddle/blob/v0.25.2/src/renderer/state.ts#L466 :+1: 

```ts
  @action public async setVersion(input: string) {
    const ver = this.getVersion(input);
    if (!ver) {
      console.warn(`State: setVersion() got an unknown version ${input}`);
      await this.setVersion(this.versionsToShow[0].version);
      return;
    }
```

So, (1), `this.versionsToShow` is an empty array. Since we (probably) always have a set of versions to use -- e.g. even if the session is offline and can't download a new list, we ship with a json file of known releases at the time that Electron Fiddle's binary was packaged -- so, just as a guess, maybe there's nothing to show because all something's in the version textentry causing everything to be filtered out?

And (2), `setVersion()` was called not with a semver version, but with an entire filepath. So at some point we're not doing input sanitization to ensure version numberts are semver. ... TBH I don't even have a guess on this one, I'm not sure where a non-semver version can be introduced into AppState.

dsanders11:
Fixes #778.

cc @ckerr, @erickzhao

### DIFF
--- a/src/renderer/components/dialog-add-version.tsx
+++ b/src/renderer/components/dialog-add-version.tsx
@@ -138,7 +138,7 @@ export class AddVersionDialog extends React.Component<
     // swap to old local electron version if the user adds a new one with the same path
     if (isValidElectron && existingLocalVersion?.localPath) {
       // set previous version as active version
-      this.props.appState.setVersion(existingLocalVersion.localPath);
+      this.props.appState.setVersion(existingLocalVersion.version);
     } else {
       this.props.appState.addLocalVersion(toAdd);
     }


erickzhao (approved):
Good catch!

PR #919, opened 2021-11-30
fix: set MIT license in `package.json` transform

erickzhao:


malept (approved):
Looks legit. Do we have tests for the transform?

erickzhao:
Yep, and I forgot to update them in the initial commit. See 3c10928

PR #930, opened 2022-01-04
fix: update inspector output from Node

dsanders11:
The output was changed in https://github.com/nodejs/node/commit/f3e107aeef2f75cb3a705fd15260f65dd77e78d4 and was first released in v10.0.0, so every supported version of Electron uses the up-to-date version of the output.

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -653,7 +653,8 @@ export class AppState {
     }
 
     if (strData.startsWith('Debugger listening on ws://')) return;
-    if (strData === 'For help see https://nodejs.org/en/docs/inspector') return;
+    if (strData === 'For help, see: https://nodejs.org/en/docs/inspector')
+      return;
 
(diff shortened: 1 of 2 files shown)


erickzhao (changes requested):
Could we change this for backward compatibility reasons? I can see many valid uses of supporting obsolete Electron versions for maintainers.

dsanders11:
@erickzhao, this doesn't have any functional impact, it's just intended to prevent clutter in the console output, and it wasn't working anyway. The output changed in Electron v3.0.0, so it hasn't worked in a long time, so all this will do is invert the current behavior - it will correctly ignore the output for every version of Electron from v3.0.0 forward, and not for versions before v3.0.0.

erickzhao (approved):
Electron 3 teeters the edge of "really old" where a change like this should be fine. LGTM.

PR #931, opened 2022-01-04
fix: show console.log output on Windows

### LINKED ISSUE #699: windows console.log not showing until program ended
On windows 10, using fiddler 0.23.1, console.log's in the main process don't show in the console output until you end the program.

dsanders11:
Fixes #699.

On Windows, `console.log("Hello World")` will send `Hello World\n` to `pushOutput`, so it was being buffered indefinitely, since it would never include `\r\n` which is when the buffer would be outputted.

This PR should handle both `\r\n` and `\n` line endings on Windows.

cc @felixrieseberg

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -619,7 +619,7 @@ export class AppState {
    * @returns {void}
    */
   @action public flushOutput(): void {
-    this.pushOutput('\r\n', { bypassBuffer: false });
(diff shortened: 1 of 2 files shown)


erickzhao:
Tagging @codebytere as this is a follow-up to #676

codebytere (approved):

ckerr (approved):

PR #933, opened 2022-01-05
fix: remove race condition when stopping a fiddle

dsanders11:
Fixes a race condition where stopping a Fiddle and starting it again immediately would kill the new child process.


ckerr (commented):
For sure this is better, but both versions are a little brittle about relying on the pid. It's unlikely to happen so quickly after this child is killed, but pids can be reused.

What if we were to remove the pid code altogether and look for subprocess.exitCode to be non-null?

dsanders11:
@ckerr, good suggestion. That also lets the `is-running` dependency be dropped. I've updated the PR and title to match.

erickzhao (approved):

ckerr (approved):

Issue #942, opened 2022-01-14
New Fiddle (Cmd+N) doesn't clear modules


nornagon:
I'd expect the module list to be cleared when making a new Fiddle.

PR #943, opened 2022-01-14
fix: clear installed modules on new fiddle

codebytere:
Closes https://github.com/electron/fiddle/issues/942.

Also cleans up some other cruft.

### DIFF
--- a/src/renderer/app.tsx
+++ b/src/renderer/app.tsx
@@ -89,7 +89,7 @@ export class App {
   ): Promise<EditorValues> {
     const values = this.state.editorMosaic.values();
 
-    if (options && options.include !== false) {
+    if (options) {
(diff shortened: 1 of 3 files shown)


src/renderer/app.tsx
```diff
@@ -89,9 +89,7 @@ export class App {
   ): Promise<EditorValues> {
     const values = this.state.editorMosaic.values();
 
-    if (options && options.include !== false) {
-      values[PACKAGE_NAME] = await getPackageJson(this.state, options);
-    }
+    values[PACKAGE_NAME] = await getPackageJson(this.state, options);
```
codebytere, line 92: `options.include` is never passed anywhere, so this was always true.

erickzhao (approved):

PR #944, opened 2022-01-14
chore: update docstrings

dsanders11:
Did a pass over the whole codebase to update docstrings. Could use a close review, I went a bit cross-eyed by the end and might have made a mistake or two.


erickzhao (approved):
LGTM

Issue #896, opened 2021-11-10
I want to customize the author in package.json


exceedsystem:
I want to set the author of package.json that I publish to Gist to an arbitrary user name instead of the login user name of computer.

exceedsystem:
@BlackHole1 

Thank you!

PR #945, opened 2022-01-16
feat: support package.json author field

BlackHole1:
close: #896

Preview Image:
图片

Preview Gist URL:
https://gist.github.com/BlackHole1/e7f428532b78e59096ac73f617b29bab#file-package-json-L8

cc: @erickzhao

### DIFF
--- a/src/renderer/components/settings-general-package-author.tsx
+++ b/src/renderer/components/settings-general-package-author.tsx
@@ -49,9 +49,8 @@ export class PackageAuthorSettings extends React.Component<
   }
 
   public render() {
-    const { packageAuthor } = this.props.appState;
-
-    const packageAuthorLabel = 'Set package.json author field';
+    const packageAuthorLabel =
+      'Set the package.json author field for your exported Fiddle projects.';
 
     return (
       <div>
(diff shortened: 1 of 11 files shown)


src/renderer/components/settings-general-package-author.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    const packageAuthorLabel = 'Set package.json author field';
+
+    return (
+      <div>
+        <h4>Package Author</h4>
+        <Callout>
+          <FormGroup label={packageAuthorLabel}>
+            <InputGroup
+              placeholder={packageAuthor}
```
erickzhao, line 62: Since `this.props.appState.packageAuthor` is tied to `this.state.value`, it seems like the placeholder would never show?
```suggestion
```

tests/renderer/components/__snapshots__/settings-general-package-author-spec.tsx.snap
```diff
@@ -0,0 +1,20 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`PackageAuthorSettings component renders 1`] = `
+<div>
+  <h4>
+    Package Author
+  </h4>
+  <Blueprint3.Callout>
+    <Blueprint3.FormGroup
+      label="Set package.json author field"
```
erickzhao, line 10:
```suggestion
      label="Set the package.json author field for your exported Fiddle projects."
```

erickzhao (changes requested):
Some small feedback details

erickzhao (approved):

Issue #958, opened 2022-01-26
Expand npm package features


erickzhao:
Namely:
* Version selection
* Auto-complete

PR #959, opened 2022-01-27
feat: add npm registry autocomplete to package manager

erickzhao:
ref https://github.com/electron/fiddle/issues/958

Adds autofill with Algolia's `npm-search` client. Allows the end-user to select versions from the package manager and get suggestions from the NPM registry.

## Acceptance criteria

- [x] Can add a package and run a Fiddle containing the package.
- [x] Defaults to the latest version of the package.
- [x] Can change versions between runs with a select input.
- [x] Has backward compatibility with older `package.json` formats (we used to pass each version as `*`).
- [x] Exporting the Fiddle to a Gist or local folder preserves package information.
- [x] Importing the Fiddle to a Gist or local folder preserves package information.
- [x] Loads the version list for imported gists


BlackHole1 (approved):
LGTM

BlackHole1:
So cool!


https://user-images.githubusercontent.com/8198408/165210474-1cd63d76-8bd4-427d-a433-17a474ff5336.mp4

src/renderer/npm-search.tsx
```diff
@@ -0,0 +1,47 @@
...
+  };
+}
+
+class NPMSearch {
+  private index: SearchIndex;
+  private searchCache: Map<string, SearchResponse<AlgoliaHit>>;
+  constructor() {
+    const client = algoliasearch(
+      'OFCNCOG2CU',
+      '4efa2042cf4dba11be6e96e5c394e1a4',
```
BlackHole1, line 22: I am concerned that this could be abused by others, is it possible to confirm that this key has very low permissions?
> BTW. is this managed by the electron account?

erickzhao: > is it possible to confirm that this key has very low permissions?

From my discussions with Algolia support, this is a read-only API key that should be safe to put into our client.

> is this managed by the electron account?

The index itself is managed by Algolia: https://github.com/algolia/npm-search

BlackHole1: Okey

PR #961, opened 2022-02-02
feat: use WCO on macOS

clavin:
Merges the title bar into the toolbar on macOS, creating an integrated toolbar and title bar.

Screen Shot 2022-02-02 at 11 43 23 AM
 
🗒 **Note**: Currently, some toolbar buttons sort of go off screen if you make the window too small, but I believe this already happens in the current version of Fiddle so perhaps a separate issue should be filed for that instead of addressing it here.


erickzhao (changes requested):
Looks like unit tests are failing since we changed the return value of `getMainWindowsOptions()`.

ckerr:
@erickzhao could you take another look, CI is green now

erickzhao (approved):

Issue #962, opened 2022-02-08
Fiddle falsely assumes there are always dependencies


codebytere:
As of now, fiddle assumes [dependencies are non-null](https://github.com/electron/fiddle/blob/a593bb6ef05af9915ac6cf6ab27128a4f3a6d908/src/renderer/remote-loader.ts#L117-L120) and errors in the case where there are no dependencies or there are only devDependencies.

Reproducible by loading https://gist.github.com/t57ser/ce254e6e2e828949809bf872ed4168b6 and observing the following error:

<img width="407" alt="Screen Shot 2022-02-08 at 3 32 22 PM" src="https://user-images.githubusercontent.com/2036040/153007970-37702bf2-b3e0-40e4-9eff-38b4acff0427.png">

PR #963, opened 2022-02-08
fix: loading gists with deps/devDeps

codebytere:
Closes https://github.com/electron/fiddle/issues/962.

Fixes an issue whereby Fiddle assumes [dependencies are non-null](https://github.com/electron/fiddle/blob/a593bb6ef05af9915ac6cf6ab27128a4f3a6d908/src/renderer/remote-loader.ts#L117-L120) and errors in the case where there are no dependencies or there are only devDependencies.

### DIFF
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -115,8 +115,18 @@ export class RemoteLoader {
 
       for (const [id, data] of Object.entries(gist.data.files)) {
         if (id === PACKAGE_NAME) {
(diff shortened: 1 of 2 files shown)


ckerr (approved):
The devDependencies inclusion seems unrelated to this PR, arguably should be a separate fix, but both fixes look right to me. :tada:

ckerr:
(using admin permissions to merge: 'semantic pull request' CI bot seems to be hiccuping so we're not all-green, but that's unrelated to the PR)

Issue #966, opened 2022-02-10
Loading a Fiddle from disk doesn't take the Electron version into account


codebytere:
When a Fiddle is loaded from disk, it should take into account the Electron version specified in `package.json` and default the runner version to it.

Issue #965, opened 2022-02-10
Loading a Fiddle from disk doesn't take modules into account


codebytere:
If a Fiddle is loaded from the user's local disk, Fiddle should also load the modules present in `package.json`, as is currently done [when loading Fiddle Gists](https://github.com/electron/fiddle/blob/4ddb2c1f1acb2eb2609c43d2362de6741a31d099/src/renderer/remote-loader.ts#L129)

PR #969, opened 2022-02-11
feat: local fiddles respect `package.json`

codebytere:
Closes https://github.com/electron/fiddle/issues/966.
Closes https://github.com/electron/fiddle/issues/965

Follow up to https://github.com/electron/fiddle/pull/967 - ensure that local fiddles loaded from disk also respect modules and the version of Electron optionally specified in `package.json`

### DIFF
--- a/src/renderer/file-manager.ts
+++ b/src/renderer/file-manager.ts
@@ -1,5 +1,6 @@
 import * as fs from 'fs-extra';
 import * as path from 'path';
+import semver from 'semver';
 
 import { Files, FileTransform, PACKAGE_NAME } from '../interfaces';
 import { IpcEvents } from '../ipc-events';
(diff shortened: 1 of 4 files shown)


src/renderer/file-manager.ts
```diff
@@ -68,7 +69,39 @@ export class FileManager {
...
+      if (name === PACKAGE_NAME) {
+        const { remoteLoader } = window.ElectronFiddle.app;
+        const { dependencies, devDependencies } = JSON.parse(value);
+        const deps: Record<string, string> = {
+          ...dependencies,
+          ...devDependencies,
+        };
+
+        if (deps.electron) {
+          // Strip off semver range prefixes.
```
ckerr, line 85: The code is probably fine, but when I'm reading through this first time, I hit this paragraph and didn't understand it after five seconds, so I'm going to say the comment is not doing its job of explaining what's going on. :smile: 

Please change it to something like

```js
// Strip off semver range prefixes, e.g.
// before -> after
```

Where `before -> after` is a demo of what to expect

src/renderer/file-manager.ts
```diff
@@ -68,7 +69,39 @@ export class FileManager {
...
+          remoteLoader.setElectronVersion(version);
+
+          // We want to include all dependencies except Electron.
+          delete deps.electron;
+        }
+
+        this.appState.modules = new Map(Object.entries(deps));
+        continue;
+      }
+
```
ckerr, line 104: Should this new code go in `app.replaceFiddle()` somewhere so that other codepaths that replace the fiddle will also get these Electron version safety checks?

tests/renderer/file-manager-spec.ts
```diff
@@ -63,6 +63,50 @@ describe('FileManager', () => {
...
+      expect(app.replaceFiddle).toHaveBeenCalledWith(editorValues, {
+        filePath,
+      });
+    });
+
+    it('correctly adds modules specified in package.json', async () => {
+      const pj = {
+        main: 'main.js',
+        dependencies: {
+          'meaning-of-life': '*',
```
ckerr, line 92: :high_brightness:

src/renderer/file-manager.ts
codebytere: Would you potentially be ok with taking that as a follow-up? i'd want to think about the best way to do that but agree we should extract this to some extent 🤔

ckerr: I wouldn't block this PR on it but I would want to ensure we commit to a followup. Otherwise we'll have multiple codepaths that do almost-but-not-quite the same thing, by accident rather than by design, which is a pain to debug & maintain.

ckerr (approved):

Issue #949, opened 2022-01-20
Clicking on filenames in the sidebar should do something


nornagon:
Currently, clicking anywhere here seems to have no effect:
![Screen Shot 2022-01-20 at 3 11 52 PM](https://user-images.githubusercontent.com/172800/150436862-ebf55a89-c9e7-4632-bc04-078f88f7c536.png)

codebytere:
@nornagon instinctively, what behavior would you _want_ to see happen there? for instance, would you want it to focus the associated editor?

nornagon:
In particular, if the editor isn't visible, it should become visible. Focus makes sense too.

PR #979, opened 2022-02-18
feat: Click the sidebar to activate the current file

Spencer17x:
Closes https://github.com/electron/fiddle/issues/949

### DIFF
--- a/src/less/components/sidebar.less
+++ b/src/less/components/sidebar.less
@@ -16,3 +16,7 @@
 .bp3-tree-node-content-1 {
   padding-left: 0;
 }
+
+.pointer {
+  cursor: pointer;
+}
--- a/src/renderer/components/editor.tsx
+++ b/src/renderer/components/editor.tsx
@@ -67,6 +67,7 @@ export class Editor extends React.Component<EditorProps> {
     // Because you can't focus immediately after redisplay, you must wait until the mount is complete
     if (editorMosaic.focusedFile === id) {
       editor.focus();
+      this.props.setFocused(id);
     }
   }
 
--- a/src/renderer/components/editor.tsx
+++ b/src/renderer/components/editor.tsx
@@ -56,7 +56,7 @@ export class Editor extends React.Component<EditorProps> {
     const { appState, editorDidMount, id } = this.props;
     const { editorMosaic } = appState;
 
-    appState.editorMosaic.addEditor(id, editor);
+    editorMosaic.addEditor(id, editor);
 
     // And notify others
     if (editorDidMount) {
--- a/src/renderer/components/editor.tsx
+++ b/src/renderer/components/editor.tsx
@@ -61,6 +61,8 @@ export class Editor extends React.Component<EditorProps> {
     if (editorDidMount) {
       editorDidMount(editor);
     }
+
+    this.editor.focus();
   }
 
   /**
--- a/src/renderer/components/editor.tsx
+++ b/src/renderer/components/editor.tsx
@@ -54,6 +54,7 @@ export class Editor extends React.Component<EditorProps> {
    */
   public async editorDidMount(editor: MonacoType.editor.IStandaloneCodeEditor) {
     const { appState, editorDidMount, id } = this.props;
+    const { editorMosaic } = appState;
 
     appState.editorMosaic.addEditor(id, editor);
 
@@ -62,7 +63,11 @@ export class Editor extends React.Component<EditorProps> {
       editorDidMount(editor);
     }
 
-    this.editor.focus();
+    // Click file tree, if the file is hidden, focus it
+    // Because you can't focus immediately after redisplay, you must wait until the mount is complete
+    if (editorMosaic.focusedFile === id) {
+      editor.focus();
+    }
   }
 
   /**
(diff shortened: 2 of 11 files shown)


BlackHole1 (changes requested):
Can you add unit tests for this?

BlackHole1:
nit: It would be nice to add a style for the mouse

```css
cursor: pointer;
```

**before:**
image

**after:**
image2

BlackHole1:
I think the following logic might be better, what do you think?

When the file is hidden (the eyes on the right are not visible), clicking on the text (the text on the left) will make it visible again

When the file is visible (the eye on the right is visible), clicking on the text (the text on the left) will focus on the contents of this file.

image

Spencer17x:
In my opinion, is it better to change this interaction form to the same as IDEA? What do you think?

Spencer17x:
https://user-images.githubusercontent.com/32794413/154681865-4119a0be-5f55-4d0c-9893-c1cdfdcb85fe.mov

Like IDEA interaction: remove the display and hide icon of the sidebar, when the file is clicked, the sidebar only displays the active state of the file, and when double-clicked, the file is opened and the file of the sidebar is activated

Spencer17x:
ping @codebytere @erickzhao @nornagon

src/renderer/components/sidebar-file-tree.tsx
```diff
@@ -52,11 +52,13 @@ export class SidebarFileTree extends React.Component<
           presence !== EditorPresence.Hidden ? 'eye-open' : 'eye-off';
 
         return {
+          isSelected: currentEditorId === editorId,
           id: index,
           hasCaret: false,
           icon: 'document',
           label: (
             <ContextMenu2
+              onClick={() => this.toggleVisibility(editorId)}
```
codebytere, line 61: In keeping with @nornagon's comments from the issue, i agree that we should show and focus the editor if it's not visible. What this does, however, is makes it such that clicking once shows the editor, and another click hides it. I don't think that's intuitive behavior nor behavior users would want. We should _only_ show, never hide here.

codebytere (changes requested):
Good work so far! There are some behavioral changes i'd like to see but i think we can hopefully get this merged soon.

src/renderer/components/sidebar-file-tree.tsx
BlackHole1: According to what you say:
1. If not visible, display and focus
2. if visible, what is the behaviour? Is it just focus?

ckerr: Yes. When clicked, ensure the clicked item is visible and has focus.

Spencer17x:
I adjusted it, please review it @BlackHole1

src/less/components/sidebar.less
```diff
@@ -16,3 +16,8 @@
 .bp3-tree-node-content-1 {
   padding-left: 0;
 }
+
+
+.pointer {
```
BlackHole1, line 21:
```suggestion

.pointer {
```

BlackHole1 (changes requested):
need `rebase` main

src/renderer/components/editor.tsx
```diff
@@ -61,6 +61,8 @@ export class Editor extends React.Component<EditorProps> {
     if (editorDidMount) {
       editorDidMount(editor);
     }
+
+    this.editor.focus();
```
BlackHole1, line 65: Remove. We don't need this feature

BlackHole1 (changes requested):
Need update `Snapshot`.

Please execute `yarn run test -- -u` in project root path.

src/renderer/components/editor.tsx
Spencer17x: Ok, I have made the modification

BlackHole1 (approved):
LGTM

cc: @codebytere

tests/renderer/components/__snapshots__/commands-spec.tsx.snap
```diff
@@ -28,6 +28,11 @@ exports[`Commands component renders 1`] = `
       />
     </Blueprint3.ControlGroup>
   </div>
+  <div
+    className="title"
+  >
+    Electron Fiddle
+  </div>
```
BlackHole1, line 35:
```suggestion
```

Revert.

see: https://github.com/electron/fiddle/blob/28ec7c77d7276f9b8e8ba28c700ff39f7e35ba45/src/renderer/components/commands.tsx#L60-L63

CI machine is `linux`, not `macOS`. This will cause the unit tests in the CI to fail

BlackHole1 (changes requested):

tests/renderer/components/__snapshots__/commands-spec.tsx.snap
Spencer17x: Ok, I've made some changes

Spencer17x:
@coveralls I have modified it, please check it

erickzhao (commented):
Playing around with this a bit, it would be nice if focusing a specific editor window would also change the focused editor name in the sidebar.

Spencer17x:
@erickzhao I have made some modifications, please review

erickzhao (approved):

codebytere (approved):

Issue #978, opened 2022-02-18
Add Node Typings to Intellisense


dsanders11:
Currently Fiddle doesn't have any intellisense for Node, so hovering over something like `process` or `setImmediate` won't give you anything useful. In theory it shouldn't be too hard to wire this in.

cc @ckerr

PR #988, opened 2022-03-02
feat: add Intellisense for Node.js

codebytere:
Closes https://github.com/electron/fiddle/issues/978


src/renderer/electron-types.ts
```diff
@@ -19,41 +24,84 @@ export class ElectronTypes {
...
-      this.setTypesFromFile(file, ver);
+      const file = this.getCacheFile(ver.version);
+      await ElectronTypes.ensureElectronVersionIsCachedAt(version, file);
+      this.setTypesFromFile(file, ver.version);
     }
   }
 
+  public async setNodeTypes(version: string): Promise<void> {
+    // Get the Node.js version corresponding to the current Electron version.
+    const v = releases.find((release: any) => {
```
bnb, line 66: might be worth typing this `release` prop

src/renderer/electron-types.ts
```diff
@@ -19,41 +24,84 @@ export class ElectronTypes {
...
   public async setVersion(ver?: RunnableVersion): Promise<void> {
     this.clear();
 
     if (!ver) return;
+
+    await this.setElectronTypes(ver);
+    await this.setNodeTypes(ver.version);
+  }
+
+  public async setElectronTypes(ver: RunnableVersion): Promise<void> {
```
ckerr, line 40:
```suggestion
  public async setElectronTypes(ver: RunnableVersion): Promise<void> {
```

Looks like this is only called from setVersion(); this could be private

src/renderer/electron-types.ts
```diff
@@ -19,41 +24,84 @@ export class ElectronTypes {
...
     const { localPath: dir, source, version } = ver;
 
+    // If it's a local development version, pull Electron types from out directory.
     if (dir) {
       const file = path.join(dir, 'gen/electron/tsc/typings', ELECTRON_DTS);
-      this.setTypesFromFile(file, ver);
+      this.setTypesFromFile(file, ver.version);
```
ckerr, line 46: _nit_ `version` already extracted on line 41

```suggestion
      this.setTypesFromFile(file, version);
```

Same thing below on 49, 58, 60

ckerr (changes requested):
Please fix tests

dsanders11:
The latest CI failure on this PR looks to have been due to the Node version in the CI runner. I bumped those to v16 in #1117, so a rebase on `main` should get past that error. 👍

codebytere:
done @dsanders11, thanks! this should be ready for some more review :)

package.json
```diff
@@ -51,13 +51,15 @@
     "extract-zip": "^2.0.1",
     "fs-extra": "^9.1.0",
     "getos": "^3.2.1",
+    "latest-version": "^6.0.0",
```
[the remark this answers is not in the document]
codebytere: oops good catch - i'd initially used this and then ditched it later.

src/renderer/electron-types.ts
```diff
@@ -20,41 +25,82 @@ export class ElectronTypes {
...
     this.dispose();
+
     try {
-      console.log(`Updating Monaco with "${ELECTRON_DTS}@${ver.version}"`);
+      const files = await readdir(dir);
+      for (const file of files) {
+        if (file.endsWith('.d.ts')) {
+          const { base: filename } = path.parse(file);
+          console.log(`Updating Monaco with "${filename}@${version}"`);
+          this.disposable = this.monaco.languages.typescript.javascriptDefaults.addExtraLib(
```
dsanders11, line 90: Should `disposable` be changed to `disposables`, an array of `Disposable`? It looks like now there can be multiple, and this might be leaking disposables added in a previous iteration of this for loop. So probably need to make it an array and change `dispose()` loop through it disposing of them.

dsanders11 (commented):
Left a functional comment, and this need to be rebased now that #1109 has been merged.

src/renderer/electron-types.ts
codebytere: done!

dsanders11 (approved):

ckerr (approved):
This is pretty great!

Issue #977, opened 2022-02-17
Codesigning not working for windows


milenkovic:
It seems that code signing is not working for windows:
- https://github.com/electron/fiddle/runs/4875090891?check_suite_focus=true#step:7:38

Also, not sure if this file `cert.p12` is generated from any place in the code.

BlackHole1:
see: https://www.virustotal.com/gui/file/f72ce5d38be3ddc9ee0a6b36c732e683a4c2f82dda77f9f2c3df924e798df6e0/details

<img width="507" alt="image" src="https://user-images.githubusercontent.com/8198408/156796089-42c650e3-0106-4033-9360-9a5f5695375f.png">


The file has been signed and the warning here has no effect.

PR #990, opened 2022-03-04
chore(forge): remove unnecessary warn log

BlackHole1:
Closed: #977


BlackHole1:
BTW, the previous code actually belongs to: `dead code`. They are of no use and cause incorrect log output. 👀

erickzhao (approved):

PR #1001, opened 2022-03-07
chore(ci): add yarn install retry logic

BlackHole1:
See: https://github.com/electron/fiddle/runs/5438013000?check_suite_focus=true#step:6:19


erickzhao (approved):
Unfortunate that there's not a cleaner solution but this looks like it works!

PR #1009, opened 2022-03-14
fix: don't use GitHub access token when loading from docs

### LINKED ISSUE #1008: Can't Load Fiddle From Docs When Logged Into GitHub
I'm logged into GitHub on Fiddle with a personal access token so I can publish gists. When I try to load a fiddle from the docs I get:

```
Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.
```

dsanders11:
Fixes #1008.

This could also be fixed by authorizing the access token with the Electron org, but that seems like a bad idea to me, I'm loading publicly visible resources. I want as little surface area as possible as far as Electron org authorization goes.

There is a comment that non-auth API access gets worse rate-limiting, but I think this is a reasonable tradeoff, given the above.

### DIFF
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -59,7 +59,7 @@ export class RemoteLoader {
     path: string,
   ): Promise<boolean> {
     try {
-      const octo = await getOctokit(this.appState);
+      const octo = await getOctokit();
       const folder = await octo.repos.getContents({
         owner: ELECTRON_REPO,
         repo: ELECTRON_ORG,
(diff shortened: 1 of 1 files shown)


MarshallOfSound (changes requested):
This should be used as a fallback instead of a replacement, unauthed requests get a dramatically lower rate limit that is hit almost instantly in offices and very quickly even by yourself.

Like try one of it fails due to rate limit try the other

dsanders11:
Didn't realize the dual-purpose of also fixing rate limits. Looking at the code, I don't think it's worth trying to implement a fallback since this is a niche case. It looks like authorizing the PAT with the Electron org is acceptable enough since the token is scoped to just gist stuff. I'll go ahead and close this.

PR #1012, opened 2022-03-14
fix: restore commands padding on Windows/Linux

### LINKED ISSUE #985: toolbar looks strange on Linux
Introduced by #961. 

Before this PR, the look on Linux was:

old-look

After the PR, the look is:

new-look

Ayman161803:
This PR Fixes #985. Here's a pic of how the toolbar looks after the changes that have been made.

Screenshot from 2022-03-14 19-20-52

### DIFF
--- a/src/less/components/commands.less
+++ b/src/less/components/commands.less
@@ -6,8 +6,6 @@ header {
   -webkit-app-region: drag;
 }
 
-
-
 .commands.is-mac {
   padding:
     10px
--- a/src/less/components/commands.less
+++ b/src/less/components/commands.less
@@ -8,7 +8,7 @@ header {
 
 
 
-header .is-mac{
+.commands.is-mac {
   padding:
     10px
     calc(100vw - (env(titlebar-area-x, 0) + env(titlebar-area-width, 0)))
--- a/src/less/components/commands.less
+++ b/src/less/components/commands.less
@@ -17,7 +17,6 @@ header {
 }
 
 
-
 .commands {
   display: flex;
   justify-content: space-between;
--- a/src/less/components/commands.less
+++ b/src/less/components/commands.less
@@ -11,10 +11,7 @@ header {
   justify-content: space-between;
   height: 50px;
   padding:
-    10px
-    calc(100vw - (env(titlebar-area-x, 0) + env(titlebar-area-width, 0)))
-    10px
-    env(titlebar-area-x, 0);
+    10px;
   box-sizing: border-box;
   font-family: @fonts-common;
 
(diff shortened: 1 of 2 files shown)


Ayman161803:
PR as a fix to #985

src/less/components/commands.less
```diff
@@ -11,10 +11,7 @@ header {
   justify-content: space-between;
   height: 50px;
   padding:
-    10px
-    calc(100vw - (env(titlebar-area-x, 0) + env(titlebar-area-width, 0)))
-    10px
-    env(titlebar-area-x, 0);
+    10px;
```
erickzhao, line 14: I think because this is specifically for Windows/Linux styles, we'd want to make this styling different depending on OS.

 We can do this conditionally by applying a CSS class (e.g. `is-mac`) in the React code via `process.platform === 'darwin'` and applying different paddings depending on that class.

erickzhao (changes requested):

src/less/components/commands.less
Ayman161803: Hey there! I have made recommended changes to the PR i.e. a CSS class that gets applied only for Mac OS. FYI  ```header .is-mac ``` selector in ```commands.less``` makes sure the padding specified with ```.is-mac``` styles gets higher specificity.

erickzhao (dismissed):
1 small change again. Thank you for this!

Ayman161803:
Hey I have committed recommended changes to this PR. Thank you for suggesting the same.

src/less/components/commands.less
```diff
@@ -6,15 +6,24 @@ header {
...
+
+
+.commands.is-mac {
   padding:
     10px
     calc(100vw - (env(titlebar-area-x, 0) + env(titlebar-area-width, 0)))
     10px
     env(titlebar-area-x, 0);
+}
+
```
dsanders11, line 18:
```suggestion
```

src/less/components/commands.less
```diff
@@ -6,15 +6,24 @@ header {
...
+
```
dsanders11, line 19:
```suggestion
```

dsanders11 (approved):
Made some suggestions to remove extra blank lines (for some reason I couldn't do it as a single suggestion), but otherwise LGTM. I've confirmed the UI looks correct on both Windows and Linux with this change, thanks @Ayman161803!

Ayman161803:
Committed suggested recommendations

dsanders11:
Thanks for the contribution @Ayman161803!

PR #1013, opened 2022-03-14
fix(show-me/webframe): remove unnecessary link to renderer.js file

aryanshridhar:
Added the background in commit description.

Should be a short one to review @erickzhao :)


aryanshridhar:
FYI @thewheat.

PR #1015, opened 2022-03-16
refactor(src/renderer): Removed process.platform usage from src/renderer 

Ayman161803:
I have been trying to understand ```Enable Context Isolation in Electron Fiddle``` for a while. I came across the use of Node's ```process.platform``` within  ```src/renderer/components/commands.tsx``` while attempting to solve  issue #985 through PR #1012. I have tried to refactor code so that ```process.platform``` is no longer called from ```src/renderer/```. Since the ```process.platform```'s return value would be constant throughout the entire process and has  been accessed multiple times in ```state.ts```, ```npm.ts``` and ```command.ts```, I have initialised ```window.ElectronFiddle.processPlatform``` using ```GET_PROCESS_PLATFORM``` event in ```preload/preload.ts``` which has then been accessed instead of using ```process.platform```. This was a part of me trying to understand  ```Context Isolation``` and the fiddle repo in general so that I write a better proposal for ```Enable Context Isolation in Electron Fiddle```.


Ayman161803:
Hey! I went through the IPC docs again to better understand ```Context Isolation```  and it looks like the changes that I have made ```will not contextually isolate``` the renderer's process from the preload script :/ . The reason being, even after the changes that I have made, the ```renderer will still share the window object with the preload script``` in contrast to what should be case as mentioned in the docs below. I will be closing this PR for the same reasons. It was great to improve my understanding in these concepts. Cheers! 
Screenshot from 2022-03-19 01-01-14

PR #1018, opened 2022-03-18
fix: focus the window when loading fiddle with protocol

### LINKED ISSUE #1007: Fiddle Doesn't Come to Foreground When Loading Fiddle From Browser
Currently when opening a fiddle from the web browser, if Fiddle is already open, it will show you the "Are You Sure?" dialog, but it won't bring Fiddle to the foreground, so you wouldn't know it's doing so. On macOS it correctly comes to the foreground.

akgupta0777:
Fixes #1007.
 
Previously when loading fiddle and if a fiddle is already loaded then the window doesn't come to foreground. This behaviour is seen on windows. This PR fixes that issue.

**Quick Demo :**

https://user-images.githubusercontent.com/51379307/159056900-837972f6-0373-426a-977f-4dea867ffac1.mp4

I am new here and my solution may not be the best and efficient one,😅 So I request you to provide the valuable feedback to me.

**What I did**
I used the `BrowserWindow` API and called its show method whenever any potential protocol launch occurs.

### DIFF
--- a/src/main/protocol.ts
+++ b/src/main/protocol.ts
@@ -6,6 +6,7 @@ import * as nodeUrl from 'url';
 import { IpcEvents } from '../ipc-events';
 import { isDevMode } from '../utils/devmode';
 import { ipcMainManager } from './ipc';
+import { getOrCreateMainWindow } from './windows';
 
 const PROTOCOL = 'electron-fiddle';
 const squirrelPath = path.resolve(
@@ -66,6 +67,8 @@ const handlePotentialProtocolLaunch = (url: string) => {
     default:
       return;
   }
+  const win = getOrCreateMainWindow();
+  win.show();
 };
 
 const isProtocolString = (arg: string) => arg.startsWith(`${PROTOCOL}://`);
--- a/src/main/protocol.ts
+++ b/src/main/protocol.ts
@@ -67,8 +67,7 @@ const handlePotentialProtocolLaunch = (url: string) => {
     default:
       return;
   }
-  const win = getOrCreateMainWindow();
-  win.show();
+  getOrCreateMainWindow().focus();
 };
 
 const isProtocolString = (arg: string) => arg.startsWith(`${PROTOCOL}://`);
--- a/tests/main/protocol-spec.ts
+++ b/tests/main/protocol-spec.ts
@@ -180,12 +180,13 @@ describe('protocol', () => {
 
     it('focuses window when loading a fiddle', () => {
       listenForProtocolHandler();
-      getOrCreateMainWindow().blur();
+
+      const mainWindow = getOrCreateMainWindow();
       const handler = (app.on as any).mock.calls[0][1];
 
       handler({}, 'electron-fiddle://electron/4.0.0/test/path');
 
-      expect(getOrCreateMainWindow().focus).toHaveBeenCalled();
+      expect(mainWindow.focus).toHaveBeenCalled();
     });
   });
 });
--- a/tests/main/protocol-spec.ts
+++ b/tests/main/protocol-spec.ts
@@ -4,6 +4,7 @@
 
 import { app } from 'electron';
 import * as fs from 'fs';
+import { getOrCreateMainWindow } from '../../src/main/windows';
 
 import { IpcEvents } from '../../src/ipc-events';
 import { ipcMainManager } from '../../src/main/ipc';
@@ -175,6 +176,16 @@ describe('protocol', () => {
       handler({}, 'electron-fiddle://');
 
       expect(ipcMainManager.send).toHaveBeenCalledTimes(0);
+    });
+
+    it('focuses window when loading a fiddle', () => {
+      listenForProtocolHandler();
+      getOrCreateMainWindow().blur();
+      const handler = (app.on as any).mock.calls[0][1];
+
+      handler({}, 'electron-fiddle://electron/4.0.0/test/path');
+
+      expect(getOrCreateMainWindow().focus).toHaveBeenCalled();
     });
   });
 });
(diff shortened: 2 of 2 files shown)


src/main/protocol.ts
```diff
@@ -66,6 +67,8 @@ const handlePotentialProtocolLaunch = (url: string) => {
     default:
       return;
   }
+  const win = getOrCreateMainWindow();
+  win.show();
```
dsanders11, line 71:
```suggestion

  // Focus the window in case it is in the background
  getOrCreateMainWindow.focus();
```

dsanders11 (changes requested):
Thanks for tracking this down and the fix @akgupta0777!

I've edited the PR description to use "Fixes #<issue>" so that GitHub will automatically close the issue when this gets merged. 🙂 

Small suggested change to use `BrowserWindow.focus()` since I don't think `show()` is necessary here, and added a comment.

Could you see if adding a test for this is feasible? Test would go in `main/tests/main/protocol-spec.ts`. You can use `getOrCreateMainWindow` to get the window at the start of the test, use `.blur()` to unfocus it, assert that `isFocused()` is false, then trigger the protocol load and expect `isFocused()` to be true after it is done. Since window changes aren't necessarily instant you may want to use the `waitFor` test util to wait for it to become focused, see [this code](https://github.com/electron/fiddle/blob/1ede592ae5a572f37776aeab2a14409c32833719/tests/renderer/binary-spec.ts#L109-L111) for an example usage.

src/main/protocol.ts
akgupta0777: @dsanders11 Thanks for the feedback. Yes the focus method is more suitable here and also more readable.

I will sure write tests for this change.

dsanders11: So I double checked and we actually [mock out `BrowserWindow`](https://github.com/electron/fiddle/blob/main/tests/mocks/browser-window.ts) for tests, so the test can just confirm that `focus` is called. 🙂

dsanders11:
One more thing, the commit message isn't clear what's being fixed. Could you change it to something like "fix: focus the window when loading fiddle with protocol"? Thanks!

akgupta0777:
@dsanders11 I think the test can't able to access the main window.

I am doing this
```
it('focuses the window when loading fiddle', async () => {
      listenForProtocolHandler();
      getOrCreateMainWindow();
      browserWindows[0]?.blur();  // applying blur

      const handler = (app.on as any).mock.calls[0][1];
      const mockWindow = new BrowserWindowMock();

      handler({}, 'electron-fiddle://electron/v17.1.2/docs/fiddles/ipc/pattern-3'); // triggering protocol launch
      await waitFor(() => browserWindows[0]?.isFocused() == true) // waiting till the browserWindow gets into focus
      
       expect(mockWindow.focus).toHaveBeenCalled(); // checking focus has been called or not.
    });
  });
```

The above test always fails with timeout error: 2000ms thrown by `waitFor()`. I even increased the timeout to 30 seconds to test but it still doesn't able to check the isFocused on browser window.

akgupta0777:
So I have added test with mocking getOrCreateMainWindow() and checking if its focus method is called or not after blur. 

I cannot able to use `BrowserWindowMock` as `getOrCreateMainWindow` returns window which doesn't have isFocus or any other thing. I wonder why ?. I console logged the `getOrCreateMainWindow()` to see what functions I get and I only see focus , blur etc but not other `BrowserWindow API` functions.

BlackHole1 (approved):
LGTM

BlackHole1:
> So I have added test with mocking getOrCreateMainWindow() and checking if its focus method is called or not after blur.
> 
> I cannot able to use `BrowserWindowMock` as `getOrCreateMainWindow` returns window which doesn't have isFocus or any other thing. I wonder why ?. I console logged the `getOrCreateMainWindow()` to see what functions I get and I only see focus , blur etc but not other `BrowserWindow API` functions.

Hi @akgupta0777 👋. This is because the unit tests in fiddle don't actually start an electron application. Instead, we mock most of the electron api to test it.

See:
* https://github.com/electron/fiddle/blob/f6ff9ea314ac7fdb0f349d31d1855ed6321ca9d5/tests/mocks/electron.ts#L155
* https://github.com/electron/fiddle/blob/f6ff9ea314ac7fdb0f349d31d1855ed6321ca9d5/tests/mocks/browser-window.ts#L4-L15

> isFocus => isFocused

akgupta0777:
> > So I have added test with mocking getOrCreateMainWindow() and checking if its focus method is called or not after blur.
> > I cannot able to use `BrowserWindowMock` as `getOrCreateMainWindow` returns window which doesn't have isFocus or any other thing. I wonder why ?. I console logged the `getOrCreateMainWindow()` to see what functions I get and I only see focus , blur etc but not other `BrowserWindow API` functions.
> 
> Hi @akgupta0777 👋. This is because the unit tests in fiddle don't actually start an electron application. Instead, we mock most of the electron api to test it.
> 
> See:
> 
> * https://github.com/electron/fiddle/blob/f6ff9ea314ac7fdb0f349d31d1855ed6321ca9d5/tests/mocks/electron.ts#L155
> * https://github.com/electron/fiddle/blob/f6ff9ea314ac7fdb0f349d31d1855ed6321ca9d5/tests/mocks/browser-window.ts#L4-L15
> 
> > isFocus => isFocused

I have used that all, But the focus isn't instantaneous so i have to use waitFor but I don't know which condition to wait for.

BlackHole1:
That's fine for now. Because the api's here are all fake (jest.mock) and all synchronous. So there is no need to use waitFor

akgupta0777:
> Hey @akgupta0777 - I've made two small tweaks on your branch:
> 
> * Order of imports in the test (note that we group all the relative imports together)
> * Since the test uses mocks, no need to call `blur`
> 
> With those changes and CI is green, this LGTM. 🙂

LGTM 👍😃

PR #1019, opened 2022-03-19
fix: disabled download button for certain versions on darwin arm64

### LINKED ISSUE #1017: On Apple Silicon, Can't Download Pre-11.0 Versions
I'm gonna start by saying I do know why this is, Electron [didn't start shipping Apple Silicon builds until 11.0](https://www.electronjs.org/blog/apple-silicon).

That said, the current experience is you click the download button, it spins, then goes back to saying download. Behind the scenes it's hitting a 404 when trying to download, looking for the `arm64` version.

I see two options to fix this:
- Disable the download button for these versions with a tooltip on hover that explains why
- Download the x64 version for pre-11.0 and let Rosetta 2 do its thing

I think I favor the first option. For the most part folks don't need to go back that far, and will only need to do so less going forward. Might be some weird quirks with old builds running in Rosetta 2, so might not want to open that can of worms.

Ayman161803:
Fixes #1017 

I have disabled the download button for certain versions on darwin-arm64 as suggested by @dsanders11. I used the semver package to compare versions and added necessary stylings to ```commands.less```. I have tested the changes made with results as given below.
Screenshot from 2022-03-19 12-15-26

 The sample pic above is when condition ```process.platform==="linux" && process.arch==="x64" && semver.lt(item.version,"15.0.0")```  is applied to disable buttons


dsanders11:
Thanks for taking this up @Ayman161803. I was actually referring to the download buttons under preferences when I made that issue, but you're right, the ones in the version select also have the same problem. 😄

<img width="1107" alt="Fiddle Prefs -> Electron" src="https://user-images.githubusercontent.com/5820654/159114705-c2c78e80-b6b9-4d9c-8317-12f6e14770d4.png">

Unfortunately I think this needs to be changed a bit. The changes to `VersionSelect` will break other functionality, such as in `src/renderer/components/dialog-bisect.tsx`, which makes use of the `itemDisabled` prop. There's multiple possible reasons a version in `VersionSelect` might be disabled, so any change needs to keep that in mind. Maybe that change for `VersionSelect` should be entirely contained within `renderItem`.

dsanders11 (changes requested):
Let's also cover the download buttons in the Electron section in preferences.

Ayman161803:
Hey @dsanders11 ! I have made recommended changes to this PR i.e. ```shifted disable logic to renderItem``` and ```disabled appropriate download buttons on preference page``` as well. Please take the look at the same.

src/renderer/components/settings-electron.tsx
```diff
@@ -390,6 +391,25 @@ export class ElectronSettings extends React.Component<
           />
         </Tooltip>
       );
+    } else if (
+      process.platform === 'darwin' &&
+      process.arch === 'arm64' &&
+      semver.lt(version, '11.0.0')
```
BlackHole1, line 397: It might be better to put it in `utils/disable-download.ts`

```typescript
export function disableDownload(version: string) {
  return process.platform === 'darwin' && process.arch === 'arm64' && semver.lt(version, '11.0.0')
}
```

In the meantime, the relevant unit tests can be added:

```typescript
expect(disableDownload('10.0.0')).toBe(process.platform === 'darwin' && process.arch === 'arm64');
```

src/renderer/components/version-select.tsx
```diff
@@ -148,6 +150,34 @@ export const renderItem: ItemRenderer<RunnableVersion> = (
     return null;
   }
 
+  if (
+    process.platform === 'darwin' &&
+    process.arch === 'arm64' &&
+    semver.lt(item.version, '11.0.0')
```
BlackHole1, line 156: See above.

BlackHole1 (changes requested):

BlackHole1:
Before adding unit tests, you also need to update the `snapshots`.
Run `yarn run test -- -u` for automatic updates

src/renderer/components/settings-electron.tsx
Ayman161803: Committed suggested recommendations

Ayman161803:
Committed recommended changes. I am pretty new to writing unit tests. Please suggest changes if any. Thank you for your suggestions.

BlackHole1:
Component unit tests e.g.

```diff
@@ -0,0 +1,42 @@
diff --git a/tests/renderer/components/version-select-spec.tsx b/tests/renderer/components/version-select-spec.tsx
index 3a6bab6..16bdb50 100644
--- a/tests/renderer/components/version-select-spec.tsx
+++ b/tests/renderer/components/version-select-spec.tsx
@@ -14,12 +14,15 @@ import {
   renderItem,
   VersionSelect,
 } from '../../../src/renderer/components/version-select';
+import { disableDownload } from '../../../src/utils/disable-download';
 
 import { StateMock, VersionsMock } from '../../mocks/mocks';
 
 const { downloading, ready, unknown, unzipping } = VersionState;
 const { remote, local } = VersionSource;
 
+jest.mock('../../../src/utils/disable-download.ts');
+
 describe('VersionSelect component', () => {
   let store: StateMock;
 
@@ -85,6 +88,21 @@ describe('VersionSelect component', () => {
 
       expect(item).toBe(null);
     });
+
+    it('xxxxx', () => {
+      (disableDownload as any).mockReturnValueOnce(true);
+
+      const item = renderItem(mockVersion1, {
+        handleClick: () => ({}),
+        index: 0,
+        modifiers: { active: true, disabled: false, matchesPredicate: true },
+        query: '',
+      })!;
+
+      const ItemWrapper = shallow(item);
+
+      expect(ItemWrapper.find('.disabled-menu-tooltip')).toHaveLength(1);
+    });
   });
 
   describe('getItemLabel()', () => {

```

BlackHole1:
> Also, I tried updating snapshots as well but it doesn't look like they are getting updated.

Sorry, that was my mistake. It won't update here because the logic restricts os and arch as well. It's right not to update

src/utils/disable-download.ts
```diff
@@ -0,0 +1,9 @@
+import semver from 'semver';
+
+export function disableDownload(version: string) {
+  return (
+    process.platform === 'darwin' &&
+    process.arch === 'arm64' &&
+    semver.lt(version, '11.0.0')
+  );
```
BlackHole1, line 8: To improve the stability of subsequent unit tests, please change to:

```suggestion
  return (
    !process.env.JEST &&
    process.platform === 'darwin' &&
    process.arch === 'arm64' &&
    semver.lt(version, '11.0.0')
  );
```

See:
* https://github.com/electron/fiddle/pull/1022
* https://github.com/electron/fiddle/pull/992

BlackHole1 (approved):

src/utils/disable-download.ts
BlackHole1, line 8: To improve the stability of subsequent unit tests, please change to:

```suggestion
  return (
    !process.env.JEST &&
    process.platform === 'darwin' &&
    process.arch === 'arm64' &&
    semver.lt(version, '11.0.0')
  );
```

See:
* https://github.com/electron/fiddle/pull/1022
* https://github.com/electron/fiddle/pull/992

BlackHole1 (changes requested):

src/utils/disable-download.ts
Ayman161803: Enabling the above change would break the test that I had written for disableDownload. Hence, I have ```deleted my test for utils/disable-download.ts```. I have, however, ```added component tests to components versionSelect and ElectronSettings``` which use disableDownload function. The test that I had written for disableDownload is as follows and  would always fail (no matter what) for MacOS since ```!process.env.JEST``` will be returning false here.

Screenshot from 2022-03-22 20-00-48

Ayman161803: Sure. Will try to think of a solution as well.

BlackHole1: I am writing the code for this and I will submit it later

BlackHole1: Done.
See: https://github.com/electron/fiddle/pull/1019/commits/8b83a71c81afbef4ba92b27e378d93231d562154

> I have committed the code to the current PR. you need to synchronize locally (pull)

BlackHole1: Unit test pass: https://github.com/electron/fiddle/runs/5656821882?check_suite_focus=true#step:6:75 🎉

Ayman161803: Great!!

BlackHole1 (approved):

BlackHole1 (approved):
LGTM.

cc: @dsanders11

tests/renderer/components/settings-electron-spec.tsx
```diff
@@ -223,4 +226,43 @@ describe('ElectronSettings component', () => {
...
+
+      store.versions = { version: ver };
+      store.versionsToShow = [ver];
+
+      const wrapper = shallow(<ElectronSettings appState={store as any} />);
+
+      expect(wrapper.find('.disabled-version')).toHaveLength(1);
+    });
+
+    it('disables download button when return value is true', () => {
```
dsanders11, line 249: These two tests have the same name and are very similar. Does there need to be two here?

tests/utils/disable-download-spec.ts
```diff
@@ -0,0 +1,24 @@
+import { disableDownload } from '../../src/utils/disable-download';
+import { isJest } from '../../src/utils/is-jest';
+
+jest.mock('../../src/utils/is-jest');
+
+describe('disableDownload', () => {
+  // versions below 11.0.0 are not available for download on darwin arm64
+  it('returns false for downloadable versions and true otherwise', () => {
+    if (process.platform == 'darwin' && process.arch == 'arm64') {
+      (isJest as any).mockReturnValueOnce(false).mockReturnValueOnce(false);
```
dsanders11, line 10: This can just be `mockReturnValue(false)` which sets it to always return false.

It feels a bit strange to me to be mocking `isJest` to be false since the point of the helper is to detect that it's in Jest.

src/utils/disable-download.ts
```diff
@@ -0,0 +1,11 @@
+import semver from 'semver';
+import { isJest } from './is-jest';
+
+export function disableDownload(version: string) {
```
dsanders11, line 4: Let's add a JSDoc for this.

src/utils/disable-download.ts
```diff
@@ -0,0 +1,11 @@
...
+  return (
```
dsanders11, line 5: Please add a comment explaining why download is disabled in this situation.

dsanders11 (changes requested):

tests/utils/disable-download-spec.ts
BlackHole1: I think I've come up with a better solution. I will make the changes recently. 😄

tests/utils/disable-download-spec.ts
```diff
@@ -0,0 +1,24 @@
+import { disableDownload } from '../../src/utils/disable-download';
+import { isJest } from '../../src/utils/is-jest';
+
+jest.mock('../../src/utils/is-jest');
+
+describe('disableDownload', () => {
+  // versions below 11.0.0 are not available for download on darwin arm64
+  it('returns false for downloadable versions and true otherwise', () => {
+    if (process.platform == 'darwin' && process.arch == 'arm64') {
```
dsanders11, line 9: Can we mock `platform` and `arch` here? Otherwise this test only tests the code path if it's run on Apple silicon, which the CI tests won't cover. Let's try to cover both code paths with mocks - don't rely on the specifics of what machine is running the test.

src/utils/disable-download.ts
```diff
@@ -0,0 +1,11 @@
...
+    !isJest() &&
```
dsanders11, line 6: I'm not a big fan of the `isJest` stuff. Can we simply mock the return value for `disableDownload` in the relevant tests where it might mess with snapshots? Simply returning false if it's Jest feels wrong, leading to confusing test behavior like mocking `isJest` to return false.

BlackHole1: Let me change it later

tests/renderer/components/settings-electron-spec.tsx
Ayman161803: The tests were simply to make sure disableDownload() works for cases where there is a single download button and cases where there are multiple buttons. Nothing else in particular. I have removed one of the tests because it does look like they are trying to achieve the same thing.

src/utils/disable-download.ts
Ayman161803: Added JSDocs with a link to the apple silicon version 11.0.0 release blog.

BlackHole1:
Done🚀. All Unit test pass 🎉

BlackHole1:
Ping @dsanders11

dsanders11 (approved):
LGTM.

Sorry it took so long to circle back around on this @Ayman161803. Thanks for implementing it!

Ayman161803:
Heyy! Thanks for having this merged.

PR #1020, opened 2022-03-21
fix: use shell-env to fetch PATH

codebytere:
For some strange reason `await shellPath()` will no longer resolve on my macOS device. This fixes that.


ckerr (approved):

ckerr:
CI failure is unrelated to PR -- the semantic pull request bot seems to be unresponsive.

PR #1021, opened 2022-03-21
fix: add Contacts/Calendars etitlements

codebytere:
Fixes an issue where Fiddle apps could not access either user Contacts or Calendar data.

Discovered when trying to triage an issue with https://github.com/codebytere/node-mac-contacts


forge.config.js
```diff
@@ -36,6 +36,12 @@ const config = {
         'Access is needed by certain built-in fiddles in addition to any custom fiddles that use the Camera',
       Microphone:
         'Access is needed by certain built-in fiddles in addition to any custom fiddles that use the Microphone',
+      Calendars:
+        'Access is needed by certain built-in fiddles in addition to any custom fiddles that may access Calendars',
+      Contacts:
+        'Access is needed by certain built-in fiddles in addition to any custom fiddles that may access Contacts',
+      Reminders:
```
codebytere, line 43: Note that `com.apple.security.personal-information.calendars` covers both Reminders and Calendar access, but both descriptions are needed - see https://developer.apple.com/documentation/eventkit/ekeventstore?language=objc

erickzhao (approved):

PR #1026, opened 2022-03-29
chore: Remove old module.parent deprecation with require.main

### LINKED ISSUE #1023: NodeJS module.parent is deprecated.
https://github.com/electron/fiddle/blob/50b6656690801882a9fc907669de3002199e4d9b/src/main/main.ts#L152

In the above line `module.parent` is deprecated. I read the node Docs and they have suggested this 

image

Should we refactor the code or keep it this way? Does using the deprecated one have any disadvantages ?

akgupta0777:
Closes #1023 

NodeJs module.parent is deprecated and nodejs recommends module.children which does same thing as module.parent.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -149,6 +149,6 @@ export function main(argv_in: string[]) {
 }
 
 // only call main() if this is the main module
-if (typeof module !== 'undefined' && !module.children) {
+if (typeof module !== 'undefined' && require.main !== module) {
   main(process.argv);
 }
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -149,6 +149,6 @@ export function main(argv_in: string[]) {
 }
 
 // only call main() if this is the main module
-if (typeof module !== 'undefined' && !module.parent) {
+if (typeof module !== 'undefined' && !module.children) {
   main(process.argv);
 }


src/main/main.ts
```diff
@@ -149,6 +149,6 @@ export function main(argv_in: string[]) {
 }
 
 // only call main() if this is the main module
-if (typeof module !== 'undefined' && !module.parent) {
+if (typeof module !== 'undefined' && !module.children) {
```
erickzhao, line 152: https://nodejs.org/api/deprecations.html#dep0144-moduleparent

this code change doesn't look quite right

erickzhao (changes requested):

src/main/main.ts
akgupta0777: I read the nodejs module.children documentation and it returns the same object as module.parent. 
I will change this but how do i test the changes.

akgupta0777: I need some help

BlackHole1: #### I think this is the right way:
```typescript
 if (typeof module !== 'undefined' && require.main !== module) {
```

#### Local test way:

```shell
$ yarn run start --help
```
The correct output is:

image

As long as Fiddle opens properly after the change, then the change is correct.

akgupta0777: Thanks for help @BlackHole1

akgupta0777: Yes I can confirm The fiddle is running without any problems.

BlackHole1 (approved):
LGTM

codebytere (approved):

erickzhao (approved):

Issue #1024, opened 2022-03-28
[BUG]: `Preference` Page style is incorrect on macOS


aryanshridhar:
Noticed this issue today of application header being still visible when the user opens `Preference` page :

<img width="805" alt="Screenshot 2022-03-28 at 6 48 20 PM" src="https://user-images.githubusercontent.com/53977614/160406339-c2e2b24f-3591-4b85-8421-79ba4f374955.png">

<strong>Platform</strong>: Mac OS
<strong>Platform Version</strong>: Monterey v12.1
<strong>Fiddle Version</strong>: v0.27.3 development git

aryanshridhar:
This can be fixed by removing these CSS lines - 

https://github.com/electron/fiddle/blob/50b6656690801882a9fc907669de3002199e4d9b/src/less/components/dialogs.less#L15-L19

Maybe they weren't removed after the new UI patch?

erickzhao:
Yeah, I think this was a regression from a recent PR.

aryanshridhar:
Cool, I will leave this open for any new comer to come and pick this small issue :)

aryanshridhar:
Closing this with ref to #1028.

PR #1028, opened 2022-04-01
fix(style): Preference Page style is incorrect on macOS

Cheerego7:
Fiex #1024

PR #961 in order to use WCO on macOS then remove old style code, but there is still some old style was left behind

old style code: https://github.com/electron/fiddle/commit/befa3192f4f12a561e12c56c669e9dfea6caae00.
remove old style address: https://github.com/electron/fiddle/pull/961/files#diff-aa82273862a162ec96de7bbd7bbb5fbcbd412d5f577fc209fafcb7ad943dc2f9

### DIFF
--- a/src/less/components/dialogs.less
+++ b/src/less/components/dialogs.less
@@ -12,12 +12,6 @@
   }
 }
 
-.darwin {
-  .dialogs {
-    padding-top: 23px;
-  }
-}
-
 .dialog {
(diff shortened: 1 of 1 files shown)


BlackHole1 (approved):
LGTM

clavin (approved):

erickzhao (approved):

PR #1039, opened 2022-04-14
fix: preload.js and renderer.js files gone on monaco

BlackHole1:
Related PR: https://github.com/electron/fiddle/pull/969

The fix works fine locally.

---

This is because the `addFile` function receives the result of the `readFiddle` function. And `addFile` does not support handling `package.json` files

https://github.com/electron/fiddle/blob/b21e70001412666ff1ec6deaefef881271f50bff/src/renderer/editor-mosaic.ts#L116-L119

The order of the files is.
```
0: "index.html"
1: "main.js"
2: "package.json"
3: "preload.js"
4: "renderer.js"
```

Throw error when the file is `package.json`, `preload.js` and `renderer.js` will not be further processed by the addFile function

---

The `readFiddle` function is called in a number of places and we cannot be sure if it affects other functions.

So I've made this compatible here by adding a default parameter to `readFiddle`.

### DIFF
--- a/src/renderer/file-manager.ts
+++ b/src/renderer/file-manager.ts
@@ -70,7 +70,7 @@ export class FileManager {
 
     const editorValues = {};
     const files: [string, string][] = Object.entries(
-      await readFiddle(filePath),
+      await readFiddle(filePath, true),
     );
     for (const [name, value] of files) {
       if (name === PACKAGE_NAME) {
--- a/src/utils/read-fiddle.ts
+++ b/src/utils/read-fiddle.ts
@@ -8,14 +8,21 @@ import * as path from 'path';
  * Reads a Fiddle from a directory.
  *
  * @param {string} folder
+ * @param {boolean} includePackageJson
  * @returns {Promise<EditorValues>} the loaded Fiddle
  */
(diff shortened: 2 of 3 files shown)


BlackHole1:
This is because the `addFile` function receives the result of the `readFiddle` function. And `addFile` does not support handling `package.json` files

https://github.com/electron/fiddle/blob/b21e70001412666ff1ec6deaefef881271f50bff/src/renderer/editor-mosaic.ts#L116-L119

The order of the files is.
```
0: "index.html"
1: "main.js"
2: "package.json"
3: "preload.js"
4: "renderer.js"
```

Throw error when the file is `package.json`, `preload.js` and `renderer.js` will not be processing by the addFile function

BlackHole1:
When this PR has been merged, we need to release v0.28.1 @erickzhao 🙇‍♂️

erickzhao (approved):
Fixes the immediate issue, so I'm cool with releasing this. For the future:

1. We should probably not immediately `throw` an error when trying to load an unsupported file type like `package.json`, since that breaks the loop and causes the symptom we saw in this release.
2. I'd like for there to be a separate abstraction level for `package.json` since it should not be contained in `EditorValues` if we don't want it to be in an editor.

PR #1042, opened 2022-04-15
fix: electron settings style incorrect

BlackHole1:
This is caused by https://github.com/electron/fiddle/pull/1019

Problem preview:
图片


ckerr (approved):
wfm. With this patch, the channel checkboxes are restored to their previous horizontal layout :+1:

PR #1057, opened 2022-04-24
refactor: use bind-decorator to bind methods to this

### LINKED ISSUE #1027: Use Typescript Decorators for binding this to class Methods.
Currently in a lot of files, Class Methods are defined and binded with `this` keyword so that the correct object gets referenced on calling.
An example of this - 
https://github.com/electron/fiddle/blob/50b6656690801882a9fc907669de3002199e4d9b/src/renderer/bisect.ts#L9-L12

Since project is written in TypeScript we can use typescript **experimental decorators** on class Methods that will automatically bind `this` keyword to the functions attached with the decorator.

This is a great improvement in code quality and also we can reduce a lot of lines of code.

miniak:
Fixes #1027

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@octokit/rest": "^16.43.1",
     "@sentry/electron": "^2.5.3",
     "@types/getos": "^3.0.1",
+    "bind-decorator": "^1.0.11",
     "classnames": "^2.2.6",
     "commander": "^7.1.0",
     "electron-default-menu": "^1.0.2",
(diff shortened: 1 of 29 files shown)


package.json
```diff
@@ -43,6 +43,7 @@
     "@octokit/rest": "^16.43.1",
     "@sentry/electron": "^2.5.3",
     "@types/getos": "^3.0.1",
+    "bind-decorator": "^1.0.11",
```
akgupta0777, line 46: Instead of adding a new dependency can you make a custom Autobind decorator. Learn about them [here](https://www.typescriptlang.org/docs/handbook/decorators.html)

erickzhao:
I think the MobX upgrade clears the path for us to migrate off of class components altogether moving forward as well.

PR #1060, opened 2022-04-26
refactor: improve credits page

BlackHole1:
Before:
image

After:
image

### DIFF
--- a/src/less/components/settings.less
+++ b/src/less/components/settings.less
@@ -147,29 +147,52 @@
       background: rgba(255, 255, 255, 0.1);
       padding: 10px;
       margin: 0 10px 10px 0;
-      display: flex;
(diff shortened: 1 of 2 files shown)


BlackHole1:
@ckerr Yes, there will be far fewer avatars in the new version than before.

But there is a clear intention in doing so. This is because some people's bio is too long, resulting in an incomplete display. It affects the overall aesthetics and uniformity.

see: https://electronhq.slack.com/archives/CB389B7MH/p1650802514094799

BlackHole1:
@ckerr I think I would be willing to readjust in the way you have described. I've been busy lately though, so I may need to wait a few days. 😄

felixrieseberg:
I'm closing this PR because it's stale. We can always bring it back!

PR #1061, opened 2022-04-26
build(deps): mobx@6 upgrade

erickzhao:
Bumps up our `mobx` and `mobx-react` dependencies to 6.x.y. This is part of a wider attempt to modernize our renderer codebase, since we're stuck on old versions of React and using class components.

### Codemod

In order to perform this upgrade, I followed MobX's [Migrating from MobX 4/5](https://mobx.js.org/migrating-from-4-or-5.html) guide.

Because we use `@observer` decorators extensively, I decided to use the [`npx mobx-undecorate` codemod](https://www.npmjs.com/package/mobx-undecorate) provided by the guide. This will allow us to immediately upgrade to MobX 6 while keeping our React component architecture.

We can then perform a piecewise upgrade of our React components in the future.

### Caveats

* The types for the new `observer()` higher-order component from MobX mean that we use `any` types when mounting Enzyme components. This is a trade-off I'm willing to make because Enzyme doesn't support React 17 yet and we might want to look towards migrating towards another unit testing library.
* This diff is huge. Reviewing it will be easier with the "hide whitespace" option.


tests/renderer/components/commands-address-bar-spec.tsx
```diff
@@ -31,7 +31,7 @@ describe('AddressBar component', () => {
 
   it('handles change', () => {
     const wrapper = shallow(<AddressBar appState={store as any} />);
-    const instance: AddressBar = wrapper.instance() as any;
+    const instance: any = wrapper.instance() as any;
```
BlackHole1, line 34:
```suggestion
    const instance = wrapper.instance() as any;
```

tests/renderer/components/commands-address-bar-spec.tsx
```diff
@@ -48,7 +48,7 @@ describe('AddressBar component', () => {
   it('handles submit', () => {
     const preventDefault = jest.fn();
     const wrapper = shallow(<AddressBar appState={store as any} />);
-    const instance: AddressBar = wrapper.instance() as any;
+    const instance: any = wrapper.instance() as any;
```
BlackHole1, line 51: Above

BlackHole1 (changes requested):

BlackHole1:
Perhaps it would be better if we used the following?

``` ts
// AddThemeDialog.tsx
export class AddThemeDialogComponent {
  // code
}
export const AddThemeDialog = observer(AddThemeDialogComponent);


// AddThemeDialog-spec.tsx
const wrapper = shallow(<AddThemeDialog appState={store as any} />);
const instance = wrapper.instance() as AddThemeDialogComponent;
```

ckerr (commented):
Even in "hide whitespace" mode, it seems like a lot of this diff is caused by line wrap changes :roll_eyes: This PR isn't as scary as it initially looked

ckerr (approved):
Looks fine to me. The only thing I don't like abou this PR is the reversion from specific types to `any` but as per the description that seems to be unavoidable for now.

I don't have any extra change requests that weren't already discussed by @BlackHole1

erickzhao:
FWIW the `:any` changes in our specs will likely be solved with the migration from Enzyme to RTL: https://github.com/electron/fiddle/pull/1064

Issue #956, opened 2022-01-24
[Feature Request]: Font preferences


nornagon:
I'd love to be able to set the font family / size in the code window!

PR #1069, opened 2022-05-09
feat: add font size and family customization

codebytere:
Closes https://github.com/electron/fiddle/issues/956.

Adds a settings customization option for font size and family.

Screen Shot 2022-05-10 at 12 52 33 PM


src/renderer/components/editor.tsx
```diff
@@ -75,14 +75,18 @@ export class Editor extends React.Component<EditorProps> {
...
 
+    const { fontFamily, fontSize } = appState;
+
     if (ref) {
       this.editor = monaco.editor.create(ref, {
         automaticLayout: true,
         language: this.language,
         theme: 'main',
+        ...(fontFamily && { fontFamily }),
+        ...(fontSize && { fontSize }),
```
BlackHole1, line 89:
```suggestion
        fontFamily,
        fontSize,
```

src/main/windows.ts
```diff
@@ -82,6 +82,12 @@ export function createMainWindow(): Electron.BrowserWindow {
     shell.openExternal(url);
   });
 
+  ipcMainManager.on(IpcEvents.RELOAD_WINDOW, () => {
+    if (browserWindow) {
+      browserWindow.reload();
+    }
```
BlackHole1, line 88:
```suggestion
    browserWindow?.reload();
```

src/renderer/components/settings-general-font.tsx
```diff
@@ -0,0 +1,104 @@
...
+interface FontSettingsState {
+  fontSize?: number;
+  fontFamily?: string;
+}
+
+/**
+ * Settings content to manage GitHub-related preferences.
+ *
+ * @class GitHubSettings
+ * @extends {React.Component<AppearanceSettingsProps, AppearanceSettingsState>}
```
BlackHole1, line 22:
```suggestion
 * Settings font family and size.
 *
 * @class FontSettings
 * @extends {React.Component<FontSettingsProps, FontSettingsState>}
```

BlackHole1 (changes requested):

src/renderer/components/editor.tsx
codebytere: What motivates this suggestion? Both of these can be undefined, and therefore we wouldn't want to pass these values as `undefined` - we only want to set them when they're set, otherwise we would leave them as default.

BlackHole1: e.g:
image

And I've just actually tested it.
When the value is undefined, the result is the same as if `fontFamily` and `fontSize` were not passed in.
In fact, it is a means of optimizing the code.

BlackHole1 (approved):

PR #1071, opened 2022-05-11
chore: stop building Windows on arm64

dsanders11:
Due to an update bug involving multiple architectures and Windows, stop building arm64 builds for Windows to prevent wrong architecture updates.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,6 +53,9 @@ jobs:
           arch: ia32
         - os: ubuntu-latest
           arch: armv7l
+        # Publishing artifacts for multiple Windows architectures has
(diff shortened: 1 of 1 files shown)


erickzhao (approved):
Would like a comment in the source for this, but LGTM!

BlackHole1 (approved):

BlackHole1:
image

hmmm... Perhaps we need to modify GitHub PR settings?

dsanders11:
@erickzhao, added a comment. 👍 

@BlackHole1, I think maybe that will go away once this gets merged? Might be a security thing on GH so that someone can't hide checks in their PR and make it look like everything is green.

codebytere:
@dsanders11 @erickzhao it was still required in branch protection settings - i've turned that off.

Issue #1073, opened 2022-05-16
Fiddle does not list available Electron versions for Windows x64


craxal:
1. Launch Fiddle (Windows x64) and open Preferences.
2. Select Electron, check Stable and Beta.
3. Click Update Electron Release List.

**Expected**

A list of available Electron versions is shown.

**Actual**

Nothing is listed. I am unable to choose a different version of Electron to try. Uninstalling and reinstalling does nothing.

![image](https://user-images.githubusercontent.com/3473026/168676771-0aeeb34f-0f9f-4183-8e45-cf54099bd048.png)

**Comments**

On inspecting web traffic, I notice there is a web request for a JSON file that contains all the Electron version information. My browser displays that information when I follow the link just fine, but the response to the web request sent by Fiddle comes back with a 304 Not Modified code and no JSON information.

![image](https://user-images.githubusercontent.com/3473026/168676854-66095f56-af9c-4a55-9d71-87af67cda003.png)

craxal:
Has the fix been shipped in a new version yet? This still isn't working for me as of 0.29.0

BlackHole1:
Hi @craxal 👋. This change is already in version 0.29.0.
The problem you are experiencing is bizarre and we have not received any reports of it from other users. 🥲
Do you have any proxy software turned on? Or have you tried changing your network? For example mobile WIFI? 

Did you completely delete all the files and cache from fiddle when retesting? 
You can try using something like https://crystalidea.com/uninstall-tool to uninstall fiddle. and reinstall

craxal:
@BlackHole1
> Proxy software?

No

> Changing network?

I'm running on a desktop machine in my work office. Not much I can do to change it. I don't have this problem on other machines on the same network, however, so I doubt that's the problem.

> Completely delete files and cache?

This would also be my guess, that the cache still exists somewhere. I've uninstalled Fiddle several times, making sure to delete "electron-fiddle" from `%LOCALAPPDATA%`. Is the cache located elsewhere? However, I am NOT going to use a third party tool to uninstall Fiddle. If I can just locate the cache, that might help me resolve.

craxal:
@erickzhao @BlackHole1 Can anyone please help me with this, please? This is blocking my ability to properly test/experiment/create Electron bug samples.

BlackHole1:
[...]

PR #1075, opened 2022-05-17
fix: not list available electron releases versions

BlackHole1:
Fix: #1073

### DIFF
--- a/src/renderer/versions.ts
+++ b/src/renderer/versions.ts
@@ -233,7 +233,9 @@ function saveKnownVersions(versions: Array<Version>) {
  */
 export async function fetchVersions(): Promise<Version[]> {
   const url = 'https://releases.electronjs.org/releases.json';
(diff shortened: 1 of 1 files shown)


ckerr (approved):
Looks like it should fix the 304. :+1:

Though, _if_ we're hitting this endpoint a lot -- which, since this code has been in place for awhile and we haven't gotten other bug reports, seems unlikely --  then maybe we _should_ cache it somewhere?

BlackHole1:
> Though, if we're hitting this endpoint a lot -- which, since this code has been in place for awhile and we haven't gotten other bug reports, seems unlikely.

@ckerr I think the problem is related to AWS S3 / Azure Blob storage. 🤔

>  then maybe we should cache it somewhere?

It seems to me that there is no need to cache this if it is just a JSON file? Because there is basically no pressure on the server.

Issue #1065, opened 2022-05-05
Loading fiddle cuts off the right action buttons


aryanshridhar:
When a fiddle is loaded using the `Load fiddle` button, the right actions button gets cut off the screen. I was left wondering for a minute that were did the popover button go!

<strong>Before:</strong>

<img width="1440" alt="Screenshot 2022-05-05 at 6 46 26 PM" src="https://user-images.githubusercontent.com/53977614/166937067-7e63d3f1-6628-4b68-a49d-345a278fb6bb.png">

<strong>After loading fiddle:</strong>

<img width="1440" alt="Screenshot 2022-05-05 at 6 46 37 PM" src="https://user-images.githubusercontent.com/53977614/166937206-c673dc3c-9e4e-4629-b44d-e2bfa6bd2fde.png">

This is so because the header updates itself to add the link of the gist which takes up the major space.

<strong>Platform</strong>: Mac OS
<strong>Platform Version</strong>: Monterey v12.1
<strong>Fiddle Version</strong>: v0.27.3 development git

erickzhao:
Yep, we should probably make this more responsive.

BlackHole1:
Maybe we can change it to an A tag? e.g.

[gist#git short hash]() -> [gist#abcd1234]()

aryanshridhar:
> Maybe we can change it to an A tag? e.g.
> 
> gist#git short hash -> gist#abcd1234

This will look something like - 
<img width="1358" alt="Screenshot 2022-05-05 at 11 49 17 PM" src="https://user-images.githubusercontent.com/53977614/166987901-ab8784f3-7072-469e-b2a0-d3b26a4998b5.png">

It works fine, but the user should have an atleast window width of ~1340 px. (If the user uses it in window size < 1340 px, the buttons will get cut again!)

BlackHole1:
> > Maybe we can change it to an A tag? e.g.
> > gist#git short hash -> gist#abcd1234
> 
> This will look something like - <img alt="Screenshot 2022-05-05 at 11 49 17 PM" width="1358" src="https://user-images.githubusercontent.com/53977614/166987901-ab8784f3-7072-469e-b2a0-d3b26a4998b5.png">
> 
> It works fine, but the user should have an atleast window width of ~1340 px. (If the user uses it in window size < 1340 px, the buttons will get cut again!)

Oh. I mean is `short hash`. e.g: `gist#94d865a4`, not long hash. 😄

aryanshridhar:
[...]

PR #1081, opened 2022-05-26
chore: remove extraneous text from title

codebytere:
Closes https://github.com/electron/fiddle/issues/1065.

Removes extra cruft around title such that it only flags save state. The rest is, imo, not particularly necessary and results in buttons being cut off.


erickzhao (approved):

erickzhao:
Re-kicked CI just now, seems like a flake in accessing a README image.

BlackHole1 (approved):

PR #1092, opened 2022-05-29
Make close button focusable

### LINKED ISSUE #1090: Close (X) button in Settings not focusable at all
When you go to any page in the Settings, there is an `X` that appears in the upper righthand corner. You can click that to exit this view, but it's not focusable at all when navigating with a keyboard. 

Screenshot of the top of the Executions page. In the corner you can see a white X.

This is likely because the `X`, semantically speaking, is not a `button` when it likely should be 😬 

From `settings.tsx`:
```
<div className="settings-close" onClick={appState.toggleSettings}>
    <Icon icon="cross" iconSize={25} />
</div>
```

Hanaffi:
# After 

Closes #1090

### DIFF
--- a/src/less/components/settings.less
+++ b/src/less/components/settings.less
@@ -1,4 +1,4 @@
-@import "../variables.less";
+@import '../variables.less';
 
 .settings {
   background: @background-3;
@@ -35,7 +35,7 @@
     width: 250px;
   }
 
- .bp3-input-group {
+  .bp3-input-group {
     margin-bottom: 15px;
   }
 
@@ -126,6 +126,8 @@
     cursor: pointer;
     border-radius: 15px;
     color: var(--text-color-3);
+    padding: 0 0;
+    border: 0;
 
     &:hover {
       color: @text-color-1;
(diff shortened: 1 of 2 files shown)


src/less/components/settings.less
```diff
@@ -1,4 +1,4 @@
-@import "../variables.less";
+@import '../variables.less';
```
BlackHole1, line 1:
```suggestion
@import "../variables.less";
```

src/less/components/settings.less
```diff
@@ -35,7 +35,7 @@
     width: 250px;
   }
 
- .bp3-input-group {
+  .bp3-input-group {
```
BlackHole1, line 38:
```suggestion
 .bp3-input-group {
```

BlackHole1 (changes requested):
LGTM. But has some nit:

1. Incorrect code format
2. Test snapshot forget to update (You can run: `yarn run test -- -u`)

Hanaffi:
@BlackHole1 Thanks, I'm using prettier already. The formatting part was weird as I'm using your rules.

BlackHole1 (approved):

codebytere (approved):

PR #1098, opened 2022-06-09
test: properly arrange address bar test conditions

erickzhao:
It looked like the test I wrote was trivial due to a short-circuit. Without the gist address in the input, the button that I was asserting on would always be disabled regardless of status.

This PR adds the correct arrangement step so that the test would fail without the `store.activeGistAction = action` line.


rtl-spec/commands-address-bar.spec.tsx
```diff
@@ -62,10 +63,17 @@ describe('AddressBar component', () => {
...
+    const gistId = '159cb99a70a201bd5e08194674f4c571';
+    const gistUrl = `https://gist.github.com/ghost/${gistId}`;
+    const { getByPlaceholderText, getByRole } = render(
       <AddressBar appState={(store as unknown) as AppState} />,
     );
-    store.activeGistAction = action;
+    fireEvent.change(getByPlaceholderText('https://gist.github.com/...'), {
+      target: { value: gistUrl },
+    });
+    runInAction(() => {
```
erickzhao, line 74: This is a workaround for

> [MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed.

rtl-spec/commands-address-bar.spec.tsx
```diff
@@ -62,10 +63,17 @@ describe('AddressBar component', () => {
...
-    const { getByRole } = render(
+    const gistId = '159cb99a70a201bd5e08194674f4c571';
+    const gistUrl = `https://gist.github.com/ghost/${gistId}`;
+    const { getByPlaceholderText, getByRole } = render(
       <AddressBar appState={(store as unknown) as AppState} />,
     );
-    store.activeGistAction = action;
+    fireEvent.change(getByPlaceholderText('https://gist.github.com/...'), {
+      target: { value: gistUrl },
+    });
```
BlackHole1, line 73: Maybe we should wait for https://github.com/electron/fiddle/pull/1099 to merge first. Because of this, we can avoid using the fireEvent method and instead use the `type` API in `@testing-library/user-event`

e.g: 
```ts
userEvent.type(getByPlaceholderText('https://gist.github.com/...'), gistUrl)
```

https://testing-library.com/docs/ecosystem-user-event#typeelement-text-options

BlackHole1: Update: #1099 already merged.

BlackHole1 (approved):
LGTM

PR #1099, opened 2022-06-09
test(rtl): refactor commands-bisect component test

erickzhao:
Replaces the old `commands-bisect-spec.tsx` Enzyme test with new shiny RTL ones.

Along the way, there are a couple non-test changes to note:
* New dependency: `@testing-library/user-event` is apparently the way to go to simulate user events over the built-in `fireEvent` helper method.
* I added new `aria-label` properties to the 👍 and 👎 buttons. This makes the buttons more accessible!


tests/mocks/state.ts
```diff
@@ -59,7 +59,7 @@ export class StateMock {
   public electronMirror = ELECTRON_MIRROR;
   public isBisectCommandShowing = false;
 
-  public Bisector = new BisectorMock();
+  public Bisector: BisectorMock | undefined = new BisectorMock();
```
BlackHole1, line 62: Under what circumstances will undefined be returned here?

tests/renderer/components/__snapshots__/commands-bisect-spec.tsx.snap
```diff
@@ -1,51 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
BlackHole1, line 1: Looks like we won't be needing snapshots anymore, right?

BlackHole1 (commented):
LGTM. A very great change!

It seems that there will be no more unit test failures caused by new contributors changing the UI layout and forgetting to execute `yarn run test -- -u`.

tests/mocks/state.ts
erickzhao: `Bisector` is undefined if there is no active bisect running. This is tested under the `'renders bisect dialog button if no bisect instance'` test!

tests/renderer/components/__snapshots__/commands-bisect-spec.tsx.snap
erickzhao: We could always use inline Jest snapshots to explicitly test UI bits, but I don't see us using large snapshot files.

BlackHole1: Okey

BlackHole1 (approved):

PR #1101, opened 2022-06-13
fix: headings

### LINKED ISSUE #1087: Heading levels should only increase by one
There are a few places where the HTML of the headings doesn't follow logical order. This is an accessibility issue because it makes navigating more confusing for users of assistive technology ([see this deque article for more details](https://dequeuniversity.com/rules/axe/4.0/heading-order)).

Here's an example from the Settings page, where the main header is an `h2` element then the next header is `h4`:

And another example from the Credits page, where the main header is an `h2` element then the next header is `h5`:

(there's no `h1` element either which is [also an issue](https://dequeuniversity.com/rules/axe/4.1/page-has-heading-one) 👀

sanz17:
Fixed the uneven headings stylings, and also places where it should begin with h1 headings
closes #1087


BlackHole1 (changes requested):
Please reset `package-lock.json` and `yarn.lock` file.

> If you want an init fiddle project, you can run `yarn install --force --frozen-lockfile`.

codebytere:
@sanz17 we don't want a `package-lock.json` at all given that we have a yarn lockfile and use yarn in this repo - can you please delete it?

codebytere (commented):
@sanz17 we don't want a package-lock.json at all given that we have a yarn lockfile and use yarn in this repo - can you please delete it?

sanz17:
deleted package-lock.json

codebytere (approved):

PR #1107, opened 2022-06-16
BREAKING CHANGE: Upgrade yarn to Berry

tiziodcaio:
Hello I tried successfully to upgrade to yarn v3... Here the result.
If you have any question no problem to answer it! 😄

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,7 +27,7 @@ jobs:
           node-version: 15.x
           cache: yarn
       - name: Install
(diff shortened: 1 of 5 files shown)


codebytere (changes requested):
Hi - while we appreciate the effort here, this is a significant change that was never discussed with maintainers via issue and which changes package manager semantics. Yarn Berry diverges notably from Yarn Classic - we as maintainers have so far intentionally made the choice to use Yarn Classic and not Yarn Berry, and further, this type of change is not one we would typically accept from an outside maintainer.

If you would like to discuss this further, you're welcome to make an issue outlining your motivation behind this change, which we can then discuss before proceeding further.

ckerr:
`electron/fiddle` should use whatever `electron/electron` is using.

Since `electron/electron` is not using Berry, I'd vote for closing this PR.

PR #1109, opened 2022-06-16
forge: introduce webpack plugin and remove parcel

aryanshridhar:
Introduces forge+webpack configuration and removes parcel for the bundling process.

<strong>TODOs</strong>
* Handle the management of assets files in the webpack configuration.
* Figure out the webpack management for `process.env.NODE_ENV` in main process.
* Fix the current issue/current blocker (listed below)

<strong>Current blocker - </strong>

```
Uncaught ReferenceError: require is not defined
    events external node-commonjs "events":1
    Webpack 12
external node-commonjs "events":1
```

Corresponding unanswered (according to our use case) thread: https://stackoverflow.com/questions/66112475/electron-forge-with-webpack-getting-require-is-not-defined-when-importing-eve.


forge.config.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 const packageJson = require('./package.json');
+const root = process.cwd();
```
dsanders11, line 4: Very minor nit, but can we move this down to be in the section on lines 6-7 instead? Usually the section of requires at the top of the file only contains requires.

aryanshridhar: Cool!

tools/webpack/webpack.renderer.config.js
```diff
@@ -0,0 +1,80 @@
...
+});
+
+// Fonts
+rules.push({
+  test: /\.(eot|ttf|woff2?)(\?v=\d+\.\d+\.\d+)?$/i,
+  type: 'asset/resource',
+  generator: {
+    filename: 'fonts/[name][ext]',
+  },
+});
```
dsanders11, line 66: Let's change these. The current code, pushing into the array exported by the other modules, it actually mutating the array value. That's generally not a good idea as it can introduce subtle bugs (any other import after this one will see the mutated value).

Let's use spread syntax to accomplish this without mutating `rules` and `plugins`. It would look something like this:

```
module.exports = {
  module: {
    rules: [
        ..rules,
        {
            test: /\.(eot|ttf|woff2?)(\?v=\d+\.\d+\.\d+)?$/i,
            type: 'asset/resource',
            generator: {
                filename: 'fonts/[name][ext]',
            },
            // etc
        }
    ]
  },
  plugins: [
      ...plugins,
      new MonacoWebpackPlugin({
          languages: ['typescript', 'javascript', 'html', 'css'],
      }),
      // etc
  ]
```

dsanders11 (changes requested):

src/main/windows.ts
```diff
@@ -50,7 +57,9 @@ export function createMainWindow(): Electron.BrowserWindow {
   console.log(`Creating main window`);
   let browserWindow: BrowserWindow | null;
   browserWindow = new BrowserWindow(getMainWindowOptions());
-  browserWindow.loadFile('./dist/static/index.html');
+  browserWindow.loadURL(
+    !!process.env.JEST ? './dist/static/index.html' : MAIN_WINDOW_WEBPACK_ENTRY,
```
dsanders11, line 61: I think this should actually be `./.webpack/renderer/main_window/index.html`... I don't think the tests are actually making usage of it, but let's change it to that.

Same with `preload` above, let's change that to `./.webpack/renderer/main_window/preload.js`.

tools/webpack/webpack.renderer.config.js
aryanshridhar: Yup, Sounds right! I was only doing that in only renderer process as it's the last piece of code that executes :)

src/renderer/templates.ts
```diff
@@ -2,17 +2,17 @@ import * as path from 'path';
...
  * @returns {Promise<EditorValues>}
  */
 export function getTemplateValues(name: string): Promise<EditorValues> {
-  const folder = path.join(
-    __dirname,
-    '../../static/show-me',
-    name.toLowerCase(),
-  );
-  return readFiddle(folder);
+  // TODO: Figure out a better method to handle template path
```
dsanders11, line 14: Let's drop this comment.

src/renderer/components/settings-credits.tsx
```diff
@@ -19,9 +18,9 @@ interface Contributor {
   api: string;
   login: string;
   avatar: string;
-  name: string;
-  bio: string;
-  location: string;
+  name: string | null;
+  bio: string | null;
+  location: string | null;
```
dsanders11, line 23: Are these required changes to make things work, or just updating this to be more accurate? If the latter, let's not include it in this PR, can be done in a follow-up. Since this is already a large PR, I'd like to keep things focused.

forge.config.js
```diff
@@ -1,9 +1,11 @@
 const path = require('path');
 const fs = require('fs');
 const packageJson = require('./package.json');
+const contributors = require('./tools/contributors');
```
dsanders11, line 4: nit: `const { maybeFetchContributors } = require('./tools/contributors');` and avoid the awkward `contributors` name.

src/renderer/components/settings-credits.tsx
aryanshridhar: I had to add these as within test `mockContributorsBroken` is defined holding these values as `null` as well (It gives an type error when calling the method - `getContributors`)

aryanshridhar: But now with using `wrapper.setState({ contributors })`, This isn't required, so I will be reverting this :)

.github/workflows/build.yaml
```diff
@@ -28,6 +28,12 @@ jobs:
           cache: yarn
       - name: Install
         run: yarn --frozen-lockfile --network-timeout 100000 || yarn --frozen-lockfile --network-timeout 100000 || yarn --frozen-lockfile --network-timeout 100000
+      - name: Create fake contributors
+        uses: 1arp/create-a-file-action@0.2
+        with:
+          path: 'static'
+          file: 'contributors.json'
+          content: "[]"
```
dsanders11, line 36: Just adding a note here to investigate why `tests/globalSetup.js` doesn't work for CI for creating this file. 👍

VerteDinde (approved):
Nice job!

PR #1115, opened 2022-06-22
fix: disable right click contextMenu in settings page

### LINKED ISSUE #1070: `ContextMenu` should be disabled in Settings page
Noticed a minor issue of the Right click `ContextMenu` which is can be seen in `Preference`/`Settings` page as well.

Screenshot 2022-05-10 at 10 04 39 PM

aryanshridhar:
Minor commit to fix #1070.
Edit: Not sure on how to mock the `contextMenu` behavior.

### DIFF
--- a/tests/renderer/components/settings-spec.tsx
+++ b/tests/renderer/components/settings-spec.tsx
@@ -90,10 +90,37 @@ describe('Settings component', () => {
 
     // trigger mock 'keyup' event
     map.keyup({ code: 'Escape' });
-    expect(Object.keys(map)).toHaveLength(1);
+    expect(Object.keys(map)).toContain('keyup');
+    expect(Object.keys(map)).toHaveLength(2); // ['keyup','contextmenu']
     expect(store.isSettingsShowing).toBe(false);
 
-    // check if event listener is removed upon unmount
+    // check if the event listeners are removed upon unmount
+    wrapper.unmount();
+    expect(Object.keys(map)).toHaveLength(0);
+  });
+
+  it('makes sure the contextmenu is disabled', () => {
+    expect(store.isSettingsShowing).toBe(true);
+    // mock event listener API
+    const map: any = {};
+    window.addEventListener = jest.fn().mockImplementation((event, cb) => {
+      map[event] = cb;
+    });
+
+    window.removeEventListener = jest.fn().mockImplementation((event) => {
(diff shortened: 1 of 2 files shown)


tests/renderer/components/settings-spec.tsx
```diff
@@ -118,15 +118,15 @@ describe('Settings component', () => {
...
     // Toggle the setting again as if it was closed
     instance.toggleHasPopoverOpen();
 
     // trigger mock 'keyup' event
     map.keyup({ code: 'Escape' });
-    expect(Object.keys(map)).toHaveLength(1);
+    expect(Object.keys(map)).toHaveLength(2); // {'keyup','contextmenu'}
     expect(store.isSettingsShowing).toBe(false);
 
     // check if event listener is removed upon unmount
```
dsanders11, line 132: Let's update this comment to be "// check if event listeners are removed upon unmount" now that there's multiple.

tests/renderer/components/settings-spec.tsx
```diff
@@ -118,15 +118,15 @@ describe('Settings component', () => {
...
     expect(store.isSettingsShowing).toBe(true);
 
     // Toggle the setting again as if it was closed
     instance.toggleHasPopoverOpen();
 
     // trigger mock 'keyup' event
     map.keyup({ code: 'Escape' });
-    expect(Object.keys(map)).toHaveLength(1);
+    expect(Object.keys(map)).toHaveLength(2); // {'keyup','contextmenu'}
     expect(store.isSettingsShowing).toBe(false);
```
dsanders11, line 131:
```suggestion

    // trigger mock 'contextmenu' event
    const preventDefault = jest.fn();
    map.contextmenu({ preventDefault });
    expect(Object.keys(map)).toHaveLength(2); // {'keyup','contextmenu'}
    expect(preventDefault).toHaveBeenCalled();

```

I haven't tested it, but you could try that to get a bit of test coverage. 🙂

dsanders11 (changes requested):
Overall LGTM, couple minor comments. I think there's a lot more issues with the contextmenu throughout Fiddle (it's in a lot of places that probably don't make sense) but this is a step in the right direction.

BlackHole1 (approved):
LGTM

dsanders11 (approved):

Issue #1111, opened 2022-06-21
Black screen when navigating to Bisect helper


aryanshridhar:
Fiddle displays an blank screen when one navigates to the Bisect page - 

![Screenshot 2022-06-21 at 5 05 42 PM](https://user-images.githubusercontent.com/53977614/174790130-8729d93d-f523-40b4-ba5b-6fdd0befb449.png)

**Trace**

```
Uncaught TypeError: Cannot destructure property 'version' of 'currentVersion' as it is undefined.
    at VersionSelect.render (main.99d03ea8.js:8201:7)
    at Object.allowStateChanges [as _allowStateChanges] (action.ts:154:16)
    at observerClass.ts:142:29
    at trackDerivedFunction (derivation.ts:183:24)
    at Reaction.track (reaction.ts:137:24)
    at VersionSelect.reactiveRender [as render] (observerClass.ts:140:18)
    at VersionSelect.makeComponentReactive (observerClass.ts:153:27)
    at VersionSelect.target.render (observerClass.ts:63:38)
    at finishClassComponent (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:17160:31)
    at updateClassComponent (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:17110:24)
    at beginWork (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:18620:16)
    at HTMLUnknownElement.callCallback (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:188:14)
    at HTMLUnknownElement.sentryWrapped (helpers.ts:87:17)
    at Object.invokeGuardedCallbackDev (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:237:16)
    at invokeGuardedCallback (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:292:31)
    at beginWork$1 (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:23203:7)
    at performUnitOfWork (/Users/aryanshridhar/Desktop/fiddle/node_modules/react-dom/cjs/react-dom.development.js:22154:12)
[...]

PR #1118, opened 2022-06-24
fix: join bisect page crash when electron version total < 11

BlackHole1:
close: #1111

### DIFF
--- a/src/renderer/components/dialog-bisect.tsx
+++ b/src/renderer/components/dialog-bisect.tsx
@@ -41,9 +41,11 @@ export const BisectDialog = observer(
       this.isEarliestItemDisabled = this.isEarliestItemDisabled.bind(this);
       this.isLatestItemDisabled = this.isLatestItemDisabled.bind(this);
 
+      const allVersions = this.props.appState.versionsToShow;
+
       this.state = {
-        allVersions: this.props.appState.versionsToShow,
-        startIndex: 10,
+        allVersions: allVersions,
+        startIndex: allVersions.length - 1 >= 10 ? 10 : allVersions.length - 1,
         endIndex: 0,
       };
     }
--- a/src/renderer/components/dialog-bisect.tsx
+++ b/src/renderer/components/dialog-bisect.tsx
@@ -44,8 +44,8 @@ export const BisectDialog = observer(
       const allVersions = this.props.appState.versionsToShow;
 
       this.state = {
-        allVersions: allVersions,
-        startIndex: allVersions.length - 1 >= 10 ? 10 : allVersions.length - 1,
+        allVersions,
+        startIndex: allVersions.length > 10 ? 10 : allVersions.length - 1,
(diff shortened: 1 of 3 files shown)


src/renderer/components/dialog-bisect.tsx
```diff
@@ -41,9 +41,11 @@ export const BisectDialog = observer(
       this.isEarliestItemDisabled = this.isEarliestItemDisabled.bind(this);
       this.isLatestItemDisabled = this.isLatestItemDisabled.bind(this);
 
+      const allVersions = this.props.appState.versionsToShow;
+
       this.state = {
-        allVersions: this.props.appState.versionsToShow,
-        startIndex: 10,
+        allVersions: allVersions,
```
dsanders11, line 47:
```suggestion
        allVersions,
```

This can be simplified since it has the same name as the key.

src/renderer/components/dialog-bisect.tsx
```diff
@@ -41,9 +41,11 @@ export const BisectDialog = observer(
...
+        startIndex: allVersions.length - 1 >= 10 ? 10 : allVersions.length - 1,
```
dsanders11, line 48:
```suggestion
        startIndex: allVersions.length > 10 ? 10 : allVersions.length - 1,
```

I think this makes it more clear what's going on here.

tests/renderer/components/dialog-bisect-spec.tsx
```diff
@@ -79,7 +79,7 @@ describe('BisectDialog component', () => {
       const wrapper = shallow(<BisectDialog appState={store as any} />);
       const instance: any = wrapper.instance() as any;
 
-      expect(instance.state.startIndex).toBe(NUM_VERSIONS);
+      expect(instance.state.startIndex).toBe(NUM_VERSIONS - 1);
```
dsanders11, line 82: Could we extend this test to have test coverage for both cases: > 10 and <= 10?

I think we could change the test to be `describe.each([8, 15])('BisectDialog component', (numVersions) => {` and adjust accordingly.

This `expect` could then be changed to: `expect(instance.state.startIndex).toBeLessThan(numVersions);`, which would ensure we don't regress where the `startIndex` is larger than the number of versions. You could then add another `expect` to check the logic around setting the start index at 10 if there's more than 10 versions, like: `expect(instance.state.startIndex).toBe(numVersions > 10 ? 10 : numVersions - 1);`

I think that would give us robust test coverage for this issue. 🙂

dsanders11 (commented):
Nice work debugging this issue @BlackHole1! Fix LGTM, just a few comments to further improve it. 👍

tests/renderer/components/dialog-bisect-spec.tsx
BlackHole1: Good idea!

dsanders11 (approved):

PR #1119, opened 2022-06-24
build: upgrade to Forge 6.0.0-beta.64

dsanders11:
Another prerequisite PR for #1109, this fixes a Forge-related error that was happening when closing Fiddle when using the Webpack plugin.


BlackHole1 (approved):

VerteDinde (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.29.0",
   "description": "The easiest way to get started with Electron",
   "repository": "https://github.com/electron/fiddle",
-  "main": "./dist/src/main/main",
+  "main": "./dist/src/main/main.js",
```
VerteDinde, line 7: Small nit, but do we need this .js addition?

dsanders11: We do. Couldn't tell you why, but `yarn make`/`yarn package` no longer worked with `beta64` without it. 🙂 So that was the only change that needed to be made to work with `beta64`, as far as I could tell.

codebytere (approved):

PR #1134, opened 2022-07-15
refactor: integrate fiddle-core

aryanshridhar:
Integrates fiddle core into the `renderer` codebase

<strong>Todos: </strong>

* ~Fix the `TODOS` over the codebase. (Only one left, not a major one!)~
* ~Fix jests tests (to be done later)~
* Handle `bisecting` of versions through fiddle-core.
* ~Handle different source mirrors in fiddle-core (`global`/`china`)~
* ~Add the `downloadProgress` callback to `fiddle-core` (Opened https://github.com/electron/fiddle-core/pull/30 for the same)~

Tackles #644 as well :)


src/interfaces.ts
```diff
@@ -2,13 +2,6 @@ export type Files = Map<string, string>;
 
 export type FileTransform = (files: Files) => Promise<Files>;
 
-export enum VersionState {
-  ready = 'ready',
-  downloading = 'downloading',
-  unzipping = 'unzipping',
-  unknown = 'unknown',
-}
```
VerteDinde, line 10: Why remove the enum here? Is there a way we can retain this type, so we can make sure that only allowed strings are used when working with versions?

aryanshridhar: Oh yeah! But rather than creating a new enum here, it's better if we use the [one defined](https://github.com/electron/fiddle-core/blob/9007a90720c98e46e749b241cfe3d6bbf4d374c3/src/installer.ts#L22) in fiddle-core itself, right ?

aryanshridhar: Oh yeah! Probably, the same should be done in fiddle-core as well right ?
[(one such example)](https://github.com/electron/fiddle-core/blob/9007a90720c98e46e749b241cfe3d6bbf4d374c3/src/installer.ts#L180)

src/renderer/runner.ts
```diff
@@ -350,59 +374,57 @@ export class Runner {
...
-      currentElectronVersion,
-      flushOutput,
-      pushOutput,
-      executionFlags,
-    } = this.appState;
-
-    const { version, localPath } = currentElectronVersion;
-    const binaryPath = getElectronBinaryPath(version, localPath);
-    console.log(`Runner: Binary ${binaryPath} ready, launching`);
+  private async runFiddle(params: RunFiddleParams) {
```
aryanshridhar, line 380: While fixing tests -- realized that this function always returns `RunResult.INVALID`
Fixed this now :)

tests/renderer/state-spec.ts
```diff
@@ -77,6 +76,10 @@ describe('AppState', () => {
     expect(appState).toBeTruthy();
   });
 
+  afterAll(async () => {
+    await new Promise((resolve) => setTimeout(resolve, 2000));
```
aryanshridhar, line 80: This timeout waits for the `appState.runner` promise to resolve (most likely!) Otherwise CI throws an error saying -
```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From tests/renderer/state-spec.ts.
```
(As replicated [here](https://github.com/electron/fiddle/runs/7584809546?check_suite_focus=true#step:7:10))

I am quite baffled by this as I thought something like - `await appState.runner` will wait for the promise. Tried that but still CI throws an error.
After many rounds of trial and error, I came up with this _ugly_ solution. Let me know on how can this be resolved in a much better way(If not, then probably - I will add a comment to this code)

src/renderer/state.ts
```diff
@@ -157,6 +165,21 @@ export class AppState {
...
+  // TODO(aryanshridhar): Docs recommends not to use this in production!
+  public baseVersions: BaseVersions = new BaseVersions(getElectronVersions());
+
+  // For managing downloads and versions for electron
+  public installer: Installer = new Installer({
+    electronDownloads: ELECTRON_DOWNLOAD_PATH,
+    electronInstall: ELECTRON_INSTALL_PATH,
+  });
+
+  public runner: Promise<Runner> = Runner.create({
```
aryanshridhar, line 178: This `runner` variable can be confused as an instance of [`Runner`](https://github.com/electron/fiddle/blob/440a94b4f833b077bf576022960d4fe737f04382/src/renderer/runner.ts#L30) class. Open to suggestions for better naming for this variable :) (Probably, `coreRunner`?)

aryanshridhar:
Added a couple of comments above that I felt can be improved. Would love opinions upon it :)

src/renderer/state.ts
```diff
@@ -530,19 +577,55 @@ export class AppState {
...
+      await this.installer.install(version);
+      return;
+    }
+
     console.log(`State: Downloading Electron ${version}`);
-    await setupBinary(
-      ver,
-      this.electronMirror.sources[this.electronMirror.sourceType],
-    );
+    // Fix the Checksum error
```
dsanders11, line 600: Is this comment meant to be a TODO? I don't see the context here.

src/renderer/components/dialog-add-version.tsx
```diff
@@ -9,15 +9,16 @@ import {
 import { observer } from 'mobx-react';
 import * as path from 'path';
 import * as React from 'react';
+import * as fs from 'fs-extra';
```
dsanders11, line 12: Since these are already in alphabetical order, let's keep them that way and move this up to be in alphabetical order.

src/renderer/components/settings-electron.tsx
```diff
@@ -320,7 +320,7 @@ export const ElectronSettings = observer(
       if (state === VersionState.downloading) {
         icon = 'cloud-download';
         humanState = 'Downloading';
-      } else if (state === VersionState.unknown) {
+      } else if (state === VersionState.missing) {
         // The only way for a local version to be unknown
```
dsanders11, line 324:
```suggestion
        // The only way for a local version to be missing
```

src/renderer/transforms/forge.ts
```diff
@@ -1,8 +1,10 @@
 import { execSync } from 'child_process';
 
 import { Files, PACKAGE_NAME } from '../../interfaces';
-import { getElectronBinaryPath } from '../binary';
+import * as path from 'path';
```
dsanders11, line 4: Let's keep this in the grouping at the top with `child_process`.

dsanders11 (commented):
I've done a first pass at review but haven't gone that deep yet. I've left inline comments, and here's a couple top-level comments:

* I'm seeing `yarn.lock` change when running `yarn install`, I think there's a couple dependencies that you missed in the updating of that file in this PR
* If I click "Download" on a version under "Settings", when I close "Settings" the version run button at the top of Fiddle is showing "Checking Status" and spins indefinitely, until I choose a different version from the selector.

src/interfaces.ts
```diff
@@ -3,10 +3,11 @@ export type Files = Map<string, string>;
...
 
 export enum VersionState {
-  ready = 'ready',
+  missing = 'missing',
   downloading = 'downloading',
-  unzipping = 'unzipping',
-  unknown = 'unknown',
+  downloaded = 'downloaded',
+  unzipping = 'installing',
+  ready = 'installed',
```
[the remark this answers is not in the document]
VerteDinde: I agree with this - the strings in the enum should match their string values

src/renderer/runner.ts
```diff
@@ -350,32 +373,23 @@ export class Runner {
...
-    console.log(`Runner: Binary ${binaryPath} ready, launching`);
-
+  private async runFiddle(params: RunFiddleParams): Promise<RunResult> {
+    const { localPath, isValidBuild, version, dir } = params;
+    const { runner, pushOutput, flushOutput } = this.appState;
+    const fiddleRunner = await runner;
     const env = this.buildChildEnvVars();
 
-    // Add user-specified cli flags if any have been set.
-    const options = [dir, '--inspect'].concat(executionFlags);
```
VerteDinde, line 374: Any reason why this was removed, or was this re-added in a different area? We do want to have the ability to inject cli flags here if we can

src/renderer/state.ts
```diff
@@ -157,6 +165,21 @@ export class AppState {
   private readonly defaultVersion: string;
   public appData: string;
 
+  // Populating versions in fiddle-core
+  // TODO(aryanshridhar): Docs recommends not to use this in production!
+  public baseVersions: BaseVersions = new BaseVersions(getElectronVersions());
```
VerteDinde, line 170: Calling out the TODO here: what do we not want to use in production? Is there anything we need to change here before we could ship this in Fiddle?

src/renderer/state.ts
VerteDinde: Depending on the functionality of the other existing class, something like `versionRunner` may work

src/renderer/state.ts
```diff
@@ -271,6 +295,16 @@ export class AppState {
     this.removeAcceleratorToBlock = this.removeAcceleratorToBlock.bind(this);
     this.hideChannels = this.hideChannels.bind(this);
     this.showChannels = this.showChannels.bind(this);
+    this.changeRunnableState = this.changeRunnableState.bind(this);
+
+    // Populating the current state of every version present
+    versions.forEach((ver: RunnableVersion) => {
+      // The local electron builds `state` are already setup in verions.ts
```
VerteDinde, line 302: Nit:
```suggestion
      // A local electron build's `state` is setup in versions.ts
```

src/renderer/state.ts
```diff
@@ -514,8 +554,15 @@ export class AppState {
         console.log(`State: Version ${version} already removed, doing nothing`);
       }
     } else {
-      if (state === VersionState.ready) {
-        await removeBinary(ver);
+      if (state === VersionState.ready || state == VersionState.downloaded) {
+        await this.installer.remove(version);
+        if (this.installer.state(version) === 'missing') {
```
VerteDinde, line 559:
```suggestion
        if (this.installer.state(version) === VersionState.missing) {
```

src/renderer/state.ts
```diff
@@ -831,6 +914,33 @@ export class AppState {
     window.location.hash = hash;
   }
 
+  /**
+   * Maps the InstallState to Versionstate enum
+   * This lives temporary here until fiddle-core exports
+   * the same type.
```
VerteDinde, line 920: Is this dependent on a new release of fiddle-core, or does this still need to be changed in fiddle-core?

src/renderer/state.ts
```diff
@@ -831,6 +914,33 @@ export class AppState {
...
+      case 'downloaded':
+        return VersionState.downloaded;
+      case 'downloading':
+        return VersionState.downloading;
+      case 'installed':
+        return VersionState.ready;
+      case 'installing':
+        return VersionState.unzipping;
+      case 'missing':
+        return VersionState.missing;
```
VerteDinde, line 937: Same comment as David had above, let's get these string values and key aligned.

VerteDinde (changes requested):
Rough first pass review, left a few comments.

src/renderer/state.ts
aryanshridhar: This is because `BaseVersion` is [documented](https://github.com/electron/fiddle-core/blob/main/src/versions.ts#L84) to be a only used while testing.
Our use case here - If fiddle-core is not populated with versions, it tries to fetches the version from [here](https://github.com/electron/fiddle-core/blob/main/src/versions.ts#L203). This causes a couple of errors - 
* We already have this code and electron version are fetched in fiddle prior.
* Since, it uses [node-fetch](https://github.com/electron/fiddle-core/blob/main/src/versions.ts#L4), it throws an error when this function is called here. (The error that I mentioned [here](https://electronhq.slack.com/archives/C03G7RWF6KH/p1657741854218999))

src/renderer/components/version-select.tsx
```diff
@@ -89,14 +92,18 @@ export function getItemLabel({ source, state, name }: RunnableVersion): string {
...
     case VersionState.ready:
+    case VersionState.downloaded:
       return 'saved';
     case VersionState.downloading:
       return 'cloud-download';
     case VersionState.unzipping:
       return 'compressed';
+    // Ideally the code execution shoudln't reach at this point.
+    default:
+      return 'cloud';
```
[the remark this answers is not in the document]
aryanshridhar: Oh yup, right! Thanks :)

src/interfaces.ts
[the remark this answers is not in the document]
aryanshridhar: Oh Cool! I felt it will be more intuitive for the user. (As `installing` is not much intuitive signifying that the version is actually being getting unzipped)

src/renderer/state.ts
aryanshridhar: Oh yeah! This is to be removed :)

src/renderer/components/dialog-add-version.tsx
aryanshridhar: How about we introduce the [sort-import ](https://eslint.org/docs/latest/rules/sort-imports)rule in eslint rather than doing it manually everytime ?

src/renderer/transforms/forge.ts
aryanshridhar: Same comment as mentioned here.
Maybe its better if I put up a thread on slack regarding it :)

src/renderer/state.ts
aryanshridhar: I added this because it this can be integrated in fiddle-core and then imported here.
(We are passing the states as string in fiddle-core, [one such example](https://github.com/electron/fiddle-core/blob/e22ae4d453ba21c5d40f47b153105f632a624c9a/src/installer.ts#L117)).

aryanshridhar: Local branch changes - (https://github.com/electron/fiddle-core/compare/main...aryanshridhar:fiddle-core:Add-VersionStateEnum?expand=1)

aryanshridhar: I have opened up a pr for it - https://github.com/electron/fiddle-core/pull/33
Once this is merged, we don't need to have this extra check.

src/renderer/runner.ts
aryanshridhar: Yup, nice catch! Missed that!

aryanshridhar:
Updates on the pr - 

> If I click "Download" on a version under "Settings", when I close "Settings" the version run button at the top of Fiddle is showing "Checking Status" and spins indefinitely, until I choose a different version from the selector.

* Debugged this and turns out fiddle-core blocks the installation of other versions if one is already installing (https://github.com/electron/fiddle-core/pull/35)
* Rebased the pr so now the imports are always sorted :)
* Removed the VersionState interface to avoid deduplication of code (https://github.com/electron/fiddle-core/pull/33). We can now directly import the interface from fiddle-core and use it.

Also, would love reviews on [this](https://github.com/electron/fiddle/pull/1134#discussion_r935013423) comment!

aryanshridhar:
Rebased and now the checks are passing :rocket: (A pr to fiddle-core upgrading `get` is only left out)

src/renderer/state.ts
erickzhao: Resolving as we said it was okay in our weekly mentor meeting. 👍

src/interfaces.ts
```diff
@@ -1,14 +1,9 @@
+import { InstallState } from '@aryanshridhar/fiddle-core';
```
VerteDinde, line 1: One merge-blocker - testing with a fork of `fiddle-core` is perfectly fine, but we need to cut a new release and use the real `fiddle-core` package before we merge this into Fiddle's main branch.

aryanshridhar: Yup right! The main branch of `electron/fiddle-core` has all the right set of APIs required. We just need to release it :)

src/renderer/runner.ts
```diff
@@ -182,7 +194,10 @@ export class Runner {
       return RunResult.INVALID;
     }
 
-    const isReady = getIsDownloaded(version, localPath);
+    const isReady =
+      state === InstallState.installed ||
+      state === InstallState.downloaded ||
```
erickzhao, line 199: question: It's been a while since I've taken a look at `fiddle-core` code. What is the difference between `InstallState.installed` and `InstallState.downloaded` if they are both considered ready to run here?

erickzhao (changes requested):
First pass with a single blocking change request

src/renderer/runner.ts
aryanshridhar: `downloaded` is the state when the binary is downloaded and kept as a `zip` file whereas the `installed` is when the downloaded binary is currently unzipped for the fiddle to run (`fiddle-core` doesn't unzip all downloaded versions to improve disk space usage)
In our use case - We want both the states to be marked `ready` since `fiddle-core.spawn` by default unzips the version if it's not installed.

src/renderer/runner.ts
```diff
@@ -182,7 +194,10 @@ export class Runner {
...
+      isValidBuild;
```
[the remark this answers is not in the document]
aryanshridhar: Nope, I can't think of a such case where they will be different.
This is because - everytime the version is changed, it triggers [this](https://github.com/aryanshridhar/fiddle/blob/Integrate-fiddlecore/src/renderer/state.ts#L373-L376) event which updates the state of `currentElectronVersion`.

src/interfaces.ts
```diff
@@ -1,14 +1,9 @@
...
+
 export type Files = Map<string, string>;
 
 export type FileTransform = (files: Files) => Promise<Files>;
 
-export enum VersionState {
-  ready = 'ready',
-  downloading = 'downloading',
-  unzipping = 'unzipping',
-  unknown = 'unknown',
```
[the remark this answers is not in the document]
aryanshridhar: Yeah, What `fiddle` does is similar to what `fiddle-core` does.
We mark versions to be `missing` if they are either [removed](https://github.com/electron/fiddle-core/blob/7d17028513f853abbe5eb3a6ebf99d326252e1ef/src/installer.ts#L190-L192) by user or not being currently [`downloaded/installed`](https://github.com/electron/fiddle-core/blob/main/src/installer.ts#L82-L84).

src/renderer/state.ts
```diff
@@ -529,19 +578,61 @@ export class AppState {
...
+      await this.installer.install(version);
+      return;
+    }
+
     console.log(`State: Downloading Electron ${version}`);
-    await setupBinary(
-      ver,
-      this.electronMirror.sources[this.electronMirror.sourceType],
-    );
+    await this.installer.install(version, {
```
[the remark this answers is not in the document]
aryanshridhar: Debugged this and turns out that the issue is way deep! Probably with the new `get` version.
(I will open a new thread regarding it in slack)

src/renderer/constants.ts
```diff
@@ -1,6 +1,16 @@
...
 
 export const USER_DATA_PATH = window.ElectronFiddle.appPaths.userData;
+export const ELECTRON_DOWNLOAD_PATH = path.join(
+  USER_DATA_PATH,
+  '.electron-bin',
+  'zips',
+);
+export const ELECTRON_INSTALL_PATH = path.join(
+  USER_DATA_PATH,
+  '.electron-bin',
```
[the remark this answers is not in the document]
aryanshridhar: Oh yeah! However, we won't be still able to use the existing downloaded versions as `fiddle-core` keeps them in a zip file and also uses a different naming convention of files.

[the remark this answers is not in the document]
erickzhao: That's a big breaking change for users (especially if they have 10+ Electron versions downloaded), and I don't think I'm super comfortable releasing this as-is.

Is there a way we can either:
1. Run a startup script that moves Fiddle scripts to a format that fiddle-core understands?
2. Make changes to Fiddle Core so it understands folders rather than zips?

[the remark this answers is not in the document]
VerteDinde: I agree with Erick, this is a pretty major change for existing Fiddle users. Of the two options, could we modify fiddle-core to understand both zip and folder structures? I feel like that would be the more future-proof option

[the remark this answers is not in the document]
BlackHole1: I also agree with the second.
The first one will be more costly, if the user downloads very many versions of the electron, the migration script will be very slow to execute and will predictably have a lot of error handling logic. And with a lot of unit tests to go along with it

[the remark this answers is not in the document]
BlackHole1: One more thing to consider: when using fiddle to load a locally built electron, the electron is a folder, not a zip file.

image

VerteDinde (approved):
Great work! Approved, contingent on changing the fiddle-core version from the fork to the main package 👍

erickzhao (approved):

ckerr:
So cool that this is in `main` now. Thank you @aryanshridhar, this is great! :heart:

PR #1137, opened 2022-07-19
fix: check for static_dir path at runtime, not build time

VerteDinde:
We're currently checking the __dirname for webpack at build time - we actually need the value that's generated at runtime.

This PR expands on our definePlugin method to use the runtime value (h/t to @codebytere for the find!)


codebytere (approved):

VerteDinde:
This is causing a separate error on when trying to boot up Fiddle, do not merge as is.

PR #1143, opened 2022-08-03
chore: improve main webpack bundle

BlackHole1:
In the previous configuration, we would have packed some dependencies into `index.js`, but this was not really necessary. This is because node_modules will be present in both the development and application environments.

#### before:
image

#### now:
image

#### before:
image

#### now:
image

----

#### now file content:
image

## Problems

- [ ] `node_modules` is empty in pack 😶‍🌫️


BlackHole1:
It looks like this improvement is unnecessary. This is because it is an expected feature of `forge` (node_modules is empty)

PR #1147, opened 2022-08-04
build: enforce sort-imports eslint rule

aryanshridhar:
Adds the `sort-import` rule in codebase.

<strong>Summary -</strong>
* Added the [sort-import](https://eslint.org/docs/latest/rules/sort-imports) rule in eslint. However, it lacks the feature of auto fixing changes using `eslint --fix`. Hence, this is where [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import) helps us! (It also provides better  control on how imports are sorted)
* This requires `eslint-import-resolver-typescript` to resolve imports in typescript files.
* Added a new command `lint:js` which sort the imports in `.js` files.
* Ran `yarn lint` which autofixes all the problems for us :)

<strong>Rules kept off -</strong>
* [import/no-named-as-default-member](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-as-default-member.md)
* [import/namespace](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/namespace.md)

Edit: This can be merged once large pending pull requests are merged first - to avoid large merge conflicts :)


src/main/main.ts
```diff
@@ -1,21 +1,22 @@
-import { initSentry } from '../sentry';
-initSentry();
```
BlackHole1, line 2: The sentry should be left as it is, the sooner the better

src/renderer/components/editors-non-ideal-state.tsx
```diff
@@ -1,6 +1,5 @@
-import * as React from 'react';
-
 import { Button, NonIdealState } from '@blueprintjs/core';
+import * as React from 'react';
```
BlackHole1, line 2: Is it possible to have `React` always in the first row? with the highest priority

BlackHole1 (changes requested):

BlackHole1:
It would have been better if some details had been optimized. 😀

src/renderer/components/editors-non-ideal-state.tsx
erickzhao: I think that might require a separate plugin here: https://github.com/lydell/eslint-plugin-simple-import-sort

aryanshridhar: We can do this by defining custom `pathGroups` and ignoring the pattern then - [docs](https://github.com/import-js/eslint-plugin-import/blob/5fe9aa44eb9d20b376d1b32a5d3384c46a01309e/docs/rules/order.md#pathgroupsexcludedimporttypes-array)
```js
{
  "import/order": [
    "error",
    {
      "pathGroups": [
        {
          "pattern": "react",
          "group": "builtin",
          "position": "before"
        }
      ],
      "pathGroupsExcludedImportTypes": ["react"]
    }
  ]
}
```

This is how the sorting order looks like then - 
```js
import * as React from 'react';

import { Button, NonIdealState } from '@blueprintjs/core';

import { EditorMosaic } from '../editor-mosaic';

export function renderNonIdealState(editorMosaic: EditorMosaic) {
```

aryanshridhar: > I think that might require a separate plugin here: https://github.com/lydell/eslint-plugin-simple-import-sort

I thought of using this first until I read [this para](https://github.com/lydell/eslint-plugin-simple-import-sort#not-for-everyone) in its docs. (I thought it's better to have more control on how we sort the imports)
Quoting a para from its docs - 
> If you want more options, I recommend using the [import/order](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md) rule (from [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import/)) instead. It has plenty of options, and the maintainers seem interested in expanding the feature where it makes sense.

src/main/main.ts
aryanshridhar: Oh Cool! Wasn't aware of it :)

BlackHole1 (approved):
LGTM. 👍

PR #1153, opened 2022-08-06
refactor: improve utils code

BlackHole1:
Improve readability and maintainability of code

### DIFF
--- a/src/utils/electron-name.ts
+++ b/src/utils/electron-name.ts
@@ -6,9 +6,11 @@
 export function getElectronNameForPlatform(): string {
   if (process.platform === 'win32') {
     return 'electron.exe';
-  } else if (process.platform === 'darwin') {
-    return 'Electron.app';
-  } else {
-    return 'electron';
   }
+
+  if (process.platform === 'darwin') {
+    return 'Electron.app';
+  }
+
+  return 'electron';
 }
--- a/src/utils/electron-name.ts
+++ b/src/utils/electron-name.ts
@@ -4,13 +4,15 @@
  * @returns {string}
  */
 export function getElectronNameForPlatform(): string {
-  if (process.platform === 'win32') {
-    return 'electron.exe';
+  switch (process.platform) {
+    case 'win32': {
+      return 'electron.exe';
+    }
+    case 'darwin': {
+      return 'Electron.app';
+    }
+    default: {
+      return 'electron';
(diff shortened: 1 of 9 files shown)


VerteDinde (approved):

src/utils/electron-name.ts
```diff
@@ -6,9 +6,11 @@
...
-  } else if (process.platform === 'darwin') {
+  }
+
+  if (process.platform === 'darwin') {
     return 'Electron.app';
-  } else {
-    return 'electron';
   }
+
+  return 'electron';
```
malept, line 15: suggestion: this could just be a switch statement.

src/utils/exec.ts
```diff
@@ -23,28 +46,11 @@ export async function exec(dir: string, cliArgs: string): Promise<string> {
...
         if (error) {
-          reject(error);
+          return reject(error);
         }
 
-        resolve(typeof result === 'string' ? result : `${result}`);
+        resolve(String(result));
       },
     );
   });
```
malept, line 55: suggestion: if we're generally refactoring for clarity in this PR, it seems like you could just `promisify` the exec call.

src/utils/gist.ts
```diff
@@ -13,17 +13,9 @@
...
-  if (input.startsWith('https://gist.github.com')) {
-    if (input.endsWith('/')) {
-      input = input.slice(0, -1);
-    }
-    id = input.split('/').pop()!;
-  }
-
-  return id.match(/[0-9A-Fa-f]{32}/) ? id : null;
+  return id?.[0] || null;
 }
```
malept, line 19: question: do we have unit tests for this function?

BlackHole1: Yes. Here: https://github.com/electron/fiddle/blob/3814ec9be6e432eca26eaade778efd15d0968c48/tests/utils/gist-spec.ts#L41-L85

BlackHole1:
@malept Done. Please re-review

src/utils/gist.ts
malept: It'd be nice if the null case(s) were covered in the testsuite.

BlackHole1: Good idea.

BlackHole1: Done

PR #1160, opened 2022-08-11
build(deps): upgrade to blueprintjs@4

aryanshridhar:
Upgrades `blueprintjs` to v4 (4.8.0)
This also upgrades the following packages with it - 
* @blueprintjs/popover2@1.5.1
* @blueprintjs/select@4.5.3

to make sure that the codebase relies on only v4 version of `blueprintjs/core` - 
```
❯ yarn why @blueprintjs/core
yarn why v1.22.19
warning ../../package.json: No license field
[1/4] 🤔  Why do we have the module "@blueprintjs/core"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@blueprintjs/core@4.8.0"
info Has been hoisted to "@blueprintjs/core"
info Reasons this module exists
   - Specified in "dependencies"
   - Hoisted from "@blueprintjs#popover2#@blueprintjs#core"
   - Hoisted from "@blueprintjs#select#@blueprintjs#core"
info Disk size without dependencies: "9.64MB"
info Disk size with unique dependencies: "61.88MB"
info Disk size with transitive dependencies: "69.61MB"
info Number of shared dependencies: 34
✨  Done in 0.75s.
```


aryanshridhar:
Quite funny to see [this test](https://github.com/electron/fiddle/blob/main/tests/renderer/components/version-select-spec.tsx#L94-L107) fail here.
When logged out the html it clearly contains the `.disabled-menu-tooltip` class and surprisingly enzyme is not able to find it
Here's the logged html -
```
  console.log
    <span class="disabled-menu-tooltip bp4-popover2-target"><li class="" role="none"><a role="menuitem" tabindex="-1" label="Not downloaded" icon="cloud" class="bp4-menu-item bp4-active bp4-disabled bp4-selected"><span class="bp4-menu-item-icon"><span icon="cloud" aria-hidden="true" tabindex="-1" class="bp4-icon bp4-icon-cloud"><svg data-icon="cloud" width="16" height="16" viewBox="0 0 16 16" role="img"><path d="M12 6c-.03 0-.07 0-.1.01A5 5 0 002 7c0 .11.01.22.02.33A3.51 3.51 0 000 10.5C0 12.43 1.57 14 3.5 14H12c2.21 0 4-1.79 4-4s-1.79-4-4-4z" fill-rule="evenodd"></path></svg></span></span><div class="bp4-fill bp4-text-overflow-ellipsis">1.0.0</div><span class="bp4-menu-item-label">Not downloaded</span></a></li></span>
```
Not sure what's wrong here :thinking:

georgexu99:
Hey! I took a look at the failing test- in the interest of moving the PR forward, we can write the test using a snapshot instead as it seems that the wrapper is pulling some funnys ;)

We should rewrite the next test for the same reason- the wrapper may be giving us a false positive.

Here's a gist describing what I'm [thinking](https://gist.github.com/georgexu99/afd39267213fc7e4ef91568542aa8a9d). This would be similar to how we test menu items in the [general appearance settings](https://github.com/electron/fiddle/blob/8cbc9c998153d512da28413204a3d5618aa36dd7/tests/renderer/components/settings-general-appearance-spec.tsx#L228)

dsanders11:
@georgexu99, that seems reasonable. Could you open a PR for just that change? This one is pretty big and hairy to review, so would be easier to land that generic test refactor separately.

aryanshridhar:
Actually, I already have these changes in my local branch - https://github.com/aryanshridhar/fiddle/tree/Add-VersionSelectSnapshots
We can pick up from there if that seems fine :)

aryanshridhar:
Rebased @dsanders11.
However, as this pr got really old (lot of merge conflicts); Manually testing the application now to check if everything works as desired.

Issue #1176, opened 2022-09-19
RFC: Add setting for publishGistAsRevision


georgexu99:
**problem**: it is hard to spot the differences between the fiddle default gist and bug repros.

**soln**:  add checkbox in setting to toggle publishing gists as revisions. 

Github automatically shows diffs of previous uploads. We can leverage this by uploading the base fiddle template, then uploading the repro code as a gist update, rather than the initial publish.

georgexu99:
exploring implementation (can't self-assign due to lack of perms)

georgexu99:
closed with https://github.com/electron/fiddle/pull/1177

PR #1177, opened 2022-09-20
feat: add publishGistAsRevision()

georgexu99:
https://github.com/electron/fiddle/issues/1176

This PR adds a checkbox to the Github settings to allow users to select whether they want to publish their gist as a revision of the default Fiddle startup editor values. This is useful when users submit issues, with Fiddle gists as repro examples: this change will allow maintainers to quickly spot diffs easier.

✅  Added option in settings to include checkbox
✅    Added tests to expand coverage (wip)
✅  Validated on Mac, (needs to be tested on Windows and Linux)

ezgif com-gif-maker (1)

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -140,12 +140,12 @@ export const GistActionButton = observer(
           files: gistFilesList as any, // Note: GitHub messed up, GistsCreateParamsFiles is an incorrect interface
         });
 
+        appState.gistId = gist.data.id;
+        appState.localPath = undefined;
+
         if (appState.isPublishingGistAsRevision) {
           await this.handleUpdate(appState.isPublishingGistAsRevision);
         }
-
-        appState.gistId = gist.data.id;
-        appState.localPath = undefined;
 
         console.log(`Publish Button: Publishing complete`, { gist });
         this.renderToast({
@@ -203,10 +203,7 @@ export const GistActionButton = observer(
       const values = await window.ElectronFiddle.app.getEditorValues(options);
 
       appState.activeGistAction = GistActionState.updating;
-      console.log('hi i am here');
-      await new Promise((r) => setTimeout(r, 5000));
 
-      console.log('hi i am here 2');
       try {
         const {
           data: { files: oldFiles },
@@ -217,6 +214,8 @@ export const GistActionButton = observer(
           // Gist files are deleted by setting content to an empty string.
           if (!(id in files)) files[id] = { content: '' };
         }
+
+        console.log({ gist_id: appState.gistId! });
 
         const gist = await octo.gists.update({
           gist_id: appState.gistId!,
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -22,8 +22,10 @@ import {
 import { IpcEvents } from '../../ipc-events';
 import { ensureRequiredFiles } from '../../utils/editor-utils';
 import { getOctokit } from '../../utils/octokit';
+import { getTemplate } from '../content';
 import { ipcRendererManager } from '../ipc';
 import { AppState } from '../state';
+// import { FIDDLE_GIST_DESCRIPTION_PLACEHOLDER } from '../constants';
 
 interface GistActionButtonProps {
   appState: AppState;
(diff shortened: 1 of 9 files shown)


tests/renderer/components/commands-publish-button-spec.tsx
```diff
@@ -124,6 +124,7 @@ describe('Action button component', () => {
     beforeEach(() => {
       // create a button that's primed to publish a new gist
       ({ instance } = createActionButton());
+      state.isPublishingGistAsRevision = false;
```
georgexu99, line 127: I opted not to add a test for when `isPublishingGistAsRevision = true` since we mock out octokit- let me know if it's worth it to find a way to see if updateGist() is called with the right params when publishGist() is called with `isPublishingGistAsRevision = true`

georgexu99: update: made a test 😛

src/renderer/components/commands-action-button.tsx
```diff
@@ -120,15 +121,26 @@ export const GistActionButton = observer(
...
       try {
+        const gistFilesList = appState.isPublishingGistAsRevision
+          ? this.gistFilesList(defaultGistValues)
+          : this.gistFilesList(currentEditorValues);
+
         const gist = await octo.gists.create({
           public: !!gitHubPublishAsPublic,
           description,
-          files: this.gistFilesList(values) as any, // Note: GitHub messed up, GistsCreateParamsFiles is an incorrect interface
+          files: gistFilesList as any, // Note: GitHub messed up, GistsCreateParamsFiles is an incorrect interface
```
BlackHole1, line 137: Add `TODO`. When octo fixes this problem, we can remove `as any`

src/renderer/components/settings-general-github.tsx
```diff
@@ -61,17 +64,46 @@ export const GitHubSettings = observer(
...
+
     public render() {
       const { gitHubToken } = this.props.appState;
+      const { isPublishingGistAsRevision } = this.props.appState;
 
       const maybeSignedIn = !!gitHubToken
         ? this.renderSignedIn()
         : this.renderNotSignedIn();
 
+      const publishGistAsRevisionInstructions = `
```
BlackHole1, line 88: Maybe use `private static publishGistAsRevisionInstructions = ?` will better. e.g.

```ts
private static publishGistAsRevisionInstructions = `
        Enable this option to always publish your fiddle as a revision of the
        default fiddle gist values.`.trim();
```

src/renderer/state.ts
```diff
@@ -219,6 +222,7 @@ export class AppState {
       isInstallingModules: observable,
       isKeepingUserDataDirs: observable,
       isOnline: observable,
+      isPublishingGistAsRevision: observable,
```
erickzhao, line 225: IMO we should make this true by default since it's a low-impact change.

src/renderer/components/commands-action-button.tsx
```diff
@@ -120,15 +121,27 @@ export const GistActionButton = observer(
...
+        // TODO: remove as any when octo is fixed
         const gist = await octo.gists.create({
           public: !!gitHubPublishAsPublic,
           description,
-          files: this.gistFilesList(values) as any, // Note: GitHub messed up, GistsCreateParamsFiles is an incorrect interface
+          files: gistFilesList as any, // Note: GitHub messed up, GistsCreateParamsFiles is an incorrect interface
         });
 
+        if (appState.isPublishingGistAsRevision) {
+          this.handleUpdate(appState.isPublishingGistAsRevision);
```
erickzhao, line 142: issue: We should `await` this function

src/renderer/components/commands-action-button.tsx
```diff
@@ -180,7 +193,7 @@ export const GistActionButton = observer(
     /**
      * Update an existing GitHub gist.
      */
-    public async handleUpdate() {
+    public async handleUpdate(isPublishingGistAsRevision = false) {
```
erickzhao, line 196: nit: we could rename this parameter as `silent` to better understand what it's supposed to do as opposed to why we're doing it.

src/renderer/components/commands-action-button.tsx
```diff
@@ -120,15 +121,27 @@ export const GistActionButton = observer(
...
+      const defaultGistValues = await getTemplate(appState.version);
+      const currentEditorValues = await window.ElectronFiddle.app.getEditorValues(
+        options,
+      );
 
       try {
+        const gistFilesList = appState.isPublishingGistAsRevision
+          ? this.gistFilesList(defaultGistValues)
+          : this.gistFilesList(currentEditorValues);
+
```
erickzhao, line 133: issue(blocking): When I was testing this PR out, the `package.json` being written to the gist to appear as a "revision", which seems wrong.

See https://gist.github.com/erickzhao/b4bba605a5a4af69fe57380c8470ed32/revisions

erickzhao (changes requested):

src/renderer/state.ts
georgexu99: I agree! The setting button is default checked to equal true

src/renderer/components/commands-action-button.tsx
```diff
@@ -199,21 +215,26 @@ export const GistActionButton = observer(
           if (!(id in files)) files[id] = { content: '' };
         }
 
+        console.log({ gist_id: appState.gistId! });
```
erickzhao, line 218: oops
```suggestion
```

erickzhao (approved):
approved minus the log line i added by accident

codebytere (approved):
nice 🌟

Issue #1083, opened 2022-05-26
Hover labels on file tree show up when settings are open


carolstran:
👋🏼  I noticed when tabbing to your General Settings page that you can't navigate to the actual settings form. Instead, you end up tabbing through the elements on the main editor page. So then if you stop... you sometimes end up with funny floating labels like this:

![Settings page with a floating Toggle Visibility hover label in the middle.](https://user-images.githubusercontent.com/26869552/170574315-161f4ded-9785-469a-821d-612c29a4bbc8.png)

I'm guessing it's an issue with the HTML, i.e. the main editor content should be hidden semantically while the settings are open or the settings should be injected at the top of the HTML file, etc.

carolstran:
Same thing happens for the modules search input: 

![Floating module search input on the settings page.](https://user-images.githubusercontent.com/26869552/170575568-018f3cf9-0ba0-4e87-ac16-78a07290a77c.png)

dsanders11:
I believe this was fixed in #1181, doesn't repro for me any longer.

PR #1181, opened 2022-10-01
fix: tab focus for settings page

miqh:
Hello there,

Thought I'd try look into the issue (#1083) with tab focus misbehaving when the settings page is open.

I can confirm those tooltips come about because tabbing through the settings page eventually runs out of elements and goes on to tab through elements that are part of either the header or editor area.

The least invasive solution I could think of is to ensure the header and editor area elements are made invisible. `display: none` is a bit more destructive than `visibility: hidden`, because the former may interfere with components not rendering at all and potentially losing their state.

Now, I'm not awfully familiar with MobX, but I tried my best to work in some changes that do what I described above.

Happy to take feedback and revise (or add a test, which I couldn't work out how to do for this) as necessary. 🙂

---

As an aside, I was curious why tabbing inside those overlay dialogs worked better, so I also looked there at what's going on. It seems to be that the underlying Blueprint dialog implementations have some special code which ensure tabbing only cycles through within the overlay.

### DIFF
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -63,3 +63,7 @@ body {
   text-align: center;
   font-size: 12px;
 }
+
+.tabbing-hidden {
+  visibility: hidden;
+}
--- a/src/renderer/components/header.tsx
+++ b/src/renderer/components/header.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 
+import { reaction } from 'mobx';
+
 import { AppState } from '../state';
 import { Commands } from './commands';
 import { WelcomeTour } from './tour-welcome';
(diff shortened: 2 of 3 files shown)


miqh:
@erickzhao, funny you should ask that...

I actually started out making changes in `app.tsx` to apply the CSS classes there because it's obviously a more central choke point. However, I wasn't able to figure out how to do it.

I think something needs to be done with that plain old JSX element being passed into `ReactDOM.render()` to make it aware and observing the `isSettingsShowing` property part of application state. My MobX-fu and understanding isn't up to scratch on how best to do this I'm afraid. All I know is that it's probably not wise to be conditionally rendering `Header` or `OutputEditorsWrapper` for fear that might do strange things to application state.

If you can point me in the right direction I can try have another stab, otherwise I'll just close these changes.

codebytere:
@erickzhao do we want to move forward here?

erickzhao (approved):
I think this solves a real problem and I don't particularly mind the approach!

PR #1185, opened 2022-10-09
test: use snapshots to properly test version-select component

aryanshridhar:
https://github.com/electron/fiddle/pull/1160#issuecomment-1259827500

### DIFF
--- a/tests/renderer/components/__snapshots__/version-select-spec.tsx.snap
+++ b/tests/renderer/components/__snapshots__/version-select-spec.tsx.snap
@@ -1,5 +1,90 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`VersionSelect component disableDownload disables download buttons when return value is true 1`] = `
+<Blueprint3.Popover
+  autoFocus={false}
+  boundary="scrollParent"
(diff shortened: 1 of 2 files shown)


aryanshridhar:
cc @georgexu99 @dsanders11 :)

georgexu99:
The test looks good, but I'm thinking that we should put this test with the rest of the PR? @dsanders11 

Also, perhaps this is a bit of an unfounded worry, but given the size of the other PR and its age, it may honestly be easier reapply the changes on top of main (with a careful cherry-pick or by hand), unless merging the new main changes don't pose a huge problem

erickzhao (changes requested):
I would lean against replacing the existing assertion with a snapshot test.

IMO this change makes it harder to read what the exact assertion is and makes the test more brittle to any change in the component (easy to get false positives).

dsanders11:
@erickzhao makes a good point about this making the test more brittle. Let's find a different way to fix the test failure in #1160. Sorry about the churn, @aryanshridhar.

aryanshridhar:
No worries :)
Closing in favor of point put up by @erickzhao.

Issue #1121, opened 2022-06-27
[bug] showMe->electron APIs->DesktopCapture not work on electron 19.0.6


terry2010:
![图片](https://user-images.githubusercontent.com/1849037/175857443-6fb7cac9-06c4-4846-893c-2da54e428a78.png)

desktopCapturer move to main process since 17.0.0

BlackHole1:
It looks like we need to re-engineer the architecture of `Show Me` 🤔
Make it work: insert the correct code according to the different electron versions.

erickzhao:
@BlackHole1 I think something might be interesting is to fetch the data from `electron/electron` `docs/fiddles` contents which is versioned by default

PR #1186, opened 2022-10-09
fix(show-me/desktopcapturer): allow example to work for v17 and above

thewheat:
Fixes: #1121

- The current 'Show Me' > DesktopCapturer example does not work with Electron version 17 and above due to https://www.electronjs.org/blog/electron-17-0#highlighted-features
> `desktopCapturer.getSources` is now only available in the main process. [#30720](https://github.com/electron/electron/pull/30720)

- Adding code to support v17 and above while also keeping legacy code and running the appropriate code path by detecting the Electron version

**Demo of v21 and v17 using new code**
image

image

**Demo of v16 using legacy code**
image

### DIFF
--- a/static/show-me/desktopcapturer/main.js
+++ b/static/show-me/desktopcapturer/main.js
@@ -4,7 +4,7 @@
 // For more info, see:
 // https://electronjs.org/docs/api/desktop-capturer
 
-const { app, BrowserWindow } = require('electron')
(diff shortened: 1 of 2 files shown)


BlackHole1 (approved):
LGTM
/cc @erickzhao

MarshallOfSound (approved):

Issue #1187, opened 2022-10-09
Editors list does not scroll - items cut off when there are too many files


thewheat:
When there are too files added, it can cause the top/bottom of the Editors list to get cut off as the section does not seeem to scroll (found while trying to test https://github.com/electron/fiddle/issues/690)

<img width="322" alt="image" src="https://user-images.githubusercontent.com/892961/194761450-90f18405-6144-44e7-aff4-b7f8c1c72c06.png">

Toggling Console can get the top bit to show but now the files at the bottom get cut off
<img width="543" alt="image" src="https://user-images.githubusercontent.com/892961/194761514-af006a44-1de2-4491-b779-8dd889ac2f8c.png">

* * *

Looks like there is an `overflow: hidden` causing this
<img width="1606" alt="image" src="https://user-images.githubusercontent.com/892961/194761570-5f2e9b0e-c99f-4bc7-9c1e-32631f2ddfa3.png">

We can set `overflow: auto` but scrollbar styling is not there
<img width="1502" alt="image" src="https://user-images.githubusercontent.com/892961/194761626-fda5ab4b-820a-426d-892b-898e73a36bfa.png">

BlackHole1:
@thewheat Would you be interested in submitting a PR for this?

thewheat:
Sure I'll try give it a go! Haven't dealt with the UI side of things so may take a bit of digging to get that scrollbar design in

PR #1190, opened 2022-10-13
fix: Allow scrolling in Editors Listing

thewheat:
Fixes: https://github.com/electron/fiddle/issues/1187

Add ability to scroll "Editors" listing by just changing `hidden` to `auto` for simplicity (see details of other approach below which feels a bit much in terms of code changes but if the user experience s more desirable and there are betters ways to improve it, I'll be happy to contiinue investigating with any guidance)

**Preview of change**
image

* * *

**Preview of other approach**
image

- I feel the above better in terms of user experience with the top section being static and only scroll the list of files
- It involves CSS overrides including using `!important` as it is overring Blueprint's Collapse component behaviour https://github.com/palantir/blueprint/blob/a7479b73312afba568a41773060c5302ef4f13b5/packages/core/src/components/collapse/collapse.tsx#L184-L186
- I have a branch with those changes here https://github.com/thewheat/fiddle/pull/new/fix-editors-scroll-css but felt it was doing too much overriding for my liking (am also a bit rusty with CSS in general so maybe there is a better way to do it)

### DIFF
--- a/src/renderer/components/sidebar-file-tree.tsx
+++ b/src/renderer/components/sidebar-file-tree.tsx
@@ -160,7 +160,7 @@ export const SidebarFileTree = observer(
       ];
 
       return (
-        <div style={{ overflow: 'auto' }}>
+        <div className="fiddle-scrollbar">
           <Tree contents={editorsTree} />
         </div>
       );
--- a/src/renderer/components/sidebar-file-tree.tsx
+++ b/src/renderer/components/sidebar-file-tree.tsx
@@ -160,7 +160,7 @@ export const SidebarFileTree = observer(
       ];
 
       return (
-        <div style={{ overflow: 'hidden' }}>
+        <div style={{ overflow: 'auto' }}>
           <Tree contents={editorsTree} />
         </div>
       );
--- a/src/less/components/sidebar.less
+++ b/src/less/components/sidebar.less
@@ -50,3 +50,24 @@
     color: @white;
   }
 }
+
+.fiddle-scrollbar {
+  overflow: auto;
+}
+
+.fiddle-scrollbar::-webkit-scrollbar {
+  width: 8px;
+}
+
+.fiddle-scrollbar::-webkit-scrollbar-thumb {
+  border-radius: 4px;
+  background-color: rgba(172, 172, 172, 0.6);
+
+  .bp3-dark & {
+    background-color: rgba(172, 172, 172, 0.4);
+  }
+}
+
+.fiddle-scrollbar::-webkit-scrollbar-track {
+  background-color: transparent;
+}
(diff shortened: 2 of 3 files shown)


thewheat:
I was also thinking the ideal solution would have scrollbar styled like the editors themselves
But seems that is part [Monaco Editor](https://microsoft.github.io/monaco-editor/) so not something that can be directly used. If there is a way to use it, do let me know and I can try add it in

image

thewheat:
Ah looks like some tests are failing. Let me look into that

BlackHole1:
I support the current approach.

The style is OK under Mac. See:

image

So only the scrollbar needs to be optimized for the Windows system.

> But seems that is part [Monaco Editor](https://microsoft.github.io/monaco-editor/) so not something that can be directly used. If there is a way to use it, do let me know and I can try add it in

Monaco's scrollbars are implemented by div+canvas, so for the time being, you cannot use Monaco's scrollbars.

Anyway. Thank you for your PR!

BlackHole1:
I've been busy lately, so I may not be able to assist you in refining the scrollbar style on Windows for the next few days, but I'll try to find time to help finish this PR!

BhinbahadurUK:
Thanks I noticed a delay also. Check my side maybe I bogged something again
.



Regards

codebytere (approved):
Fine with this as a stopgap!

src/renderer/components/sidebar-file-tree.tsx
```diff
@@ -160,7 +160,7 @@ export const SidebarFileTree = observer(
       ];
 
       return (
-        <div style={{ overflow: 'hidden' }}>
+        <div style={{ overflow: 'auto' }}>
```
BlackHole1, line 163:
```suggestion
        <div style={{classNames('fiddle-scrollbar', isWin && 'win')}}>
```

```css
.fiddle-scrollbar {
  overflow: auto;
}

.win.fiddle-scrollbar::-webkit-scrollbar {
  width: 8px;
}

.win.fiddle-scrollbar::-webkit-scrollbar-thumb {
  border-radius: 4px;
  background-color: rgba(172, 172, 172, 0.4);
}

.win.fiddle-scrollbar::-webkit-scrollbar-track {
  background-color: transparent;
}
```


https://user-images.githubusercontent.com/8198408/196402225-2be47ee5-7d84-4aa8-a9c9-27c696154534.mp4

BlackHole1 (changes requested):

src/renderer/components/sidebar-file-tree.tsx
thewheat: Ah nice! Did a bit of testing and I think the `win` prefix may not be needed as the styling also works on MacOS (and I presume Linux)

I've also [found an existing pattern](https://github.com/electron/fiddle/blob/9edd8f2852c1237aeed379e33a9ae9abc018de86/src/less/components/sidebar.less#L49-L51) to slightly change the scrollbar based on the theme as it seemed a contrast on the light theme was not good

Changes updated and preview below
image

Things are still running through checks and will try address if anything fails

BlackHole1 (approved):

PR #1197, opened 2022-10-18
fix: catch FiddleRunner spawn error

codebytere:
Refs https://github.com/electron/fiddle/pull/1134.

Appropriately handle failure to spawn the runner.

### DIFF
--- a/src/renderer/runner.ts
+++ b/src/renderer/runner.ts
@@ -387,31 +387,44 @@ export class Runner {
     // Add user-specified cli flags if any have been set.
     const options = [dir, '--inspect'].concat(executionFlags);
 
+    const cleanup = async () => {
+      flushOutput();
+
+      this.appState.isRunning = false;
+      this.child = null;
+
+      // Clean older folders
+      await window.ElectronFiddle.app.fileManager.cleanup(dir);
+      await this.deleteUserData();
+    };
+
     return new Promise(async (resolve, _reject) => {
-      this.child = await fiddleRunner.spawn(
-        isValidBuild && localPath ? Installer.getExecPath(localPath) : version,
-        dir,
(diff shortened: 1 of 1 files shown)


src/renderer/runner.ts
```diff
@@ -388,22 +388,30 @@ export class Runner {
...
+        this.child = await fiddleRunner.spawn(
+          isValidBuild && localPath
+            ? Installer.getExecPath(localPath)
+            : version,
+          dir,
+          { args: options, cwd: dir, env },
+        );
+      } catch (e) {
+        pushOutput(`Failed to spawn Fiddle: ${e.message}`);
+        resolve(RunResult.FAILURE);
```
BlackHole1, line 412:
```suggestion
        return resolve(RunResult.FAILURE);
```

BlackHole1 (changes requested):

src/renderer/runner.ts
```diff
@@ -387,31 +387,44 @@ export class Runner {
...
+          dir,
+          { args: options, cwd: dir, env },
+        );
+      } catch (e) {
+        pushOutput(`Failed to spawn Fiddle: ${e.message}`);
+        await cleanup();
+        resolve(RunResult.FAILURE);
+      }
+
       this.appState.isRunning = true;
```
BlackHole1, line 416: Even if the above code triggers a catch, it will still be executed here. So that's what I'm talking about: we need `return`

src/renderer/runner.ts
codebytere: yes, it looks like you saw this right before i force pushed it. It's been fixed and other code changed since then.

BlackHole1 (approved):

PR #1199, opened 2022-10-18
build(deps): switch to @electron/fiddle-core package

dsanders11:
The time has come, there can be only one. ⚔️


VerteDinde (approved):
RIP to `@vertedinde/fiddle-core` 🙏 LGTM, thanks for migrating!

PR #1201, opened 2022-10-20
test: await electronTypes.setVersion() in afterEach

dsanders11:
There's a CI flake in this test somewhere, I'm taking a stab at this being it.

Unfortunately some of the code under test can silently fail (will just `console.debug` a message) so if there are failures happening as a result of this await not happening, you wouldn't see it unless an `expect` fails.

### DIFF
--- a/tests/renderer/electron-types-spec.ts
+++ b/tests/renderer/electron-types-spec.ts
@@ -70,8 +70,8 @@ describe('ElectronTypes', () => {
     nodeTypesData = require('../fixtures/node-types.json');
   });
 
-  afterEach(() => {
-    electronTypes.setVersion();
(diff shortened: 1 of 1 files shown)


codebytere:
@dsanders11 looks like this all passed 🤔

BhinbahadurUK:
Good day


On extract box code what parser can be used to reform data to alpha order
in outer packets to boxed then reforms.

Here's Example

UK Bhinbahadur

dsanders11:
I don't think this solves the flake (after digging into it more) but I think this change is objectively correct so might as well merge it.

erickzhao (approved):

codebytere (approved):

PR #1211, opened 2022-11-29
perf: Improve performance by compressing in parallel ✈️ 

sanjaiyan-dev:
Improving the performance by running the independent `async` functions in parallel ✈️.

_Sorry if I made any mistakes :(_


dsanders11:
I'm going to close this PR since it's gone stale, and I don't think the trade off between code readability and theoretical performance improvement is worth it. Thanks for the PR though!

PR #1218, opened 2022-12-07
fix: don't check package.json version for Electron examples

### LINKED ISSUE #1217: Can't Load Examples From Docs
Loading an example from the docs for Electron versions greater than `v21.1.0` no longer works after https://github.com/electron/electron/pull/36106 since `package.json` doesn't have the `version` bumped. When Fiddle tries to load an example it'll try to set the Electron version to `v0.0.0-development` as a result.

image

dsanders11:
Fixes #1217.

This change means Fiddle will no longer be able to handle URLs like `electron-fiddle://electron/22-x-y/docs/fiddles/features/macos-dark-mode`, but AFAIK there is no use case for those kinds of URLs. If we want to support those going forward there's no clear path after the change in `electron/electron`, other than hardcoded rules around the branch name. So I think it's best to just drop support for those kinds of URLs now and only support tags.

### DIFF
--- a/tests/main/protocol-spec.ts
+++ b/tests/main/protocol-spec.ts
@@ -134,18 +134,18 @@ describe('protocol', () => {
     });
 
     it('handles an electron path url', () => {
-      // electron-fiddle://electron/{ref}/{path}
+      // electron-fiddle://electron/{tag}/{path}
       listenForProtocolHandler();
 
       const handler = (app.on as any).mock.calls[0][1];
 
-      handler({}, 'electron-fiddle://electron/4.0.0/test/path');
+      handler({}, 'electron-fiddle://electron/v4.0.0/test/path');
 
       expect(ipcMainManager.send).toHaveBeenCalledWith<any>(
         IpcEvents.LOAD_ELECTRON_EXAMPLE_REQUEST,
         [
           {
-            ref: '4.0.0',
+            tag: 'v4.0.0',
             path: 'test/path',
           },
         ],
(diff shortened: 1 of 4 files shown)


tests/main/protocol-spec.ts
```diff
@@ -134,18 +134,18 @@ describe('protocol', () => {
...
 
     it('handles an electron path url', () => {
-      // electron-fiddle://electron/{ref}/{path}
+      // electron-fiddle://electron/{tag}/{path}
       listenForProtocolHandler();
 
       const handler = (app.on as any).mock.calls[0][1];
 
-      handler({}, 'electron-fiddle://electron/4.0.0/test/path');
+      handler({}, 'electron-fiddle://electron/v4.0.0/test/path');
```
VerteDinde, line 142: I'm not blocking on this, more thinking out loud - is there any other areas where this `v` prefix may cause issues? Trying to decide if we should just strip this off of the tag name when we fetch and pass through the tag version.

VerteDinde (approved):
One comment about tag prefix to think about, but not blocking

tests/main/protocol-spec.ts
dsanders11: I don't think it will cause any problems - this was actually a case of the tests not reflecting reality, since AFAIK the URL has always included the `v` prefix in practice. In fact if you strip it off and send just the version number without the prefix, the current Fiddle won't work with that URL (see screenshot) since there's no ref on the `e/e` repo for that name. So the tests were covering a case which didn't actually work in practice, since the `octo` response was mocked.

Screen Shot 2022-12-07 at 1 18 02 PM

The logic here before was generic enough that you could use URLs like `electron-fiddle://electron/v20.0.0/`, `electron-fiddle://electron/20-x-y/`, or `electron-fiddle://electron/f20a51a/`. This just drops support for the latter two options since it's no longer possible to easily get a concrete version number from `20-x-y` or `f20a51a`.

PR #1220, opened 2022-12-08
fix: race condition when getting Node types

dsanders11:
Found the race condition that's been haunting the tests.

### DIFF
--- a/src/renderer/electron-types.ts
+++ b/src/renderer/electron-types.ts
@@ -179,19 +179,23 @@ export class ElectronTypes {
       );
     }
 
-    const { files: fileJson } = await response.json();
-    fileJson
-      .flatMap((item: { files?: { path: string }[]; path: string }) => {
-        return item.files ? item.files.map((f: any) => f.path) : item.path;
-      })
-      .filter((path: string) => path.endsWith('.d.ts'))
-      .map(async (path: string) => {
-        const res = await fetch(
-          `https://unpkg.com/@types/node@${downloadVersion}${path}`,
-        );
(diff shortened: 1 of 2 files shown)


BlackHole1 (approved):

georgexu99 (approved):
adding extra lib -> 52 times seems a bit weird to me but if that can be explained then its a gogogo from me

dsanders11:
> adding extra lib -> 52 times seems a bit weird to me but if that can be explained then its a gogogo from me

@georgexu99, there are 51 individual `*.d.ts` files are in the Node typings ([see the test fixture](https://github.com/electron/fiddle/blob/37f259a59f26410b1e2e4ca1b9b67386cf577b4a/tests/fixtures/node-types.json)). Then there's one `electron.d.ts` file, for 52 total.

malept:
> > adding extra lib -> 52 times seems a bit weird to me but if that can be explained then its a gogogo from me
> 
> @georgexu99, there are 51 individual `*.d.ts` in the Node typings ([see the test fixture](https://github.com/electron/fiddle/blob/37f259a59f26410b1e2e4ca1b9b67386cf577b4a/tests/fixtures/node-types.json)). Then there's one `electron.d.ts` file, for 52 total.

IMO that should be a comment right next to that 52 magic number in the test. Having a future developer dig through PRs to find an explanation is suboptimal.

dsanders11:
Good point @malept. I'll make a note to do a follow-up PR that adds that context and some more context to some of the other tests in the same area as well.

PR #1224, opened 2022-12-08
fix: don't clear editor is edited state if Gist publish fails

### LINKED ISSUE #1142: Ask to confirm before closing app and loosing data
I was spending quite some time on writing a fiddle and then went over to publish as Gist. Here I was greeted with an error, so I opened settings and updated by GitHub access token.

Then I was thinking that the X button top keft closes settings (assuming this was another window on top of the fiddle window), but in fact it was closing the entire app and all my changes were lost.

image

**Ideally:**
* set the window to document edited (macOS) when the user makes any input
* ask to confirm before closing the window or app (all OS)
* best: restore the contents after restart?

dsanders11:
Closes #1142.

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -115,7 +115,7 @@ export const GistActionButton = observer(
       });
     }
 
-    private async publishGist(description: string) {
+    private async publishGist(description: string): Promise<boolean> {
       const { appState } = this.props;
 
       const octo = await getOctokit(appState);
@@ -159,6 +159,8 @@ export const GistActionButton = observer(
 
         // Only set action type to update if publish completed successfully.
         this.setActionType(GistActionType.update);
+
+        return true;
       } catch (error) {
         console.warn(`Could not publish gist`, { error });
 
@@ -172,6 +174,8 @@ export const GistActionButton = observer(
           IpcEvents.SHOW_WARNING_DIALOG,
           messageBoxOptions,
         );
+
+        return false;
       }
     }
 
@@ -182,14 +186,17 @@ export const GistActionButton = observer(
       const { appState } = this.props;
       appState.activeGistAction = GistActionState.publishing;
 
-      const description = await this.getFiddleDescriptionFromUser();
+      try {
+        const description = await this.getFiddleDescriptionFromUser();
 
-      if (description) {
-        await this.publishGist(description);
-        appState.editorMosaic.isEdited = false;
+        if (description) {
+          if (await this.publishGist(description)) {
+            appState.editorMosaic.isEdited = false;
+          }
+        }
+      } finally {
+        appState.activeGistAction = GistActionState.none;
       }
-
-      appState.activeGistAction = GistActionState.none;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/renderer/components/commands-action-button.tsx
```diff
@@ -159,6 +159,8 @@ export const GistActionButton = observer(
 
         // Only set action type to update if publish completed successfully.
         this.setActionType(GistActionType.update);
+
+        return true;
```
BlackHole1, line 163:
```suggestion
```

src/renderer/components/commands-action-button.tsx
```diff
@@ -172,6 +174,8 @@ export const GistActionButton = observer(
           IpcEvents.SHOW_WARNING_DIALOG,
           messageBoxOptions,
         );
+
+        return false;
```
BlackHole1, line 178:
```suggestion

        throw error;
```

src/renderer/components/commands-action-button.tsx
```diff
@@ -182,14 +186,17 @@ export const GistActionButton = observer(
...
-      if (description) {
-        await this.publishGist(description);
-        appState.editorMosaic.isEdited = false;
+        if (description) {
+          if (await this.publishGist(description)) {
+            appState.editorMosaic.isEdited = false;
+          }
+        }
+      } finally {
+        appState.activeGistAction = GistActionState.none;
```
BlackHole1, line 198: Here we need to consider the case where `getFiddleDescriptionFromUser` fails

```suggestion
        if (!description) {
          return;
        }
        
        await this.publishGist(description)
        appState.editorMosaic.isEdited = false;
      } catch (_error) {
        appState.editorMosaic.isEdited = true;
      } finally {
        appState.activeGistAction = GistActionState.none;
```

BlackHole1 (changes requested):
LGTM, but there are some nitpicks.

src/renderer/components/commands-action-button.tsx
dsanders11: Not sure I understand the benefit of this change. In this PR `getFiddleDescriptionFromUser()` is already wrapped in the try/catch block, and the return value is already checked. I don't see which case is not covered.

I also don't think we should do `appState.editorMosaic.isEdited = true;`, since the files might not be edited, so toggling that to true is incorrect state. For example, saving a fiddle locally would mean `isEdited = false`, but with this suggested code change, if you try to publish after saving it, and the publish fails, then the editor is marked as `isEdited = true`, which isn't correct. The publish step can clear the `isEdited` state, it shouldn't ever toggle it to true.

BlackHole1: > I also don't think we should do appState.editorMosaic.isEdited = true;, since the files might not be edited, so toggling that to true is incorrect state. For example, saving a fiddle locally would mean isEdited = false, but with this suggested code change, if you try to publish after saving it, and the publish fails, then the editor is marked as isEdited = true, which isn't correct. The publish step can clear the isEdited state, it shouldn't ever toggle it to true.

Thank you for adding the context

> Not sure I understand the benefit of this change. In this PR getFiddleDescriptionFromUser() is already wrapped in the try/catch block, and the return value is already checked. I don't see which case is not covered.

my initial thought was that when `getFiddleDescriptionFromUser` fails, it also requires `appState.editorMosaic.isEdited = false;`

BlackHole1 (approved):

ckerr (approved):
LGTM, I like this idea.

Issue #1238, opened 2023-01-17
Action Button Stuck on "Checking status" After Downloading Version


dsanders11:
**Steps to Reproduce:**

1. Go to Settings -> Electron
2. Click "Download" next to an Electron version and wait for it to finish downloading
3. Close Settings
4. Observe that the action button is now stuck in "Checking status" state
5. Changing the version on the selector dropdown gets the button back to a normal state

ckerr:
From settings, we should be calling fiddle-core's [`installer.ensureDownloaded()`](https://github.com/electron/fiddle-core/blob/main/src/installer.ts#L302) instead of `installer.install()`

PR #1242, opened 2023-01-19
fix: don't activate when downloading versions from settings

codebytere:
Closes https://github.com/electron/fiddle/issues/1238.

`fiddle-core` was installing versions and activating them regardless of origin - when a version is downloaded from settings it should not be activated.

### DIFF
--- a/src/renderer/components/settings-electron.tsx
+++ b/src/renderer/components/settings-electron.tsx
@@ -383,7 +383,7 @@ export const ElectronSettings = observer(
           buttonProps.onClick = () => {
             isLocal
               ? appState.removeVersion(ver)
-              : appState.downloadVersion(ver);
+              : appState.downloadVersion(ver, { noActivate: true });
           };
           break;
       }
--- a/src/renderer/components/settings-electron.tsx
+++ b/src/renderer/components/settings-electron.tsx
@@ -383,7 +383,7 @@ export const ElectronSettings = observer(
           buttonProps.onClick = () => {
             isLocal
               ? appState.removeVersion(ver)
-              : appState.downloadVersion(ver, { noActivate: true });
+              : appState.downloadVersion(ver, { activate: true });
           };
           break;
       }
--- a/src/renderer/components/settings-electron.tsx
+++ b/src/renderer/components/settings-electron.tsx
@@ -383,7 +383,7 @@ export const ElectronSettings = observer(
           buttonProps.onClick = () => {
             isLocal
               ? appState.removeVersion(ver)
-              : appState.downloadVersion(ver, { activate: true });
+              : appState.downloadVersion(ver, { activate: false });
           };
           break;
       }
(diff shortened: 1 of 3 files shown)


src/renderer/state.ts
```diff
@@ -610,6 +613,12 @@ export class AppState {
       return;
     }
 
+    // Download the version without setting it as the current version.
+    if (opts?.noActivate) {
+      await this.installer.ensureDownloaded(version);
```
dsanders11, line 618: This skips the options passed to `installer.install`, as well as the log statement on line 613/622. Suggestion: pull out the options into `const options = {...};` and then do `this.installer.ensureDownloaded(version, options);`.

Unrelated to this change, but it looks like there's another bug around showing progress on the spinner when downloading under Settings, it works properly on the version select. 🤔

src/renderer/components/settings-electron.tsx
```diff
@@ -383,7 +383,7 @@ export const ElectronSettings = observer(
           buttonProps.onClick = () => {
             isLocal
               ? appState.removeVersion(ver)
-              : appState.downloadVersion(ver);
+              : appState.downloadVersion(ver, { noActivate: true });
```
ckerr, line 386: _nit_

Negated logic like this is often harder to follow / easier to accidentally get wrong later on.

Would it be possible to make the opts `activate` instead, defaulting to true?

ckerr (approved):
I do have a nit to pick but the main idea in this PR LGTM

src/renderer/state.ts
codebytere: What do you mean here precisely? what options would apply to `ensureDownloaded` in this context?

dsanders11: [These options](https://github.com/electron/fiddle/blob/4beb8aba7deb2dc4f60fe874912fe3db45251765/src/renderer/state.ts#L615-L627). Both code paths should probably be consistent in which mirrors they use, and the `progressCallback` option is definitely needed of #1243 will no longer work.

codebytere: @dsanders11 done!

src/renderer/components/settings-electron.tsx
```diff
@@ -383,7 +383,7 @@ export const ElectronSettings = observer(
           buttonProps.onClick = () => {
             isLocal
               ? appState.removeVersion(ver)
-              : appState.downloadVersion(ver);
+              : appState.downloadVersion(ver, { activate: true });
```
dsanders11, line 386: I think this should be flipped to `false`?

src/renderer/state.ts
```diff
@@ -577,12 +577,18 @@ export class AppState {
   }
 
   /**
-   * Download a version of Electron.
+   * Download a version of Electron and set it as the current
+   * version in use unless otherwise specified.
    *
    * @param {RunnableVersion} ver
+   * @param {Object} opts
+   * @param {Boolean} [options.activate=100] - Whether to set ver as current
```
dsanders11, line 585: `options.activate=100` => `opts.activate=true`?

dsanders11 (approved):
:shipit:

Issue #1246, opened 2023-01-25
BUG: Doubleclicking the buttons in the title bar maximizes or minimizes the window.


GGIEnrike:
This really frustrates me. When I quickly click the start/stop button twice to restart my fiddle, the window suddenly minimizes. Maximizing/minimizing of the window should only be triggered by clicking anything BUT the buttons on the title bar. 

The way I fixed it in my app was by keeping a global variable in the renedered of a state of whether the mouse is inside the button.
Something like this (I am not familiar with fiddle's source):

```JavaScript
const titleBar = document.querySelector('#titleBar')
const buttons = document.querySelectorAll('.barButton')

let canDoubleClick = true

buttons.forEach((but) => {
    but.addEventListener('mouseenter', () => {
        canDoubleClick = false
    })

    but.addEventListener('mouseleave', () => {
        canDoubleClick = true
    })
})

titleBar.addEventListener('doubleClick', () => {
    if (canDoubleClick) {
        window.electronAPI.doubleClicked()
    }
})
```

PR #1248, opened 2023-01-26
fix: overzealous window maximize on double-click

codebytere:
Closes https://github.com/electron/fiddle/issues/1246.


ckerr (commented):
Introduced in https://github.com/electron/fiddle/pull/1059

Looks like the tests there need to be updated for this PR?

ckerr (approved):

PR #1258, opened 2023-02-05
fix: GitHub tokens can have uppercase and underscores

dsanders11:
This functionality has probably been broken for a bit. Personal Access Tokens (PATs) from GitHub can include underscores and uppercase letters (may not have been true when that code was written), so update the regex.


src/renderer/components/dialog-token.tsx
```diff
@@ -120,7 +120,7 @@ export const TokenDialog = observer(
       const text = (clipboard.readText() || '').trim();
 
       if (text.length !== 40) return;
-      if (!/^[a-z0-9]+$/.test(text)) return;
+      if (!/^[a-zA-Z0-9_]+$/.test(text)) return;
```
MarshallOfSound, line 123: Should we be more specific, modern tokens are `ghp_[a-zA-Z0-9]{n}` (I can't remember what N is)

dsanders11: `n` must be 36, because it still is 40 characters long.

If we update it to be more specific, would probably want to also cover the new fine-grained tokens, which look to be in the format `github_pat_[a-zA-Z0-9_]{n}`, with N being 82, at least in a quick spot check, not sure if it changes since they're more complex.

MarshallOfSound: The patterns are fixed because they're used for secret scanning

BlackHole1: Github PAT Regex: https://gist.github.com/magnetikonline/073afe7909ffdd6f10ef06a00bc3bc88

erickzhao (commented):
Ah yeah, this code was written for Classic PATs. We should probably only support the fine-grained access ones going forward but keep existing classic PATs working to avoid breaking users.

dsanders11:
Updated with the regex that @BlackHole1 provided, merged in `main` and simplified the check.

BlackHole1 (approved):

PR #1263, opened 2023-02-09
chore: cleanup typings in tests

dsanders11:
[The [Type] Safety Dance](https://www.youtube.com/watch?v=0QDKLglEP5Y)

Cleans up a bunch of usage of `as any` typings in the tests, drops three unused files, tightens up some other typings.

Caught two bugs in the tests, so I'm going to claim it was totally worth the rabbit hole I went down. :eyes:
* https://github.com/electron/fiddle/commit/c042d21f5a26f2520278cbafec9a3683354c1267#r99937481
* https://github.com/electron/fiddle/commit/c042d21f5a26f2520278cbafec9a3683354c1267#r99937524


erickzhao (commented):
Did everything just work when you removed our mocks for the actual state objects? Magical.

dsanders11:
> Did everything just work when you removed our mocks for the actual state objects? Magical.

They're still using the mocks, just the typing is changed. I opted to change types so that there are the least amount of type assertions, so test code that doesn't need to use any of the mock state checking except for say one spot in the file was changed to use the actual object type, and a type assertion in the one spot it needs to access the mock. There are a couple of files where the inverse is true, and most of the test code needs to interact with it as a mock, so use a type assertion for the actual type there when passing to functions that expect the real type, instead of the existing `as any` usages.

codebytere (approved):

erickzhao (approved):

Issue #1203, opened 2022-11-01
Migrate `fiddle-core` to main process


aryanshridhar:
Covers the major part of code migration of Node API to the renderer process.

`IPC.INITIALIZE_FIDDLE`- Initialize the fiddle core in main process.
`IPC.RUN_FIDDLE`- IPC event to run fiddle.
`IPC.STOP_FIDDLE`- IPC event to stop the running fiddle.

aryanshridhar:
(Will create separate pr for better coder review/git history)

dsanders11:
Closed by #1452.

PR #1265, opened 2023-02-11
fix: keep versions in-sync with fiddle-core

### LINKED ISSUE #1241: Can't Run Fiddle on Newly Released Electron Versions Without Restart
If a new Electron version has been released since you last launched Fiddle, you won't be able to run with that version despite it being available in the version selector, due to "unrecognized version". Closing the app and opening it again fixes that.

This is due to how it is integrated with `@electron/fiddle-core` and `BaseVersions` - the list of versions is handed to `fiddle-core` on start, and then the rest of the app fetches the latest versions, and updates the UI. So `fiddle-core` has an outdated list of versions, and won't pick up the new ones until it is restarted.

Somewhat related to #1222.

dsanders11:
Update versions known to `fiddle-core` when updating versions in the UI.

This is a bit of a hack until #1203 is complete.

Fixes #1241.

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -52,6 +52,12 @@ import {
   saveLocalVersions,
 } from './versions';
 
+class UpdateableBaseVersions extends BaseVersions {
+  public updateVersions(val: unknown): void {
+    this.setVersions(val);
+  }
+}
+
 /**
  * The application's state. Exported as a singleton below.
  *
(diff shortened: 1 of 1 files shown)


dsanders11:
This can be manually confirmed to work by deleting the `known-electron-versions` key in `localStorage` and restarting Fiddle, then trying to run v23.0.0. On `main` that'll fail, with this PR it'll work.

aryanshridhar (approved):
Tested it locally and works great.
LGTM 👍

erickzhao (approved):

Issue #1244, opened 2023-01-20
BUG: New debs of Electron Fiddle are made with tar.zst (need to revert to ubuntu-20.04 or ubuntu-18.04)


theofficialgman:
ubuntu-latest is now ubuntu 22.04

ubuntu has an extension to apt that uses tar.zst to compress file in debs. this is used by default by dpkg when packaging debs in 22.04

this prevents all debs from installing on debian/raspbian

you need to switch to building with ubuntu-20.04 or ubuntu-18.04 which do not use this functionality when packaging debs

https://github.com/Botspot/pi-apps/actions/runs/3971114611/jobs/6807621730#step:4:939

theofficialgman:
alternativly you can specify the compression method as gzip through a dpkg command like option, debhelper override, or whatever method you use to build these debs

mitesh-raval:
Workaround :

Follow the steps in the following link for repackaging as tar.xz (replacing the other .deb name for electron-fiddle~.deb) : 

https://gist.github.com/goproslowyo/fe95c8e07c73593e2099687eeda78bc5

I just did this on my chromebook linux (Debian 11) and I was able to install electron-fiddle successfully.

PR #1276, opened 2023-02-28
build: migrate from GH Actions to CircleCI

VerteDinde:
This PR migrates Fiddle from GitHub Actions to CircleCI
- [x] Tested Mac, Windows and Linux binaries
- [x] Tested releasing a new version to GitHub (based on tags)

Follow up: Re-adding Coveralls as a CI check. We need access to an admin token to do so - let's get a release out first.

Notes: none


.circleci/config.yml
```diff
@@ -0,0 +1,123 @@
...
+          path: out
+      - persist_to_workspace:
+          root: '*'
+          paths:
+            - out
+  maybe-publish-to-github:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - run: sudo apt-get update && sudo apt install rpm
```
dsanders11, line 93: I don't think we need to install the `rpm` package here since they should be built already.

I think we need to add `attach_workspace` here to access what was persisted in the previous jobs.

.circleci/config.yml
```diff
@@ -0,0 +1,123 @@
...
+              arch: [ x64, ia32 ]
+      - linux-build:
+          matrix:
+            parameters:
+              arch: [ x64, arm64, armv7l ]
+      - maybe-publish-to-github:
+          requires:
+            - mac-build
+            - win-build
+            - linux-build
```
dsanders11, line 117: Let's also require `lint` for publish.

I think these have to be the full name with the arch, e.g. `linux-build-x64`, but I'm not positive. If the base name works, that's great, and would simplify other configs, but on other repos we require the full expanded matrix.

.gitignore
```diff
@@ -36,7 +36,7 @@ report.json
 /npm/path.txt
 
 # Electron Fiddle-Specific
-/static/contributors.json
+# /static/contributors.json
```
dsanders11, line 39: Let's revert to just doing say `echo "[]" > static/contributors.json` in the jobs before running tests - should work for all platforms. Otherwise we'll end up in the annoying situation of this file accidentally getting committed in PRs, etc.

dsanders11 (commented):
Meta comment: I think we want to do like the existing GHA workflow and have separate test jobs (which will be real simple), and then have the build jobs filtered to only run on tags, and require the test jobs (and lint), and then `maybe-publish-to-github` will depend on the build jobs. Probably don't even have to filter `maybe-publish-to-github` then, since it would only run if the build jobs which are filtered do, but might as well also filter it.

dsanders11:
Another comment, as I'm reminded by #127. The current state of this PR also drops Coveralls - do we want to use https://circleci.com/developer/orbs/orb/coveralls/coveralls to maintain that? I know nobody loves Coveralls, but I don't think we necessarily want to hard drop it if we can avoid it.

.gitignore
VerteDinde: Yeah, totally agreed, this was a dirty debugging tactic, should not be left this way 👍

erickzhao:
wanted to say we should be aware of existing build system issues like https://github.com/electron/fiddle/issues/1244 to see we can close it out

VerteDinde:
Addressing two comments above:
1. It looks like we need to get the Coveralls token out of our Actions and into CircleCI's context, working on getting that now.
2. Circle's Linux docker images are using Ubuntu 20.04 here, so we should avoid the issue that Erick mentioned

.circleci/config.yml
```diff
@@ -0,0 +1,212 @@
...
+
+jobs:
+  lint:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - install
+      - load-gh-token
+      - run: yarn run lint
+      - run: yarn run test:report
```
dsanders11: Do we want this here? We weren't using it in the past, and looks to me like it runs the test suite and dumps to a file.

dsanders11 (approved):

Issue #1281, opened 2023-03-13
Remote loader tries to download versions incompatible with OS


codebytere:
As in title. If a user tries to load a gist with a package.json using an Electron version too old for or incompatible with the current OS, it is loaded anyway. 

Example: https://github.com/electron/electron/issues/37491 uses v9.0.0, which should error on M1 machines.

PR #1282, opened 2023-03-13
fix: remote loader loading incompatible versions

codebytere:
Closes https://github.com/electron/fiddle/issues/1281.

Fixes an issue where if a user tries to load a gist with a package.json using an Electron version too old for or incompatible with the current OS, it is loaded anyway.

After:
<details>
Screenshot 2023-03-13 at 1 38 42 PM
</details>


BlackHole1 (approved):
LGTM

PR #1289, opened 2023-03-28
refactor: add type awareness to getItemIcon()

ckerr:
Nothing major; this is a side-refactor I wrote while looking at something else in Fiddle.

- Add type awareness (IconName) to `getItemIcon()`
- Simplify the logic of `getItemLabel()` and `getItemIcon()` by making a const Record of the InstallState enum types.

### DIFF
--- a/src/renderer/components/version-select.tsx
+++ b/src/renderer/components/version-select.tsx
@@ -86,7 +86,7 @@ export function getItemLabel({ source, state, name }: RunnableVersion): string {
  * version.
  *
  * @param {RunnableVersion} { state }
- * @returns
+ * @returns {IconName}
  */
 export function getItemIcon({ state }: RunnableVersion): IconName {
   const installStateIcons: Record<InstallState, IconName> = {
--- a/src/renderer/components/version-select.tsx
+++ b/src/renderer/components/version-select.tsx
@@ -4,6 +4,7 @@ import {
   Button,
   ButtonGroupProps,
   ContextMenu,
+  IconName,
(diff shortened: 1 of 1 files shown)


src/renderer/components/version-select.tsx
```diff
@@ -93,18 +88,16 @@ export function getItemLabel({ source, state, name }: RunnableVersion): string {
...
+export function getItemIcon({ state }: RunnableVersion): IconName {
+  const installStateIcons: Record<InstallState, IconName> = {
+    missing: 'cloud',
+    downloading: 'cloud-download',
+    downloaded: 'compressed',
+    installing: 'compressed',
+    installed: 'saved',
+  } as const;
+
+  return installStateIcons[state] || '';
```
dsanders11, line 100: Note for posterity (and my own future reference). I was concerned this change might remove the exhaustiveness checking you get in TypeScript with a switch statement with no `default`, but it looks like TypeScript also does exhaustiveness checking with a `Record` when the key is a finite type. :+1:

src/renderer/components/version-select.tsx
```diff
@@ -93,18 +88,16 @@ export function getItemLabel({ source, state, name }: RunnableVersion): string {
  * @param {RunnableVersion} { state }
  * @returns
```
dsanders11, line 89:
```suggestion
 * @returns {IconName}
```

dsanders11 (approved):

codebytere (approved):

PR #1291, opened 2023-03-30
fix: focus Fiddle app for unsaved fiddle dialog (#1036)

### LINKED ISSUE #1036: Quitting Unsaved Fiddle Should Focus Window
Same kind of issue as #1007, it's showing a "Are you sure?" modal but the window isn't focused so the behavior on macOS is trying to quit from the dock seems to have no effect since it's waiting for user input but the user can't see.

Fix should be simple, see #1018 for how #1007 was fixed.

ryan-halbrook:
Fixes #1036.

Focuses the Fiddle app before presenting unsaved fiddle dialogs. Previously, quitting the app from the Mac OS dock would appear to do nothing if the app is covered or hidden.

Changes summary:

- Use the `App` api focus method with steal option to take focus from the foreground app before showing the dialogs.
- Call this through a new IPC event type, `FOCUS_APP`. I added it because the dialog is conditionally presented from the renderer process. As far as I know, this code doesn't have access to BrowserWindow or App object.

### DIFF
--- a/src/ipc-events.ts
+++ b/src/ipc-events.ts
@@ -39,7 +39,7 @@ export enum IpcEvents {
   TASK_DONE = 'TASK_DONE',
   OUTPUT_ENTRY = 'OUTPUT_ENTRY',
   RELOAD_WINDOW = 'RELOAD_WINDOW',
-  FOCUS_APP = 'FOCUS_APP',
+  SHOW_WINDOW = 'SHOW_WINDOW',
 }
 
 export const ipcMainEvents = [
@@ -56,7 +56,7 @@ export const ipcMainEvents = [
   IpcEvents.OUTPUT_ENTRY,
   IpcEvents.TASK_DONE,
   IpcEvents.RELOAD_WINDOW,
-  IpcEvents.FOCUS_APP,
+  IpcEvents.SHOW_WINDOW,
 ];
 
 export const ipcRendererEvents = [
--- a/src/ipc-events.ts
+++ b/src/ipc-events.ts
@@ -39,6 +39,7 @@ export enum IpcEvents {
   TASK_DONE = 'TASK_DONE',
   OUTPUT_ENTRY = 'OUTPUT_ENTRY',
   RELOAD_WINDOW = 'RELOAD_WINDOW',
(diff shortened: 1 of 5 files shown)


dsanders11:
Hi @ryan-halbrook! Thanks for the PR tackling this.

Looks like there's a couple of linting errors which need to be cleaned up. Also, I see you're making the PR from the `main` branch on your fork - I'd recommend using a branch on your fork for PRs, it'll make it easier to rebase/merge changes from `main` as needed.

What do you think about refactoring your implementation to use a new generic IPC event called `SHOW_WINDOW`, which would use [`BrowserWindow.show()`](https://www.electronjs.org/docs/latest/api/browser-window#winshow) to show the window which send the IPC event, which could be found by using the `sender` property on the [`IpcMainEvent`](https://www.electronjs.org/docs/latest/api/structures/ipc-main-event) and [`BrowserWindow.fromWebContents`](https://www.electronjs.org/docs/latest/api/browser-window#browserwindowfromwebcontentswebcontents)? There may be other use cases where the renderer needs to show its window.

The current implementation in this PR doesn't cover the case where the app has focus, but the window isn't visible. This can be tested by minimizing the window and then trying to quit the app. Ideally a fix to #1036 would also cover that case.

dsanders11:
My suggested change is similar to what #348 did (this issue has regressed, unfortunately), but tweaked to show the specific window requesting to be shown, rather than the main window.

ryan-halbrook:
Hi @dsanders11! Thanks for your feedback. 

Ok, SHOW_WINDOW seems like a better addition just wondering if it'll preserve the window ordering and if we care about that for UX.

dsanders11:
> Ok, SHOW_WINDOW seems like a better addition just wondering if it'll preserve the window ordering and if we care about that for UX.

I'm not sure that it will preserve the window ordering, but I don't think we care about that in this case - I don't think any attention has been paid to that in Fiddle. There's also `BrowserWindow.showInactive` which doesn't focus the window (which #348 used), which may act differently in regard to window ordering.

ryan-halbrook:
Switched to a new PR to use a different branch. See #1294

PR #1292, opened 2023-03-30
fix: use nativeTheme module

dsanders11:
Currently the main process never sets `nativeTheme.themeSource`, so setting Fiddle to use a dark or light theme, rather than syncing with the system, ends up being out of sync with itself. If your system is set to dark mode, and you set light mode in Fiddle, on Windows 10 the titlebar will stay dark, for example. On macOS things like the background color of context menus will follow the system theme rather than the set theme in Fiddle.

Pretty straight forward fix. Tested on macOS and Windows 10.

### DIFF
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -1,3 +1,7 @@
+:root {
+  color-scheme: light dark;
+}
+
 body {
   padding: 0;
   margin: 0;
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -1,7 +1,7 @@
 // eslint-disable-next-line import/order
 import { initSentry } from '../sentry';
 initSentry();
-import { BrowserWindow, app, systemPreferences } from 'electron';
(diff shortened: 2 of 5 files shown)


src/less/main.less
```diff
@@ -1,3 +1,7 @@
+:root {
+  color-scheme: light dark;
```
dsanders11, line 2: This ensures that UI elements like scrollbars on Windows use the right theme.

src/main/main.ts
```diff
@@ -106,6 +107,23 @@ export function setupTitleBarClickMac() {
   });
 }
 
+function isNativeThemeSource(
+  val: unknown,
+): val is typeof nativeTheme.themeSource {
```
codebytere, line 112: Under what conditions would this return `false`? I could _maybe_ see the case for checking that it's 'dark', 'light', or 'system' but i don't see how it wouldn't be a string 🤔

VerteDinde (approved):

PR #1294, opened 2023-03-31
fix: show browser window before unsaved dialog

### LINKED ISSUE #1036: Quitting Unsaved Fiddle Should Focus Window

ryan-halbrook:
Fixes #1036 

Focuses the `BrowserWindow` before presenting the unsaved fiddle dialog. Previously, quitting the app from the Mac OS dock would appear to do nothing if the app is covered or hidden.

Based on comments from #1291, creating a new PR to use a separate branch as recommended by @dsanders11.


codebytere (approved):

dsanders11 (changes requested):
In my testing, the current state of the PR seems to have a race condition when quitting with a minimized window on macOS, which leads to the window sometimes closing immediately instead of blocking asking for user confirmation.

In my testing the following tweaked implementation doesn't seem to suffer from the problem. I'm not positive which change removes the race condition, but the `setTimeout` is used to push the logic to the next tick in an effort to ensure the `onbeforeunload` event is canceled.

```typescript
window.onbeforeunload = (e: BeforeUnloadEvent) => {
  // On Mac OS, quitting can be triggered from the dock,
  // show the window before presenting the dialog
  setTimeout(() => {
    this.confirmExitUnsaved().then((quit) => {
      if (quit) {
        // isQuitting checks if we're trying to quit the app
        // or just close the window
        if (state.isQuitting) {
          ipcRendererManager.send(IpcEvents.CONFIRM_QUIT);
        }
        window.onbeforeunload = null;
        window.close();
      }
    });
    ipcRendererManager.send(IpcEvents.SHOW_WINDOW);
  });

  // return value doesn't matter, we just want to cancel the event
  e.returnValue = false;
};
```

dsanders11:
@ryan-halbrook, apologies, there are a few conflicts on this PR now due to other refactors landing while this was in-flight.

ryan-halbrook:
@dsanders11 I'm not seeing the minimize problem. I'm starting Fiddle, making a change, minimizing, then quitting.

From the code, Fiddle shouldn't close the window until it executes the lines:
```
window.onbeforeunload = null;
window.close();
```
which only execute after waiting for the result from the dialog. I'm very new to this and probably missing something 😅.

dsanders11:
@ryan-halbrook, I'll take a look again. In the mean time, could you merge `main` again? 😄 We've switched CI over to CircleCI now and put out a new release, so this PR needs to pick it up so the proper CI runs.

> From the code, Fiddle shouldn't close the window until it executes the lines:

In theory, yes. In practice, there have been reports in the past about `onbeforeunload` being buggy with Electron, so I might be hitting something with that. In my testing it felt like there was a race condition when showing the window from minimized, as if the `onbeforeunload` event hadn't been canceled yet and by showing the minimized window it caused it to quit before the handler finished executing (due to the `await`). That's my working theory at least, and why I moved the relevant code into a `setTimeout` so that the `onbeforeunload` handler would immediately cancel the event.

dsanders11:
@ryan-halbrook, here's what I see on my machine (Intel MBP, macOS 13.1) with the current state of the PR. In this case the first quit didn't lose the race condition, but subsequent ones did. In my tests (probably varies by machine) the window closes immediately 90% of the time. With the proposed changes it doesn't have the problem.

Recording 2023-04-05 at 17 44 57

ryan-halbrook:
@dsanders11 I see, I'm running it on an arm-based MacBook, macOS 13.3.

dsanders11:
@ryan-halbrook, I'll test it on my M1 MBP later out of curiosity, but yea, if it's a race condition it's possible it won't replicate on all machines. My proposed code change seems robust in my testing, though.

ryan-halbrook:
Working with your proposed changes, I changed the tests to pass an event to onbeforeunload:
```
const e = {                       
  returnValue: Boolean,           
}                                 
window.onbeforeunload!(e as any); 
expect(e.returnValue).toBe(false);
```
but the expect fails now:

```
expect(jest.fn()).toHaveBeenCalledWith(...expected)                                
                                                                                   
Expected: "SHOW_WINDOW"                                                            
                                                                                   
Number of calls: 0                                                                 
                                                                                   
  440 |       window.onbeforeunload!(e as any);                                    
  441 |       expect(e.returnValue).toBe(false);                                   
> 442 |       expect(ipcRenderer.send).toHaveBeenCalledWith(IpcEvents.SHOW_WINDOW);
      |                                ^                                           
  443 |       expect(window.close).toHaveBeenCalled();                             
  444 |     });                                                                    
  445 |                                   
```              

How can I have Jest wait for the setTimeout function to complete?

dsanders11:
@ryan-halbrook, I'm not sure the new test, "sends SHOW_WINDOW event if there are unsaved changes" is necessary, since the other tests cover it. As for the problem you're seeing, maybe you could use the `waitFor` util to wait for the IPC event, similar to how other tests like this one use it:

https://github.com/electron/fiddle/blob/4f0fe847b7266d2869ba2a9f6581a788889a69d7/tests/renderer/editor-mosaic-spec.ts#L459

Perhaps (untested):

```js
await waitFor(() => ipcRenderer.send.mock.calls.find(([channelName]) => channelName === IpcEvents.SHOW_WINDOW));
```

ryan-halbrook:
@dsanders11 That works, thanks 😄 I committed the onbeforeunload change.

dsanders11 (approved):

dsanders11:
Thanks @ryan-halbrook!

ryan-halbrook:
My first PR 😁 Thanks @dsanders11!

dsanders11:
@ryan-halbrook, first of many, I hope. 🙂 🎉

Issue #1287, opened 2023-03-21
Closing the last window quits the app after a quit is cancelled


ryan-halbrook:
When closing the last window, the behavior on Mac OS is to stay open, but this doesn't work if a previous quit was canceled due to an unsaved Fiddle.

Steps to reproduce:
1. Start Fiddle app.
2. Edit a file.
3. Quit Fiddle.
4. Click cancel when prompted to exit an unsaved Fiddle.
5. Close the window. Click exit when prompted about the Fiddle being unsaved. The app quits instead of remaining open.

dsanders11:
Thanks for the clear repro steps!

I've confirmed the behavior on 0.32.1 - if anyone else tries to repro, note that the distinction between quitting and closing the window is important. It also does not appear to repro when you have multiple unsaved windows, until you exit the last window.

sekarah:
Hi! I believe I just fixed this bug. I want to submit a pull request but I don't believe I have contributor permissions to make my own branch. Should I make a fork to compare and then submit my pull request? Sorry if this question is documented anywhere -- I read the contributing guide and other github documentation but couldn't find anything. Also a beginner to open source contribution in general. 

Thanks

ryan-halbrook:
Hi @sekarah. I just learned about this. Create your own fork of electron/fiddle. In that fork, create a branch and make changes in the branch. Then submit a pull request from that branch. The Github documentation has specifics about forks and pull requests.

sekarah:
Hi again, just want to make sure my pull request conventions are right. My pull request has two commits with the following titles: 

fix: resets isQuitting flag to false when user cancels dialog
test: checks that isQuitting variable is reset after user cancels dialog 

I know that the contribution guidelines ask to follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) guide, but I'm unsure if my title should account for both commits. Any advice on what my title should be would be appreciated.

Below is the branch comparison:
https://github.com/electron/fiddle/compare/main...sekarah:fiddle:check-is-quitting?expand=1

Thanks

dsanders11:
[...]

PR #1308, opened 2023-04-13
fix: resets isQuitting flag to false when user cancels dialog

sekarah:
Fixes [#1287](https://github.com/electron/fiddle/issues/1287)

This addresses the issue raised [here](https://github.com/electron/fiddle/issues/1287), where closing the last window quits the app after a quit is cancelled. I just added logic that resets the state's isQuitting flag to false when confirmExitUnsaved returns false. Before this, the isQuitting flag remained true whenever a quit with unsaved data was prompted, regardless of whether it was cancelled or not -- this meant that closing the window afterwards would quit the app. 

I added a line to the end of the 'does nothing if user cancels the dialog' test that checks if the flag is reset.

### DIFF
--- a/src/renderer/app.tsx
+++ b/src/renderer/app.tsx
@@ -297,6 +297,8 @@ export class App {
               }
               window.onbeforeunload = null;
               window.close();
+            } else {
+              state.isQuitting = false;
             }
           });
           ipcRenderer.send(IpcEvents.SHOW_WINDOW);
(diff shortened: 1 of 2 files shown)


dsanders11:
Thanks @sekarah! If you update the description to start with "Fixes #1287", this PR will be linked to that issue, and the issue will be automatically closed as fixed when this PR is merged. 🙂

dsanders11 (approved):

codebytere (approved):

Issue #1222, opened 2022-12-08
Node.js Intellisense Not Available For New Electron Versions


dsanders11:
When adding Node.js typings to the editors the code currently uses the static `releases.json` file shipped with Fiddle, so typings for versions released after that version of Fiddle are not available until a new Fiddle release happens, which is no good.

💡 **Proposal**

Let's update `ElectronVersions` in `fiddle-core` to take an `initialVersions` in [the `create` static method](https://github.com/electron/fiddle-core/blob/8bbdce36b0f5e06529e0f01b4f5107b4880a96b8/src/versions.ts#L218-L243), which it will use to populate the versions cache iff it is not found on disk. That way the cache is always pre-populated in the offline case, and can immediately fetch newer versions in the online case.

PR #1316, opened 2023-04-15
build: move Electron releases fetch to Forge generateAssets hook

dsanders11:
Pull `releases.json` in `generateAssets` so that it is always the latest and there are no more generated PRs.

Minor drawback is that builds will no longer be fully reproducible since `releases.json` will be pulled at the time you do the build and may be different than your last build.


andersk:
> Minor drawback is that builds will no longer be fully reproducible since `releases.json` will be pulled at the time you do the build and may be different than your last build.

This has also broken the Electron Fiddle package that I maintain for NixOS. All NixOS packages are built in a sandbox with no external network access (except fixed-output derivations that check a hash of their output, for downloading sources).

dsanders11:
Sorry about that @andersk. Would gracefully failing if the fetch fails and allowing the build to continue work for you? That's what happens with `maybeFetchContributors`, which has been a build-time step for a while now.

andersk:
Yeah that sounds good; the release data will be re-fetched at runtime right?

An alternate option might be to skip the fetches if the [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) environment variable is present, indicating that a reproducible build was requested.

dsanders11:
> Yeah that sounds good; the release data will be re-fetched at runtime right?

It will be, although I'm not sure how the UI will react if it's started up with an empty `releases.json`, I'll have to test that when I make the change. There's also an open issue (#1222) where the intellisense for Node.js typings only uses the static version of `releases.json`, but I have a fix coming for that and should have a release which fixes that released in a couple weeks.

I'm curious, though, how does the credits page in settings look in the NixOS packaged version? That file is exclusively fetched at build-time, so if that fails due to no network I'd imagine the credits page is just empty.

Doing a cursory look at the `NixOS/nixpkgs` repo I see that packages can include patches - I wonder if the most robust solution would be make the change to gracefully fail if no network access, but then when you bump the Fiddle version for the package, use `yarn electron-releases` to populate `releases.json` and then dump that into a patch (same with contributors). That would effectively be what was being done before this PR landed, just without us needing to juggle generated PRs to keep that file up-to-date in this repo.

andersk:
> I'm curious, though, how does the credits page in settings look in the NixOS packaged version?

Yeah, it’s empty. (Maybe we could fallback with a link to https://github.com/electron/fiddle/graphs/contributors if there’s no pre-fetched data?)

<details>
<summary>Screenshots</summary>

credits
about

</details>

> Doing a cursory look at the `NixOS/nixpkgs` repo I see that packages can include patches -

Certainly an option, though it will complicate maintenance, meaning less frequent updates.

dsanders11:
Thanks for the screenshots!

I've tested building `main` with no internet access and it looks like the (unshipped) change in #1344 means the fetch releases build step gracefully fails, although there's a TS error since that JSON is inlined at build time so TS can tell it is an empty array. However, the app doesn't load properly if there are no known versions, [throwing an error at this line](https://github.com/electron/fiddle/blob/f06ebc37bb4bdbb6dafa70ad5af28dd2e6809742/src/renderer/versions.ts#L34).

I've put up a PR (#1351) to ensure it doesn't clobber any existing `static/releases.json` if fetching new releases fails, which ensures any patches to `static/releases.json` will be effective.

> Certainly an option, though it will complicate maintenance, meaning less frequent updates.

There's a lot changing in this repo at the moment as I'm landing a lot of refactors and fixes, leading to more releases at the moment. After another few releases the release cadence will likely slow down significantly as otherwise this repo doesn't get frequent changes.

So, v0.32.6 will be released this week (short of any hiccups) which will include gracefully handling network failure at build time. With that version, I'd recommend floating a patch to populate `static/releases.json` for the immediate future, and we can revisit this in the future after the current flurry of changes in this repo stabilize. As previously mentioned, it should be fine for that patch to get a little stale as the latest versions will be fetched on app start.

Thanks for bringing it to our attention, and for maintaining the NixOS package. 👍

PR #1318, opened 2023-04-17
ci: simplify release jobs

dsanders11:
I believe `load-gh-token` is only needed for the `publish-to-github` job since the others do dry runs and store to the workspace for the `publish-to-github` job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,6 @@ jobs:
     resource_class: macos.x86.medium.gen2
     steps:
       - install
-      - load-gh-token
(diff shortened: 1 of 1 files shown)


VerteDinde:
Could we try a test build run with these changes (either by pushing a tag or just commenting out the conditional code). I could be wrong about this, but I think we still need to gh-token for a hook in the `yarn run publish --dry-run` step, even though we don't do the full publish in that step. If I'm wrong and the build goes green, will approve 🙂 👍

VerteDinde (approved):
Thanks!

PR #1322, opened 2023-04-19
fix: properly handle clearing font size setting

dsanders11:
Fixes a bug where clearing the font size setting leads to "NaN" and can never be truly cleared.

Screenshot 2023-04-18 at 10 15 55 PM

### DIFF
--- a/src/renderer/components/settings-general-font.tsx
+++ b/src/renderer/components/settings-general-font.tsx
@@ -51,12 +51,13 @@ export class FontSettings extends React.Component<
   }
 
   /**
-   * Handles a change in the editor font family.
+   * Handles a change in the editor font size.
    *
    * @param {React.FormEvent<HTMLInputElement>} event
    */
   public handleSetFontSize(event: React.FormEvent<HTMLInputElement>): void {
-    const fontSize = parseInt(event.currentTarget.value, 10);
+    const parsedFontSize = parseInt(event.currentTarget.value, 10);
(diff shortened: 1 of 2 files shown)


src/renderer/components/settings-general-font.tsx
```diff
@@ -51,12 +51,13 @@ export class FontSettings extends React.Component<
...
   /**
-   * Handles a change in the editor font family.
+   * Handles a change in the editor font size.
    *
    * @param {React.FormEvent<HTMLInputElement>} event
    */
   public handleSetFontSize(event: React.FormEvent<HTMLInputElement>): void {
-    const fontSize = parseInt(event.currentTarget.value, 10);
+    const parsedFontSize = parseInt(event.currentTarget.value, 10);
+    const fontSize = isNaN(parsedFontSize) ? undefined : parsedFontSize;
```
BlackHole1, line 60:
```suggestion
    const fontSize = Number.isNaN(parsedFontSize) ? undefined : parsedFontSize;
```

```js
isNaN(undefined) // => true
Number.isNaN(undefined) // => false
```

Although parseInt does not return undefined, I recommend avoiding isNaN at all times for better code robustness.

BlackHole1 (changes requested):

BlackHole1 (approved):
LGTM

codebytere (approved):

PR #1323, opened 2023-04-20
fix: add Monaco theme

dsanders11:
Fixes adding Monaco themes, which currently don't take effect (and logs warnings). Can be tested using a theme from https://github.com/brijeshb42/monaco-themes/tree/master/themes.

Also fixes a couple of other small bugs around themes in general, refactors out common code, and expands test coverage.

### DIFF
--- a/src/renderer/components/dialog-add-theme.tsx
+++ b/src/renderer/components/dialog-add-theme.tsx
@@ -37,7 +37,6 @@ export const AddThemeDialog = observer(
       this.onSubmit = this.onSubmit.bind(this);
       this.onClose = this.onClose.bind(this);
       this.onChangeFile = this.onChangeFile.bind(this);
-      this.reset = this.reset.bind(this);
     }
 
     /**
@@ -120,8 +119,9 @@ export const AddThemeDialog = observer(
     }
 
     public onClose() {
-      this.props.appState.isThemeDialogShowing = false;
-      this.reset();
+      this.setState(this.resetState, () => {
+        this.props.appState.isThemeDialogShowing = false;
+      });
     }
 
     public render() {
@@ -150,14 +150,6 @@ export const AddThemeDialog = observer(
           </div>
         </Dialog>
       );
-    }
-
-    /**
-     * Reset this component's state
-     */
-    private reset(): void {
-      this.setState(this.resetState);
-      return;
     }
   },
 );
--- a/src/renderer/components/dialog-add-theme.tsx
+++ b/src/renderer/components/dialog-add-theme.tsx
@@ -1,16 +1,13 @@
 import * as React from 'react';
 
-import * as path from 'path';
-
 import { Button, Dialog, FileInput } from '@blueprintjs/core';
-import { shell } from 'electron';
 import * as fs from 'fs-extra';
 import { observer } from 'mobx-react';
 import * as MonacoType from 'monaco-editor';
 
 import { AppState } from '../state';
-import { THEMES_PATH, getTheme } from '../themes';
-import { LoadedFiddleTheme, defaultDark } from '../themes-defaults';
+import { createThemeFile, getTheme } from '../themes';
+import { FiddleTheme, defaultDark } from '../themes-defaults';
 
 interface AddThemeDialogProps {
   appState: AppState;
@@ -74,9 +71,12 @@ export const AddThemeDialog = observer(
         const editor = fs.readJSONSync(file.path);
         if (!editor.base && !editor.rules)
           throw Error('File does not match specifications'); // has to have these attributes
-        defaultTheme.editor = editor as Partial<MonacoType.editor.IStandaloneThemeData>;
-        const newTheme = defaultTheme;
-        const name = editor.name ? editor.name : file.name;
+        const newTheme: FiddleTheme = { ...defaultTheme };
+        newTheme.editor = editor as Partial<MonacoType.editor.IStandaloneThemeData>;
+        // Use file.name if no editor.name, and strip file extension (should be .json)
+        const name: string = editor.name
+          ? editor.name
+          : file.name.split('.').reverse().slice(1).reverse().join('.');
         await this.createNewThemeFromMonaco(name, newTheme);
       } catch (error) {
         appState.showErrorDialog(`${error}, please pick a different file.`);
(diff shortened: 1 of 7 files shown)


src/renderer/components/dialog-add-theme.tsx
```diff
@@ -74,9 +70,12 @@ export const AddThemeDialog = observer(
...
           throw Error('File does not match specifications'); // has to have these attributes
-        defaultTheme.editor = editor as Partial<MonacoType.editor.IStandaloneThemeData>;
-        const newTheme = defaultTheme;
-        const name = editor.name ? editor.name : file.name;
+        const newTheme: FiddleTheme = { ...defaultTheme };
+        newTheme.editor = editor as Partial<MonacoType.editor.IStandaloneThemeData>;
+        // Use file.name if no editor.name, and strip file extension (should be .json)
+        const name: string = editor.name
+          ? editor.name
+          : file.name.split('.').reverse().slice(1).reverse().join('.');
```
BlackHole1, line 78:
```suggestion
          : file.name.slice(0, file.name.lastIndexOf(".")) 
```

src/renderer/themes.ts
```diff
@@ -53,6 +55,47 @@ export async function readThemeFile(
...
+ * @returns {Promise<LoadedFiddleTheme>}
+ */
+export async function createThemeFile(
+  theme: FiddleTheme | LoadedFiddleTheme,
+  name?: string,
+): Promise<LoadedFiddleTheme> {
+  // Filter out file and css keys if they exist
+  theme = Object.fromEntries(
+    Object.entries(theme).filter(([key]) => !['file', 'css'].includes(key)),
+  ) as FiddleTheme;
```
BlackHole1, line 72:
```suggestion
  const { file: _file, css: _css, ...result } = theme;
  theme = result as FiddleTheme;
```

src/renderer/themes.ts
```diff
@@ -53,6 +55,47 @@ export async function readThemeFile(
...
+
+  if (!name) {
+    name = namor.generate({ words: 2, numbers: 0 });
+  }
```
BlackHole1, line 76:
```suggestion
  name = name || namor.generate({ words: 2, numbers: 0 });
```

tests/renderer/components/dialog-add-theme-spec.tsx
```diff
@@ -2,18 +2,38 @@ import * as React from 'react';
...
+class FileMock {
+  private bits: string[];
+  public readonly name: string;
+  public readonly path: string;
+
+  constructor(bits: string[], name: string, path: string) {
+    this.bits = bits;
+    this.name = name;
+    this.path = path;
+  }
```
BlackHole1, line 30:
```suggestion
class FileMock {
  constructor(
    private bits: string[],
    public name: string,
    public path: string
  ) {}
```

BlackHole1 (changes requested):
Improve code readability and some performance

src/renderer/themes.ts
dsanders11: That was my original approach, but it isn't applicable in this code base for two reasons:
* Fails the ESLint `@typescript-eslint/no-unused-vars` check
  * Although `.eslintrc.js` could be tweaked to use `varsIgnorePattern` to allow those
* Fails type checking: `Property 'file' does not exist on type 'LoadedFiddleTheme | FiddleTheme'. ts(2339)`

As such, I think the current code is the "least bad" solution, even though it's a bit ugly. Fighting TypeScript in this case can be a headache.

BlackHole1: > Fails the ESLint @typescript-eslint/no-unused-vars check 

```json
"@typescript-eslint/no-unused-vars": ["error", {
  "argsIgnorePattern": "^_",
}]
```

> Fails type checking: Property 'file' does not exist on type 'LoadedFiddleTheme | FiddleTheme'. ts(2339)

There are three solutions to this.

#### 1.

add `// @ts-ignore`.

```ts
// @ts-ignore
const { file: _file, css: _css, ...result } = theme;
```

#### 2.

add `as`

```ts
const { file: _file, css: _css, ...result } = theme as LoadedFiddleTheme & FiddleTheme;
```

#### 3.

add `OptionalMerge` type

```ts
type OptionalMerge<T1, T2> = {
    [K in keyof (T1 & T2)]: K extends keyof T1
        ? K extends keyof T2
            ? T1[K] | T2[K]
            : T1[K] | undefined
        : K extends keyof T2
            ? T2[K] | undefined
            : never
};

const { file: _file, css: _css, ...result } = theme as OptionalMerge<LoadedFiddleTheme, FiddleTheme>;
```

See: [typescriptlang-play](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8mwEsD2A7AhgGwLIQE4HMIAeAFQEYAaKEgJgD4oBeKAbwCgpOoBtAaSgQooAawggkAMygAKclABk1GgEoAugC4o-CAA9gEFABMAziLGTqZDlxsB+LVF36jp0eKm1rN71Hvk+qlAAPkoBXj5cmv68gSEArkYQEoIQhuE2mtp6BiZm7krpEX40AcFQCYZJKWkR3pooEABueGwAvgDcbGygkFAAgkys4ckYEJrGwLiC+OEAxsbG45PTw0hIS1MoM61dPdAAQoPsNvOLUBObMzYSaxsrNgBGaLh3W21ds6gTUMAAFhAAW2gzBYrSgaFMAxC+06bE+KG+LCgIzGUAA+iiqKdNGjTlQAHSE3AQYxxDDAKCtZh-QHQCGweDIdDYPCEIh9Kj7OjtIA)

BlackHole1: Personally, I recommend using priority is: 3 > 2 > 1

dsanders11: > add `// @ts-ignore`.

I don't consider this a valid solution, personally. 🙂

> add `as`

Since `LoadedFiddleTheme` extends `FiddleTheme`, no need for `as LoadedFiddleTheme & FiddleTheme`, just `as LoadedFiddleTheme`. Either way I don't like this option as the typing is then incorrect - destructuring the `name` property would have type `string` even though it may actually be `undefined` at run-time.

> add `OptionalMerge` type

This would work, but personally I don't think it's necessary or worth the tradeoff. `OptionalMerge` is not very readable code (that's just TypeScript types like that in general, they're higher cognitive load to parse). I think the existing code is fine, and doesn't require introducing `OptionalMerge` and changing the ESLint configuration.

BlackHole1 (approved):

PR #1324, opened 2023-04-20
refactor: move IPC event listeners to preload

dsanders11:
Push all listeners on `ipcRenderer` into an API in the preload script in preparation for context isolation. Penultimate PR before all usage of `ipcRenderer` has been moved to the preload script.


BlackHole1 (commented):
The code seems fine, but I'd like to discuss another approach: directly registering `window.ipcRenderer` in `preload.js`. Would that be better? This way, in the renderer process code, the `ipcRenderer.on(IpcEvents.FS_SAVE_FIDDLE_GIST, this.handleClick)` would become `window.ipcRenderer.on(IpcEvents.FS_SAVE_FIDDLE_GIST, this.handleClick)`.
In fact, there's no need to change anything else - just replace `ipcRenderer` with `window.ipcRenderer` in Weback.

dsanders11:
> directly registering window.ipcRenderer in preload.js. Would that be better?

That would go against the Electron documentation regarding context isolation, as [it's specifically called out as something not to do, for security reasons](https://www.electronjs.org/docs/latest/tutorial/context-isolation#security-considerations). Directly exposing `ipcRenderer` is too powerful from a security standpoint. Exposing just IPC listeners (and no sending) is less risky (if you strip the IPC event), but overall the safest approach is to limit the surface area to a few APIs which validate inputs (e.g., you can only listen to events defined in `channelMapping`).

There's also a benefit in adding a layer on top of IPC rather than directly exposing it since it can make refactors easier, and the general event pattern in this PR could be used for non-IPC events as well, such as serving as the basis for an [event bus](https://dawchihliou.github.io/articles/event-bus-for-react)/PubSub pattern.

BlackHole1:
> That would go against the Electron documentation regarding context isolation, as [it's specifically called out as something not to do, for security reasons](https://www.electronjs.org/docs/latest/tutorial/context-isolation#security-considerations). Directly exposing ipcRenderer is too powerful from a security standpoint. Exposing just IPC listeners (and no sending) is less risky (if you strip the IPC event), but overall the safest approach is to limit the surface area to a few APIs which validate inputs (e.g., you can only listen to events defined in channelMapping).

I mean is:

```
contextBridge.exposeInMainWorld('ipcRenderer', {
  on: (evetName, handler) => {
    // check eventName
    // ...

    ipcRenderer.on(name);
  }
})
```

> There's also a benefit in adding a layer on top of IPC rather than directly exposing it since it can make refactors easier, and the general event pattern in this PR could be used for non-IPC events as well, such as serving as the basis for an [event bus](https://dawchihliou.github.io/articles/event-bus-for-react)/PubSub pattern.

This sentence persuaded me. 👍

BlackHole1 (approved):

codebytere (approved):

PR #1346, opened 2023-04-28
fix: use yarn install if no modules to add

dsanders11:
`yarn add` run by itself causes an error, if there are no specific modules to add, you need to use `yarn install`.


codebytere (approved):

Sammiegirl411:
Rin

PR #1348, opened 2023-05-02
feat: add Fiddle snap package

VerteDinde:
WIP. Adds a Fiddle snap distributable and tests out our CI setup for snap development.


.circleci/config.yml
```diff
@@ -118,10 +118,16 @@ jobs:
         type: enum
         enum: [ 'x64', 'arm64','armv7l' ]
     docker:
-      - image: cimg/base:stable
+      - image: ubuntudesktop/gnome-3-38-2004
```
codebytere, line 121: What determines what version we use here? I found [this](https://hub.docker.com/layers/ubuntudesktop/gnome-3-38-2004/latest/images/sha256-f22e93b35de184d01b46c99e6c3975ea7aa0391d7ffde61404809029cb9a919b?context=explore) which looks like it's not been updated in ~2 years - could that present a problem in future?

.circleci/snapcraft-wrapper
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
```
codebytere, line 2: (nit) we can simplify this a bit:

```sh
#!/bin/sh
export SNAP="/snap/snapcraft/current"
export SNAP_NAME="$(awk '/^name:/{print $2}' $SNAP/meta/snap.yaml)"
export SNAP_VERSION="$(awk '/^version:/{print $2}' $SNAP/meta/snap.yaml)"
export SNAP_ARCH="amd64"

exec "$SNAP/usr/bin/python3" "$SNAP/bin/snapcraft" "$@"
```

codebytere (approved):
Looks generally good to me! A few comments but they're non-blocking.

dsanders11 (approved):

PR #1351, opened 2023-05-10
build: don't clobber existing releases.json on fetch failure

dsanders11:
Better handles things like doing a build with no network access after previously having done a build (clobbers `static/releases.json` at the moment) and allows for builds in a clean environment with no network access to patch `static/releases.json` and not have it clobbered.

### DIFF
--- a/tools/fetch-releases.js
+++ b/tools/fetch-releases.js
@@ -11,10 +11,20 @@ async function populateReleases() {
     elves.getReleaseInfo(version),
   );
 
-  console.log(`Updating local releases.json with ${releases.length} versions.`);
+  if (releases.length) {
+    console.log(
+      `Updating local releases.json with ${releases.length} versions.`,
+    );
 
-  await fs.remove(file);
-  await fs.outputJSON(file, releases);
+    await fs.remove(file);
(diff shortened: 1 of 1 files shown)


tools/fetch-releases.js
```diff
@@ -11,10 +11,20 @@ async function populateReleases() {
...
-  console.log(`Updating local releases.json with ${releases.length} versions.`);
+  if (releases.length) {
+    console.log(
+      `Updating local releases.json with ${releases.length} versions.`,
+    );
 
-  await fs.remove(file);
-  await fs.outputJSON(file, releases);
+    await fs.remove(file);
+    await fs.outputJSON(file, releases);
```
BlackHole1, line 20:
```suggestion
    await fs.outputJSON(file, releases);
```

The `outputJson` function uses the `fs.writeFile` API internally, and the default flag for `fs.writeFile` is `"w"`. This flag means that if the file already exists, it will be overwritten without needing to be deleted first.

BlackHole1 (changes requested):

BlackHole1 (approved):

PR #1358, opened 2023-05-15
refactor: split utils into main and renderer

dsanders11:
It's getting difficult to tell which utils are safe to use in the renderer (i.e. no Node.js usage), so separate out utils to make it clear what the expectations are for each file.

`editor-utils.ts` is still a bit tangled, the separation will be further cleaned up in future PRs which move more code to the main process.


src/main/utils/get-project-name.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @param {string} [localPath]
+ * @returns {Promise<string>}
+ */
+export async function getProjectName(localPath?: string): Promise<string> {
+  if (localPath) {
+    return path.basename(localPath);
+  }
+
+  return namor.generate({ words: 3, numbers: 0 });
+}
```
BlackHole1, line 17:
```suggestion
/**
 * Returns a name for this project
 *
 * @param {string} [localPath]
 * @returns {string}
 */
export function getProjectName(localPath?: string): string {
  if (localPath) {
    return path.basename(localPath);
  }

  return namor.generate({ words: 3, numbers: 0 });
}
```

BlackHole1: `namor.generate` return type is `string` not `Promise`

BlackHole1 (changes requested):

BlackHole1 (approved):
LGTM

Issue #1364, opened 2023-06-01
feature request: protocol api example


SimonSiefke:
Electron fiddle already has lots of great examples for using different electron apis. I'm thinking a  [protocol api](https://www.electronjs.org/docs/latest/api/protocol) example could be a useful addition to the Electron fiddle examples.

Based on the electron documentation, this is how the example could look like:

```js
const { app, BrowserWindow, protocol } = require('electron')


function createWindow () {
  const mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {}
  })
  mainWindow.loadURL('some-protocol://index.html')

}


app.whenReady().then(() => {
  protocol.handle('some-protocol', () => {
    return new Response(
      Buffer.from('<h1>Hello from protocol handler</h1>'), // Could also be a string or ReadableStream.
      { headers: { 'content-type': 'text/html' } }
    );
  });

  createWindow()
})
```

PR #1365, opened 2023-06-01
feat: add protocol api example

SimonSiefke:
Note: Since this uses protocol.handle, an upgrade to electron 25 would be required

fixes https://github.com/electron/fiddle/issues/1364


codebytere:
@SimonSiefke Can this be closed?

SimonSiefke:
Merged would be better. ^^

Issue #1366, opened 2023-06-02
Collision While Setting Versions


dsanders11:
Under certain circumstances, such as opening an example from the docs, multiple versions end up being set in parallel - the original version on startup, immediately followed by the version specified for the example being loaded. Once they're running in parallel, they can collide on setting the active version on disk with `fiddle-core`.

This situation could be improved by:
* Not unzipping a version until it is run
* Keeping track of any in-progress versions being set and the ability to abort them - ideally also aborting the download, but that is not supported by `@electron/get` at this time. Aborting an in-progress version being set should at least not conflict with any subsequent versions set.

<img width="686" alt="Screenshot 2023-06-01 at 10 32 41 PM" src="https://github.com/electron/fiddle/assets/5820654/b5b639f4-7e7e-48c5-8347-515adc3e89bc">

MohamedAli00949:
Hi @dsanders11 , 

I am really interested in solving this issue, so I have started with a suggested code to provide the abarting into `@electron/get`. I hope you take a look at it [here](https://github.com/electron/get/pull/365)

I really appreciate any help you can provide.

PR #1367, opened 2023-06-02
fix: don't install version until run

dsanders11:
Refs #1366.

Don't do the install step for a version until the user tries to run a fiddle. Doing so avoids collision on disk from multiple versions being installed at once (see reference issue). Also provides a nicer UX by not showing the unzipping spinner every time on start. Trade off is slightly longer initial run for a version since it needs to be unzipped and installed, but subsequent runs with the same version skip that step.

Should hopefully clear up a number of the disk errors being seen on Sentry around version management.

This partially reverts #1242 since that behavior becomes the only behavior.


BlackHole1 (approved):
LGTM

codebytere (approved):

PR #1376, opened 2023-06-21
fix: update settings in all open windows

### LINKED ISSUE #1352: Settings Changes Not Reflected in Other Windows
If you have multiple windows open, updating a setting in one window is not reflected in the other windows. Came across this when I updated my GitHub PAT in one window and it wasn't used in the others.

Most visible repro case would be to change the theme in one window and observe that other windows remain on the previous theme.

Suggested plan of attack based off of some brief consideration (alternatives welcome):

1. Add an `enum SettingKey` to `src/interfaces.ts` which lists all keys used to store settings in `localStorage`
2. Update all usages of `localStorage.getItem`/`localStorage.setItem` to use those enum values
3. In `AppState.save`, send an IPC event to main process, e.g. `window.ElectronFiddle.settingsChanged()`
4. In main process, listen for that IPC event, then broadcast a "refresh-settings" event to all windows except the sender
5. In `AppState` constructor, listen for that event (`window.ElectronFiddle.addEventListener('refresh-settings', ...)`) and loop through the enum keys, with a switch statement for key values and retrieve the latest value from `localStorage` and update the state accordingly for that setting

That should ensure exhaustiveness checking for the persisted settings to ensure we don't miss any during future changes.

erikian:
Here's my stab at #1352. I mostly followed the approach suggested in that issue, except that I'm listening for updates in the settings using the [`storage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) event instead of using IPC, so we can get the same result for free without touching the main process.

Just a consideration: any open windows (except the one in which the setting was changed) updates its `AppState` object in response to a setting being changed. However, this PR doesn't (directly) change anything related to the UI state. As you can see in the video below, if I have two windows with the Settings modal opened and I make changes in one of them, checkboxes are checked/unchecked in both windows, while text inputs don't change in the other window (unless I close and reopen the Settings modal manually). This is purely a side effect of how the respective components handle their state; if "simultaneous edition" is desired, it will require some changes (maybe just triggering a re-render of the Settings modal on focus, or even changing how the components respond to updates in the global state) that might or might not be part of the scope here, so I decided to check before going any further.

https://github.com/electron/fiddle/assets/19478575/b6207a00-258e-4840-a5fd-631155b818c3

Fixes #1352.

### DIFF
--- a/src/renderer/components/settings-execution.tsx
+++ b/src/renderer/components/settings-execution.tsx
@@ -12,15 +12,19 @@ import {
 } from '@blueprintjs/core';
 import { observer } from 'mobx-react';
 
-import { IPackageManager, Setting, SettingKey } from '../../interfaces';
+import {
+  GlobalSetting,
+  GlobalSettingKey,
+  IPackageManager,
+} from '../../interfaces';
 import { AppState } from '../state';
 
 /**
  * @TODO make this a proper enum again once we update Typescript
  */
 export const SettingItemType = {
-  EnvVars: Setting.environmentVariables,
-  Flags: Setting.executionFlags,
+  EnvVars: GlobalSetting.environmentVariables,
+  Flags: GlobalSetting.executionFlags,
 } as const;
 
 interface ExecutionSettingsProps {
@@ -122,7 +126,7 @@ export const ExecutionSettings = observer(
      */
     public handleSettingsItemChange(
       event: React.ChangeEvent<HTMLInputElement>,
-      type: SettingKey,
+      type: GlobalSettingKey,
     ) {
       const { name, value } = event.currentTarget;
 
@@ -139,7 +143,7 @@ export const ExecutionSettings = observer(
      *
      * @param {SettingItemType} type
      */
-    private addNewSettingsItem(type: SettingKey) {
+    private addNewSettingsItem(type: GlobalSettingKey) {
       const array = Object.entries(this.state[type]);
 
       this.setState((prevState) => ({
@@ -163,7 +167,7 @@ export const ExecutionSettings = observer(
       appState.packageManager = value as IPackageManager;
     };
 
-    public renderDeleteItem(idx: string, type: SettingKey): JSX.Element {
+    public renderDeleteItem(idx: string, type: GlobalSettingKey): JSX.Element {
       const updated = this.state[type];
 
       const removeFn = () => {
--- a/src/renderer/components/settings-execution.tsx
+++ b/src/renderer/components/settings-execution.tsx
@@ -12,13 +12,16 @@ import {
 } from '@blueprintjs/core';
 import { observer } from 'mobx-react';
 
-import { IPackageManager } from '../../interfaces';
+import { IPackageManager, Setting, SettingKey } from '../../interfaces';
 import { AppState } from '../state';
 
-export enum SettingItemType {
-  EnvVars = 'environmentVariables',
-  Flags = 'executionFlags',
-}
+/**
+ * @TODO make this a proper enum again once we update Typescript
+ */
+export const SettingItemType = {
+  EnvVars: Setting.environmentVariables,
+  Flags: Setting.executionFlags,
+} as const;
 
 interface ExecutionSettingsProps {
   appState: AppState;
@@ -119,7 +122,7 @@ export const ExecutionSettings = observer(
      */
     public handleSettingsItemChange(
       event: React.ChangeEvent<HTMLInputElement>,
-      type: SettingItemType,
+      type: SettingKey,
     ) {
       const { name, value } = event.currentTarget;
 
@@ -136,7 +139,7 @@ export const ExecutionSettings = observer(
      *
      * @param {SettingItemType} type
      */
-    private addNewSettingsItem(type: SettingItemType) {
+    private addNewSettingsItem(type: SettingKey) {
       const array = Object.entries(this.state[type]);
 
       this.setState((prevState) => ({
@@ -160,7 +163,7 @@ export const ExecutionSettings = observer(
       appState.packageManager = value as IPackageManager;
     };
 
-    public renderDeleteItem(idx: string, type: SettingItemType): JSX.Element {
+    public renderDeleteItem(idx: string, type: SettingKey): JSX.Element {
       const updated = this.state[type];
 
       const removeFn = () => {
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -21,16 +21,18 @@ import {
   GenericDialogOptions,
   GenericDialogType,
   GistActionState,
+  GlobalSetting,
+  GlobalSettingKey,
   IPackageManager,
   InstallState,
   OutputEntry,
   OutputOptions,
   RunnableVersion,
   SetFiddleOptions,
-  Setting,
-  SettingKey,
   Version,
   VersionSource,
+  WindowSpecificSetting,
+  WindowSpecificSettingKey,
 } from '../interfaces';
 import { Bisector } from './bisect';
 import { ELECTRON_DOWNLOAD_PATH, ELECTRON_INSTALL_PATH } from './constants';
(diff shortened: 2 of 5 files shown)


dsanders11:
Will give this a more thorough review later.

> I mostly followed the approach suggested in that issue, except that I'm listening for updates in the settings using the [storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) event instead of using IPC, so we can get the same result for free without touching the main process.

Nice! Wasn't aware of that event. That makes sense and simplifies things.

> However, this PR doesn't (directly) change anything related to the UI state. As you can see in the video below, if I have two windows with the Settings modal opened and I make changes in one of them, checkboxes are checked/unchecked in both windows, while text inputs don't change in the other window (unless I close and reopen the Settings modal manually). This is purely a side effect of how the respective components handle their state; if "simultaneous edition" is desired, it will require some changes (maybe just triggering a re-render of the Settings modal on focus, or even changing how the components respond to updates in the global state) that might or might not be part of the scope here, so I decided to check before going any further.

This seems fine for now. 👍

src/renderer/components/settings-execution.tsx
```diff
@@ -12,13 +12,20 @@ import {
...
-  EnvVars = 'environmentVariables',
-  Flags = 'executionFlags',
-}
+/**
+ * @TODO make this a proper enum again once we update Typescript
+ */
+export const SettingItemType = {
+  EnvVars: GlobalSetting.environmentVariables,
+  Flags: GlobalSetting.executionFlags,
+} as const;
```
dsanders11, line 28: Could you provide further context here on what the issue is, and which version of TypeScript will fix it?

src/renderer/state.ts
```diff
@@ -329,41 +351,110 @@ export class AppState {
...
+
+      try {
+        parsedValue = JSON.parse(newValue as string) as unknown;
+      } catch {
+        // The new value is a plain string, not a well-formed stringified object.
+        parsedValue = newValue;
+      }
+
+      if (key in GlobalSetting && key in this) {
+        // Any settings listed here need morethan just updating the state directly.
```
dsanders11, line 372:
```suggestion
        // Any settings listed here need more than just updating the state directly.
```

src/renderer/state.ts
```diff
@@ -329,41 +351,110 @@ export class AppState {
...
+
+        const action = actions[key];
+
+        if (typeof action === 'function') {
+          action();
+        } else {
+          // Fall back to updating the state.
+          this[key] = parsedValue;
+        }
+      }
```
dsanders11, line 387: Maybe if the key isn't recognized, a warning should be logged? Just in case something slips in without being added to the enum properly.

src/renderer/state.ts
```diff
@@ -329,41 +351,110 @@ export class AppState {
...
+        };
+
+        const action = actions[key];
+
+        if (typeof action === 'function') {
+          action();
+        } else {
+          // Fall back to updating the state.
+          this[key] = parsedValue;
+        }
```
dsanders11, line 386: The idea of using an enum for keys was to leverage exhaustiveness checking so that if a new setting is added, there will be a TS error until the key is handled here, to prevent missing a needed action (like `setTheme`). It's more verbose, but makes it more difficult to forget to handle it properly.

So originally I was thinking something like:

```js
switch(key) {
  case GlobalSetting.theme:
    this.setTheme(parsedValue as string);
    break;
  case GlobalSetting.fontFamily:
  case GlobalSetting.fontSize:
    this[key] = parsedValue;
}
```

Thoughts on that approach versus the one you've implemented here?

dsanders11 (commented):
I'd like to play with this some more when I get the time next week, but dropped some initial review comments.

src/renderer/components/settings-execution.tsx
erikian: Sure. The problem is I can't compute the values of the `SettingItemType` enum using another enum. We used to have the enum values as string literals...
```ts
export enum SettingItemType {
  EnvVars = 'environmentVariables',
  Flags = 'executionFlags',
}
```
...but since these values are keys used in the localStorage, I moved them to the `GlobalSetting` enum, so I'm trying to keep these values in sync by using the value from `GlobalSetting` in `SettingItemType`...

```ts
enum GlobalSetting {
  // ...
  environmentVariables = 'environmentVariables',
  executionFlags = 'executionFlags',
  // ...
}

export enum SettingItemType {
  EnvVars = GlobalSetting.environmentVariables,
  Flags = GlobalSetting.executionFlags,
}
```
...but Typescript won't let me do this using `SettingItemType` as an enum:

image

I've changed the signature of some methods that deal with the localStorage to accept a `GlobalSettingKey` (the union of all possible keys in `GlobalSetting`) instead of a plain `string`, and Typescript gets mad at me. However, it works fine if I use an object instead of an enum, so I did that.

Apparently, this is one of the "long-standing oddities around enums" mentioned in the [5.0 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#enum-overhaul), so changing `SettingItemType` back to an enum would require TS 5.0.

We could also just get rid of `SettingItemType` and refactor any usages to grab the values from `GlobalSetting` directly, but since that enum was already there, I didn't want to do that just yet to keep the changes as directed as possible.

src/renderer/state.ts
erikian: I missed the bit about having a compile-time error when a new setting is added, it makes total sense. We could even do that using my object approach by doing something like this...

```ts
const actions: Record<GlobalSettingKey, null | (() => void)> = {
  acceleratorsToBlock: null,
  channelsToShow: null,
  electronMirror: null,
  environmentVariables: null,
  executionFlags: null,
  fontFamily: null,
  fontSize: null,
  gitHubAvatarUrl: null,
  gitHubLogin: null,
  gitHubName: null,
  gitHubToken: null,
  hasShownTour: null,
  isClearingConsoleOnRun: null,
  isEnablingElectronLogging: null,
  isKeepingUserDataDirs: null,
  isPublishingGistAsRevision: null,
  isUsingSystemTheme: null,
  knownVersion: null,
  localVersion: null,
  packageAuthor: null,
  packageManager: null,
  showObsoleteVersions: null,
  showUndownloadedVersions: null,
  theme: () => {
    this.setTheme(parsedValue as string);
  },
};
```

...but that would still require that `action` validation, so I think using a `switch` with a type guard looks cleaner:

```ts
switch (key) {
  case 'theme': {
    this.setTheme(parsedValue as string);
    break;
  }

  case 'acceleratorsToBlock':
  case 'channelsToShow':
  case 'electronMirror':
  case 'environmentVariables':
  case 'executionFlags':
  case 'fontFamily':
  case 'fontSize':
  case 'gitHubAvatarUrl':
  case 'gitHubLogin':
  case 'gitHubName':
  case 'gitHubToken':
  case 'hasShownTour':
  case 'isClearingConsoleOnRun':
  case 'isEnablingElectronLogging':
  case 'isKeepingUserDataDirs':
  case 'isPublishingGistAsRevision':
  case 'isUsingSystemTheme':
  case 'knownVersion':
  case 'localVersion':
  case 'packageAuthor':
  case 'packageManager':
  case 'showObsoleteVersions':
  case 'showUndownloadedVersions': {
    // Fall back to updating the state.
    this[key] = parsedValue;
    break;
  }

  default: {
    settingKeyTypeGuard(key);
  }
}

// ...

function settingKeyTypeGuard(key: never): never {
  throw new Error(
    `Unhandled setting ${key}, please handle it in the \`AppState\`.`,
  );
}
```

dsanders11: Sounds good to me. 👍

src/renderer/state.ts
erikian: 🤔 There's another issue with keeping settings in sync between windows: I've also created a `WindowSpecificSetting` enum with settings meant to be window-specific rather than global (like the Electron version you want to use in that window). If I change the Electron version in the first window, the second window won't update its state right away because I only do that if the key is a member of `GlobalSetting`.

However, if the second window is reloaded for some reason, it will populate its state based on the localStorage and will pick up the Electron version that was set by the first window. Not sure if this is a common scenario, but we might need to ensure every window can save its own settings independently (maybe saving an object instead of a string and using the browserWindow ID as a key, like `localStorage.setItem(WindowSpecificSetting.version, JSON.stringify({ browserWindow1Id: '24.0.0', browserWindow2Id: '25.0.0' }))` or something like that). Do you think this should be addressed here?

dsanders11: I think it's fine to ignore that scenario in this PR, seems reasonable enough that reloading a window (and thus losing all window-specific state) will default to the last Electron version set (whichever window it was set in). Trying to store window-specific settings in `localStorage` for a specific window seems overly complex for what is gained in practice.

If you want to go down that path in a follow-up PR, I'd suggest looking at using `sessionStorage` for `WindowSpecificSetting` - it will be specific to each window, and survive reloads. We'd probably want to layer it so that each window always sets the key in `sessionStorage`, as well as `localStorage`, and tries to read first from `sessionStorage` and then falls back to `localStorage` if it's not found (like on initial page load).

dsanders11 (approved):

dsanders11:
Nice work on this @erikian, and thanks for grabbing the issue!

Issue #405, opened 2020-05-28
Support appimage releases


nom3ad:
Consider appimage as a release option for supporting wide variety  of Linux distros.

PR #1379, opened 2023-06-27
feat: support AppImage releases

robertgzr:
using @reforged/maker-appimage

Closes: https://github.com/electron/fiddle/issues/405
Signed-off-by: Robert Günzler <r@gnzler.io>

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -766,7 +766,7 @@
     username "^5.1.0"
     yarn-or-npm "^3.0.1"
 
-"@electron-forge/maker-base@6.1.1":
+"@electron-forge/maker-base@6.1.1", "@electron-forge/maker-base@^6.0.4":
   version "6.1.1"
   resolved "https://registry.yarnpkg.com/@electron-forge/maker-base/-/maker-base-6.1.1.tgz#2c8fdaaaa28ff39fbf4ff26e664dc79265f96b49"
   integrity sha512-qEW8vuSq6/r1rKh7JENK9yGmueymMzbm817GDk5usHia5o7otHYY4JMkdzZaKww56Ed/Ege2ch6LEYqYGTfl/Q==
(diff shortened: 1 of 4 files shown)


erickzhao:
@robertgzr I think it would be fine to upgrade the Node.js version to the latest version of 18 LTS here:

https://github.com/electron/fiddle/blob/a0b94dce005a923078f14831941342578eefdacf/.circleci/config.yml#L11-L13

dsanders11:
Looks like the test failures with Node.js v18.16.x are due to https://github.com/nodejs/node/issues/47563, the fix for which may land soon in v18.16.2 - in the mean time let's use `18.15` for this PR, since that line doesn't have the bug (confirmed locally, and it's mentioned in the last comment on that issue).

yarn.lock
```diff
@@ -775,6 +775,15 @@
     fs-extra "^10.0.0"
     which "^2.0.2"
 
+"@electron-forge/maker-base@^6.0.4":
+  version "6.2.1"
+  resolved "https://registry.yarnpkg.com/@electron-forge/maker-base/-/maker-base-6.2.1.tgz#04dcc6163a2938480ed5c887fa2894646c7eebb8"
+  integrity sha512-LnvGtTJ/RNojKdUKktYEcbLqPggXdMBs1uscQRgXkI3XnVGdEi+/j5+Eg5ka4d6FnsaUkz//U5yhPtNFhDbNSw==
+  dependencies:
+    "@electron-forge/shared-types" "6.2.1"
```
robertgzr, line 783: isn't it a bit weird that `maker-base` specifically wants the latest version of `shared-types` rather than `^6.0.4`?

dsanders11:
> We could use just `18.16` for the changes to the CircleCI config, to ensure it uses the latest fix release

Ugh, this is my bad - I forgot there's an issue where `nvm-windows` doesn't handle version like that, only the full one - we should add a comment in the CircleCI config. So this should use `18.15.0`. Apologies for the churn.

yarn.lock
dsanders11: The `@electron-forge/*` packages are specifically version locked to each other, so that's why `@electron-forge/maker-base@6.2.1` only wants `@electron-forge/shared-types@6.2.1`.

Had some discussion going on about the best way to prevent the usage of `@electron-forge/maker-base@6.2.1` here (which might be harmless, but I'd prefer to avoid any subtle bugs, already enough foot guns). Opening #1381 to address that and this PR can rebase once that lands.

robertgzr: nice, thanks :)

dsanders11:
> we get nvm from the windows orb, right? how would we know which version that is on 😅

That comes from the CircleCI images where they bake new ones every once in a while, looks like they [baked a new Windows Server 2022 one in May 2023](https://discuss.circleci.com/t/windows-server-2022-image-released-may-2023/48072) - `win/default` defaults to `win/server-2022` currently. Clicking through that forum post to the commit, looks like [this is the file](https://github.com/CircleCI-Public/ansible/blob/main/roles/windows-executor/defaults/main/install_dev_tools_defaults.yml) which defines the versions of dev tools, including `nvm-windows`, and it's currently on `1.1.9`. 😅

robertgzr:
I included a comment mentioning the nvm-windows situation and linking to that ansible file :)

dsanders11 (approved):

codebytere (approved):

dsanders11:
@robertgzr, now that this has shipped, a couple of observations:

* We should get the icon working properly - I played with it for a bit but wasn't successful
* The AppImage release doesn't work with launching Fiddle from the protocol handler via the docs

dsanders11:
> I included a comment mentioning the nvm-windows situation and linking to that ansible file :)

I poked CircleCI and they've updated `nvm-windows` [on the latest Windows Server 2022 image](https://discuss.circleci.com/t/windows-server-2022-august-23-edge-image-published/48818), but it hasn't made it's way to being the `current` tag yet. Confirmed it resolves the issue: https://app.circleci.com/pipelines/github/electron/fiddle/991/workflows/5321ebad-4d4d-4ce3-bb65-fa45ec243649/jobs/3594

PR #1380, opened 2023-06-28
build: cleanup package.json dependencies

dsanders11:
Should have no effect other than cleaning up `package.json` and `yarn.lock`.

* Removes unused `@octokit/action` dev dependency
* Adds `webpack`, `@octokit/core`, and `@testing-library/dom` dev dependencies to clear warnings about unmet peer dependencies


codebytere (approved):

codebytere:
@dsanders11 needs conflict cleanup

PR #1384, opened 2023-07-03
perf: import modules in parallel during app setup

dsanders11:
In theory this should result in better performance, in practice when trying to do some quick and dirty profiling the results were inconclusive. Doing these with `Promise.all` may smooth overall timings by ensuring one call taking a bit longer than usual for whatever reason doesn't slow all calls.

Since this is low risk, going to YOLO it under the assumption that at best it'll help things, at worse it's a no-op.


BlackHole1 (approved):
LGTM. I just checked, this writing style is fine. You can use it with confidence.

image

PR #1386, opened 2023-07-03
fix: improve behavior of setting sync between windows

dsanders11:
Follow-up to #1376 which fixes a few bugs:

* Don't warn when `WindowSpecificSetting` settings are touched
* Fix cases where the enum value does not match the enum key
* Better handle several keys which do not exist on `AppState`
* Sync local versions properly between windows
* Update some JSDoc types which were changed in #1376

cc @erikian

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -22,7 +22,6 @@ import {
   GenericDialogType,
   GistActionState,
   GlobalSetting,
-  GlobalSettingKey,
   IPackageManager,
   InstallState,
   OutputEntry,
@@ -32,7 +31,6 @@ import {
   Version,
   VersionSource,
   WindowSpecificSetting,
-  WindowSpecificSettingKey,
 } from '../interfaces';
 import { Bisector } from './bisect';
 import { ELECTRON_DOWNLOAD_PATH, ELECTRON_INSTALL_PATH } from './constants';
@@ -45,6 +43,7 @@ import {
   fetchVersions,
   getDefaultVersion,
   getElectronVersions,
+  getLocalVersions,
   getReleaseChannel,
   makeRunnable,
   saveLocalVersions,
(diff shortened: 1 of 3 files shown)


src/renderer/state.ts
```diff
@@ -374,36 +373,50 @@ export class AppState {
...
+          case GlobalSetting.electronMirror:
+          case GlobalSetting.environmentVariables:
+          case GlobalSetting.executionFlags:
+          case GlobalSetting.fontFamily:
+          case GlobalSetting.fontSize:
+          case GlobalSetting.gitHubAvatarUrl:
+          case GlobalSetting.gitHubLogin:
+          case GlobalSetting.gitHubName:
+          case GlobalSetting.gitHubToken:
+          case GlobalSetting.hasShownTour:
```
erikian, line 410: This is already handled elsewhere, so:

```suggestion
```

src/renderer/state.ts
```diff
@@ -1055,11 +1089,11 @@ export class AppState {
   /**
    * Save a key/value to localStorage.
    *
-   * @param {string} key
+   * @param {GlobalSetting | WindowSpecificSetting} key
```
erikian, line 1092: Not directly related to this PR, but I wonder how important JSDoc types are in a Typescript project (also sorry for the oversight!)

dsanders11: Agreed, I'm not sure the history of how this codebase ended up with them and TypeScript, but since it does have them at the moment, I try to keep them up-to-date. 😅 I think at some point it probably makes sense to just rip out the types for params since that's duplicating with TS, but we can leave things like `@param foo - this is a description of foo` since that's still useful.

src/renderer/state.ts
dsanders11: Oddly making this change caused TypeScript to complain that about `this[key] = parsedValue;` assigning `unknown` to `never`. 🤔 There's some weird TS behavior around this bit (this PR actually pops out of me trying to upgrade TS), but I've added 2e19f042a85c786f3cda04d38d944f0120ccba2a to silence the TS error.

codebytere (approved):

PR #1391, opened 2023-07-07
feat: sync version download statuses between windows

erikian:
This allows the download status and progress to be shared between windows:

https://github.com/electron/fiddle/assets/19478575/6b666696-e298-4b4a-8236-ed2c8d238be7

I also tried to make the BroadcastChannel logic extensible so we can use it to sync other data in the future.

### DIFF
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -209,7 +209,7 @@ export class AppState {
   });
 
   // Used for communications between windows
-  public broadcastChannel: AppStateBroadcastChannel = new BroadcastChannel(
+  private broadcastChannel: AppStateBroadcastChannel = new BroadcastChannel(
     'AppState',
   );
 
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -214,10 +214,13 @@ export class AppState {
   );
 
   // Notifies other windows that this version has changed so they can update their state to reflect that.
-  private updateVersionState(version: RunnableVersion) {
+  private broadcastVersionStates(versions: RunnableVersion[]) {
     this.broadcastChannel.postMessage({
-      type: AppStateBroadcastMessageType.syncVersion,
-      payload: { ...version },
+      type: AppStateBroadcastMessageType.syncVersions,
+
+      // the RunnableVersion proxies can't be cloned by structuredClone,
+      // so we have to create plain objects out of them
+      payload: versions.map((version) => ({ ...version })),
     });
   }
 
@@ -461,8 +464,8 @@ export class AppState {
         const { type, payload } = event.data;
 
         switch (type) {
-          case AppStateBroadcastMessageType.syncVersion: {
-            this.addNewVersions([payload]);
+          case AppStateBroadcastMessageType.syncVersions: {
+            this.setVersionStates(payload);
 
             break;
           }
@@ -729,6 +732,15 @@ export class AppState {
 
   public addNewVersions(versions: RunnableVersion[]) {
     for (const ver of versions) {
+      this.versions[ver.version] ||= ver;
+    }
+
+    this.broadcastVersionStates(versions);
+  }
+
+  // Updates the version states in the current window to reflect updates made by other windows.
+  private setVersionStates(versions: RunnableVersion[]) {
+    for (const ver of versions) {
       this.versions[ver.version] = ver;
     }
   }
@@ -768,7 +780,7 @@ export class AppState {
 
           await typeDefsCleaner();
 
-          this.updateVersionState(ver);
+          this.broadcastVersionStates([ver]);
         }
       } else {
         console.log(`State: Version ${version} already removed, doing nothing`);
@@ -807,10 +819,12 @@ export class AppState {
 
     console.log(`State: Downloading Electron ${version}`);
 
-    this.updateVersionState({
-      ...ver,
-      state: InstallState.downloading,
-    });
+    this.broadcastVersionStates([
+      {
+        ...ver,
+        state: InstallState.downloading,
+      },
+    ]);
 
     // Download the version without setting it as the current version.
     await this.installer.ensureDownloaded(version, {
@@ -825,13 +839,13 @@ export class AppState {
           if (ver.downloadProgress !== percent) {
             ver.downloadProgress = percent;
 
-            this.updateVersionState(ver);
+            this.broadcastVersionStates([ver]);
           }
         });
       },
     });
 
-    this.updateVersionState(ver);
+    this.broadcastVersionStates([ver]);
   }
 
   /**
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -477,6 +477,11 @@ export class AppState {
         const { type, payload } = event.data;
 
         switch (type) {
+          case AppStateBroadcastMessageType.isDownloadingAll: {
+            this.isDownloadingAll = payload;
+            break;
+          }
+
           case AppStateBroadcastMessageType.syncVersions: {
             this.setVersionStates(payload);
 
@@ -647,10 +652,18 @@ export class AppState {
 
   public startDownloadingAll() {
     this.isDownloadingAll = true;
+    this.broadcastChannel.postMessage({
+      type: AppStateBroadcastMessageType.isDownloadingAll,
+      payload: true,
+    });
   }
 
   public stopDownloadingAll() {
     this.isDownloadingAll = false;
+    this.broadcastChannel.postMessage({
+      type: AppStateBroadcastMessageType.isDownloadingAll,
+      payload: false,
+    });
   }
 
   public startDeletingAll() {
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -16,6 +16,7 @@ import {
 } from 'mobx';
 
 import {
+  AppStateBroadcastChannel,
   AppStateBroadcastMessage,
   AppStateBroadcastMessageType,
   BlockableAccelerator,
@@ -214,7 +215,7 @@ export class AppState {
 
   // Notifies other windows that this version has changed so they can update their state to reflect that.
   private updateVersionDownloadStatus(version: RunnableVersion) {
-    this.broadcastChannel.postMessage<RunnableVersion>({
+    this.broadcastChannel.postMessage({
       type: AppStateBroadcastMessageType.syncVersion,
       payload: { ...version },
     });
@@ -461,7 +462,7 @@ export class AppState {
 
         switch (type) {
           case AppStateBroadcastMessageType.syncVersion: {
-            this.addNewVersions([payload as RunnableVersion]);
+            this.addNewVersions([payload]);
 
             break;
           }
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -214,7 +214,7 @@ export class AppState {
   );
 
   // Notifies other windows that this version has changed so they can update their state to reflect that.
-  private updateVersionDownloadStatus(version: RunnableVersion) {
+  private updateVersionState(version: RunnableVersion) {
     this.broadcastChannel.postMessage({
       type: AppStateBroadcastMessageType.syncVersion,
       payload: { ...version },
@@ -768,7 +768,7 @@ export class AppState {
 
           await typeDefsCleaner();
 
-          this.updateVersionDownloadStatus(ver);
+          this.updateVersionState(ver);
         }
       } else {
         console.log(`State: Version ${version} already removed, doing nothing`);
@@ -807,7 +807,7 @@ export class AppState {
 
     console.log(`State: Downloading Electron ${version}`);
 
-    this.updateVersionDownloadStatus({
+    this.updateVersionState({
       ...ver,
       state: InstallState.downloading,
     });
@@ -825,13 +825,13 @@ export class AppState {
           if (ver.downloadProgress !== percent) {
             ver.downloadProgress = percent;
 
-            this.updateVersionDownloadStatus(ver);
+            this.updateVersionState(ver);
           }
         });
       },
     });
 
-    this.updateVersionDownloadStatus(ver);
+    this.updateVersionState(ver);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


erikian:
Looks like the Node.js version of BroadcastChannel and Jest don't get along on CircleCI (though it works fine on my machine), the tests seem to run but Jest hangs after they're finished (maybe it actually creates a worker thread that keeps running?) Anyway, I mocked it out and now it works.

dsanders11:
> Looks like the Node.js version of BroadcastChannel and Jest don't get along on CircleCI (though it works fine on my machine), the tests seem to run but Jest hangs after they're finished (maybe it actually creates a worker thread that keeps running?) Anyway, I mocked it out and now it works.

Mocking is the correct thing to do here, rather than trying to use `BroadcastChannel` from `node:worker_threads` - we don't want to trigger any real behavior during the tests. 👍

dsanders11:
I think the typings are cleaner if you use a discriminated union on the possible message types - removes the need to do a type cast.

Also shouldn't be adding anything to `ambient.d.ts` for this, can all be accomplished with imports.

```diff
diff --git a/src/ambient.d.ts b/src/ambient.d.ts
index 18c01d1f..a9315109 100644
--- a/src/ambient.d.ts
+++ b/src/ambient.d.ts
@@ -2,7 +2,6 @@ import { ReleaseInfo } from '@electron/fiddle-core';
 import * as MonacoType from 'monaco-editor';
 
 import {
-  AppStateBroadcastMessage,
   BisectRequest,
   BlockableAccelerator,
   EditorValues,
@@ -126,8 +125,4 @@ declare global {
       themePath: string;
     };
   }
-
-  interface AppStateBroadcastChannel extends BroadcastChannel {
-    postMessage<T = unknown>(params: AppStateBroadcastMessage<T>): void;
-  }
 }
diff --git a/src/interfaces.ts b/src/interfaces.ts
index 6252c926..87933298 100644
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -221,11 +221,17 @@ export enum WindowSpecificSetting {
 
 export type WindowSpecificSettingKey = keyof typeof WindowSpecificSetting;
 
-export interface AppStateBroadcastMessage<T = unknown> {
-  type: AppStateBroadcastMessageType;
-  payload: T;
+export interface AppStateBroadcastChannel extends BroadcastChannel {
+  postMessage(params: AppStateBroadcastMessage): void;
 }
 
+// Use a discriminated union of types (there's only one here at the moment) so
+// that the payload can easily be typed correctly for each message type
+export type AppStateBroadcastMessage = {
+  type: AppStateBroadcastMessageType.syncVersion;
+  payload: RunnableVersion;
+};
+
 export enum AppStateBroadcastMessageType {
   syncVersion = 'syncVersion',
 }
diff --git a/src/renderer/state.ts b/src/renderer/state.ts
index a02bd1e5..0d1f95ca 100644
--- a/src/renderer/state.ts
+++ b/src/renderer/state.ts
@@ -16,6 +16,7 @@ import {
 } from 'mobx';
 
 import {
+  AppStateBroadcastChannel,
   AppStateBroadcastMessage,
   AppStateBroadcastMessageType,
   BlockableAccelerator,
@@ -215,7 +216,7 @@ export class AppState {
 
   // Notifies other windows that this version has changed so they can update their state to reflect that.
   private updateVersionDownloadStatus(version: RunnableVersion) {
-    this.broadcastChannel.postMessage<RunnableVersion>({
+    this.broadcastChannel.postMessage({
       type: AppStateBroadcastMessageType.syncVersion,
       payload: { ...version },
     });
@@ -445,7 +446,7 @@ export class AppState {
 
         switch (type) {
           case AppStateBroadcastMessageType.syncVersion: {
-            this.addNewVersions([payload as RunnableVersion]);
+            this.addNewVersions([payload]);
 
             break;
           }
```

erikian:
Done. To be extra thorough, we'd still need to type cast if we want the general case of multiple types to work with the current Typescript version — but since we only have one type right now, it's probably simpler to leave it as is and just add the type casts if we happen to add any other types before https://github.com/electron/fiddle/pull/1389 lands.

src/renderer/state.ts
```diff
@@ -208,6 +209,19 @@ export class AppState {
     versions: this.baseVersions,
   });
 
+  // Used for communications between windows
+  public broadcastChannel: AppStateBroadcastChannel = new BroadcastChannel(
+    'AppState',
+  );
+
+  // Notifies other windows that this version has changed so they can update their state to reflect that.
+  private updateVersionDownloadStatus(version: RunnableVersion) {
```
dsanders11, line 218: Let's name this a more generic `updateVersionState`, since it really updates the overall state - at the moment that might be tightly coupled to download status, but any change to the info that `RunnableVersion` contains could make this method name inaccurate.

src/renderer/state.ts
```diff
@@ -654,7 +696,7 @@ export class AppState {
 
   public addNewVersions(versions: RunnableVersion[]) {
     for (const ver of versions) {
-      this.versions[ver.version] ||= ver;
+      this.versions[ver.version] = ver;
```
dsanders11, line 699: I'm not a fan of this behavior change as it no longer matches the name of the method, `addNewVersions`.

dsanders11 (changes requested):
Let's get some test coverage on this before we merge.

There's also a couple of corner cases we should think about more carefully:
* We should probably handle the case where one window updates the Electron versions (can happen manually under settings, or by loading a gist with a new version specified)
  * Maybe we make the broadcast message more generic, make it `syncVersions` (plural) - takes an array
  * Make a private `AppState.addNewVersionsImpl` and move the existing code there
  * Then broadcast the message in `AppState.addNewVersions` with the full list of versions - the broadcast message listener would use the private `AppState.addNewVersionsImpl` to avoid triggering a new broadcast
* What happens to the UI if one window removes a version another window is using? Might need to make a separate issue for this, we should likely prevent removing a version which any window is using - although figuring out the best way to share that state might be tricky (may need to use the main process to coordinate that)

src/renderer/state.ts
erikian: A couple of alternatives I can think of:

- instead of calling `addNewVersions` in the `message` listener, it would be simple enough to just update the version in `versions` directly:
```ts
this.versions[payload.version] = payload;
```

- rename `addNewVersions` to something like `addOrUpdateVersions` and, if you're also concerned about always replacing the current version vs. only ever adding values that don't exist yet, we could update the existing version only if some of the properties we try to update in this PR is different:
```ts
public addOrUpdateVersions(versions: RunnableVersion[]) {
  for (const ver of versions) {
    const currentVersion = this.versions[ver.version];

    if (
      !currentVersion ||
      currentVersion.downloadProgress !== ver.downloadProgress ||
      currentVersion.state !== ver.state
    ) {
      this.versions[ver.version] = ver;
    }
  }
}
```

I think the first one is the least disrupting and it definitely makes for a shorter diff, but I'm open to suggestions.

dsanders11: Per my other comment about making the broadcast message more generic and taking an array of versions, how about we add a new method, `setVersionStates(versions: RunnableVersion[])` which contains the original change here the comment is on - just a for loop and sets the state.

erikian: Just a sanity check to see if I'm still following:

> add a new method, setVersionStates(versions: RunnableVersion[]) which contains the original change here the comment is on - just a for loop and sets the state.

If I understand correctly, we'd replace the `updateVersionState` method we currently have with:
```ts
private setVersionStates(versions: RunnableVersion[]) {
  for (const ver of versions) {
    this.versions[ver.version] = ver;
  }
}
```

> Make a private AppState.addNewVersionsImpl and move the existing code there

That would be just another name for the `setVersionStates` method I just mentioned above, right?

> Then broadcast the message in AppState.addNewVersions with the full list of versions
```ts
public addNewVersions(versions: RunnableVersion[]) {
  for (const ver of versions) {
    this.versions[ver.version] ||= ver;
  }

  this.broadcastChannel.postMessage({
    type: AppStateBroadcastMessageType.syncVersions,
    payload: versions,
  });
}

// broadcast listener
switch (type) {
  case AppStateBroadcastMessageType.syncVersions: {
    this.setVersionStates(payload);

    break;
  }

  // ...
}
```

If this snippet is right, it would involve replacing our current `this.updateVersionState(ver)` calls with `this.addNewVersions([ver])` to update the version state across windows, since the `addNewVersions` method is the only place where we'd actually be broadcasting the message - and it feels odd to call `addNewVersions` when updating the download progress, for example.

dsanders11: You're right, things have shifted a little since my original comment. This is the overall sketch I have in my head now, with a slight rename to avoid confusion:
* Rename the existing (in this PR) `updateVersionState` method to `broadcastVersionStates(versions: RunnableVersion[])`
* `setVersionStates(versions: RunnableVersion[])` is a new method all together
* `addNewVersions` would also use `broadcastVersionStates`
* The broadcast listener uses `setVersionStates` as you have above

I *think* that covers all the states.

erikian: Thank you, that's clear now and it works fine. Before pushing that change, I'd like to take a step back to your original comment:

> We should probably handle the case where one window updates the Electron versions (can happen manually under settings, or by loading a gist with a new version specified)

If I understand the requirements correctly, we can also do that by simply rolling back `addNewVersions` to what it was before this PR (`this.versions[ver.version] ||= ver`) and using `this.versions[payload.version] = payload` in the broadcast listener as I suggested in this thread. As you can see in the video below, if the left window downloads v25.0.0, the right window will be updated as expected:

https://github.com/electron/fiddle/assets/19478575/7cb912aa-06e6-40a0-a1ab-e29eafe6e4cc

I built Fiddle locally and tried to open a fiddle from the docs manually from the command line, and AFAICT, it works as expected — when the fiddle is loaded in one window, the other window will reflect any changes to the version the fiddle uses:

https://github.com/electron/fiddle/assets/19478575/bad946e2-0e8c-45ae-ae17-5a7cc711359c

I'm double-checking because that seems a lot simpler, but I'm happy to toss this approach if there are any downsides or corner cases I'm missing.

(As an aside, notice in the second video that the version selector in the second window highligths v25.2.0 when I load the fiddle in the first window. The component probably uses the version in localStorage to determine which version is active, and changing that would involve some more thinking, probably worth another issue).

erikian: I removed the `releases.json` file from my AppData folder (and also had to remove one of the versions from the `./static/releases.json` project file — if the file is missing or is an empty array, Fiddle won't start correctly), and yeah my approach fails miserably in that case 😅.

The approach you suggested does work. At first, I tried to only leave one version in my `./static/releases.json` file, but it took ~33 _seconds_ for the other window to update its state after receiving the broadcast with ~1900 versions and start responding to user interaction again — this scenario sounds very unlikely, so we should be safe when updating even a few dozen versions.

tests/setup.js
```diff
@@ -32,6 +32,28 @@ jest.mock('@sentry/electron/renderer', () => ({
...
+  };
+}
+
+global.BroadcastChannel = class Singleton {
+  static instance = new FakeBroadcastChannel();
+
+  constructor(_) {
+    return Singleton.instance;
+  }
+};
```
dsanders11, line 55:
```suggestion
class FakeBroadcastChannel extends EventTarget {
  constructor(channelName) {
    super();
    this.name = channelName;
  }

  postMessage(message) {
    this.dispatchEvent(new MessageEvent('message', { data: message }));
  }
}

global.BroadcastChannel = class Singleton {
  static channels = new Map();

  constructor(channelName) {
    if (!Singleton.channels.has(channelName)) {
      Singleton.channels.set(
        channelName,
        new FakeBroadcastChannel(channelName),
      );
    }
    return Singleton.channels.get(channelName);
  }
};
```

Can extend from `EventTarget` here to handle most of the actual implementation, and flesh out `global.BroadcastChannel` to act more like the spec `BroadcastChannel`.

dsanders11:
I extended test coverage to test the broadcasting side as well (not just the receiving) and also made `AppState.broadcastChannel` private, as I think that makes the most sense. You can pick up the changes from this commit if they look good to you: https://github.com/dsanders11/fiddle/commit/21ac67af7bf3a6c6ec252b387a26f5a7b76823da

erikian:
Thanks for the test case, looks solid!

src/renderer/state.ts
```diff
@@ -435,6 +455,31 @@ export class AppState {
...
+     * Handles communications between windows.
+     */
+    this.broadcastChannel.addEventListener(
+      'message',
+      (event: MessageEvent<AppStateBroadcastMessage>) => {
+        const { type, payload } = event.data;
+
+        switch (type) {
+          case AppStateBroadcastMessageType.syncVersions: {
+            this.setVersionStates(payload);
```
dsanders11, line 468:
```suggestion
            runInAction(() => this.setVersionStates(payload));
```

src/renderer/state.ts
dsanders11: > The approach you suggested does work. At first, I tried to only leave one version in my ./static/releases.json file, but it took ~33 seconds for the other window to update its state after receiving the broadcast with ~1900 versions and start responding to user interaction again — this scenario sounds very unlikely, so we should be safe when updating even a few dozen versions.

It looks like that's because `setVersionStates` is triggering mobx to cause updates on every version change. We can fix this by wrapping it with `runInAction` so that it's a single update instead of thousands. Tested locally and it resolves this issue - see my suggested change.

erikian: I went ahead and marked `setVersionStates` as an `action` in the `makeObservable` call in the constructor, so we don't end up shooting ourselves in the foot by forgetting to wrap the calls to that method with `runInAction` 😅

dsanders11 (approved):

dsanders11:
@erikian, since #1404 landed, I've pushed c79fdf94 to also broadcast `isDownloadingAll` so that windows stay in sync with that as well - otherwise one window would see versions downloading without the the downloading all state reflecting it.

If c79fdf94 looks good to you we can go ahead and merge this. 👍

erikian:
LGTM!

PR #1395, opened 2023-07-10
feat: prevent Electron versions that are in use by some window from being removed

### LINKED ISSUE #1394: Prevent Removing Other Window Current Versions
Currently the active version is prevented from being removed in Settings -> Electron, but if multiple windows are open this doesn't prevent one window from removing another window's active version, causing problems.

An elegant solution here could be [Web locks](https://w3c.github.io/web-locks/) - create a shared lock (`version:<version>`) on the version that a window is currently using, and release it when the current version is changed. Then modify the code that prevents removing an active version to check if a version is currently locked (`navigator.locks.query` wrapped inside a helper function). Potential problem with this approach is `navigator.locks.query` is async, and the current UI doesn't handle async state. Looks like we could probably use `react-async` for the moment (with an upgrade from React v16.14 to v16.18) and then use Suspense in the future when React is upgraded to v18+.

erikian:
Some initial work on https://github.com/electron/fiddle/issues/1394:

https://github.com/electron/fiddle/assets/19478575/2290addc-79ab-4cfc-8210-b7b1805c6c21

A couple call-outs:
- this needs https://github.com/electron/fiddle/pull/1389 to land first, as the current Typescript version doesn't know about `navigator.locks`
- From the issue description:
> Potential problem with this approach is navigator.locks.query is async, and the current UI doesn't handle async state. Looks like we could probably use react-async for the moment (with an upgrade from React v16.14 to v16.18) and then use Suspense in the future when React is upgraded to v18+.

@dsanders11 I used plain lifecycle methods to update the UI when a version is set as active in another window, so we might be able to do without another dependency here if that aligns with what you had in mind.

- upgrading to React 18 would be nice (though not necessary), there's an annoying "error" in development mode when the `ElectronSettings` is unmounted because of the async state update (there's a comment in the code with more details).

Fixes #1394.


src/renderer/state.ts
```diff
@@ -208,6 +208,34 @@ export class AppState {
...
+  // Lock on the active Electron version that prevents other windows from removing it
+  private versionLock: Lock | null = null;
+
+  // Used to release the lock when the current window switches Electron versions
+  private versionLockController = new AbortController();
+
+  private static versionLockNamePrefix = 'version:';
+
+  public getVersionLockName(ver: string) {
+    return [AppState.versionLockNamePrefix, ver].join('');
```
dsanders11, line 220:
```suggestion
    return `${AppState.versionLockNamePrefix}${ver}`;
```

Seems more intuitive, IMO.

src/renderer/state.ts
```diff
@@ -208,6 +208,34 @@ export class AppState {
     versions: this.baseVersions,
   });
 
+  // Lock on the active Electron version that prevents other windows from removing it
```
dsanders11, line 211: Maybe add a note that this is stored but not used, it only needs to be stored to prevent it from being garbage collected.

src/renderer/state.ts
```diff
@@ -208,6 +208,34 @@ export class AppState {
...
+  private versionLock: Lock | null = null;
+
+  // Used to release the lock when the current window switches Electron versions
+  private versionLockController = new AbortController();
+
+  private static versionLockNamePrefix = 'version:';
```
dsanders11, line 217: Part of me thinks this should go in `src/renderer/constants.ts`, but I could go either way, I don't feel strongly.

src/renderer/components/settings-electron.tsx
```diff
@@ -390,7 +428,7 @@ export const ElectronSettings = observer(
           break;
       }
 
-      if (version === appState.currentElectronVersion.version) {
+      if (this.state.activeVersions.has(appState.getVersionLockName(version))) {
```
dsanders11, line 431: Can this implementation be refactored so that `AppState.getVersionLockName` isn't a public method? Feels like it's exposing too much internal implementation details of how active versions are tracked - more ideally it should just track the version numbers, not the lock name. I think maybe changing the implementation of `getActiveVersions` to return just the versions, not the lock names, would accomplish that.

src/renderer/components/settings-electron.tsx
```diff
@@ -60,11 +61,47 @@ export const ElectronSettings = observer(
...
+    };
+
+    public componentDidMount() {
+      this.updateActiveVersions();
+    }
+
+    // Fired when other windows change their active Electron version
+    public componentDidUpdate() {
+      this.updateActiveVersions();
+    }
```
dsanders11, line 103: I'm not quite understanding the mechanism by which this works - what causes another window changing its active Electron version to trigger `componentDidUpdate` here? I can see it's working based on your video, I just don't understand it. 🤔

dsanders11: @erikian, poked around with it locally and figured it out. Since `updateActiveVersions` unconditionally modifies the state, which triggers a component update, this is effectively creating an infinite loop on the component updating, which is why it appears to be working as intended.

dsanders11: What if we extended the broadcast mechanism in #1391 to add a message (`activeVersionsChanged`, for example), move `activeVersions` to `AppState`, and then in response to that message we update `AppState.activeVersions`? The broadcast message would have no payload, it would just cause `AppState` to update `activeVersions`, which would then cause an update to the `ElectronSettings` component.

src/renderer/state.ts
```diff
@@ -818,10 +848,36 @@ export class AppState {
...
+
+      // replace the spent AbortController
+      this.versionLockController = new AbortController();
+    }
+
     const { version } = ver;
     console.log(`State: Switching to Electron ${version}`);
     this.version = version;
 
+    navigator.locks.request(
```
dsanders11, line 865:
```suggestion
    await navigator.locks.request(
```

erikian: That would hang the execution until the lock is released, resulting in a blank screen on startup. For our purposes, we don't actually need the lock to be released to keep going — we just want it to be there.

src/renderer/state.ts
erikian: TBH I wasn't thinking about GC here, I'm just using it to determine whether there's an active lock. Might as well use a boolean flag instead 🤔

src/renderer/components/settings-electron.tsx
erikian: > Since updateActiveVersions unconditionally modifies the state, which triggers a component update, this is effectively creating an infinite loop on the component updating, which is why it appears to be working as intended.

My bad big time 🤦‍♂️ I thought this component was re-rendering because of the `storage` event that is fired when the other window selects a version.

> What if we extended the broadcast mechanism in https://github.com/electron/fiddle/pull/1391 to add a message (activeVersionsChanged, for example), move activeVersions to AppState, and then in response to that message we update AppState.activeVersions? The broadcast message would have no payload, it would just cause AppState to update activeVersions, which would then cause an update to the ElectronSettings component.

Sounds good to me.

dsanders11:
I've been pondering it some more, and I think we should expand the usage of the locking as a general solution. There's two main things which need to be solved regarding versions and multiple windows:
1. Prevent clashing actions (removing while in use, simultaneous downloads, etc)
2. Update the UI to reflect which versions can be removed

The current state of this PR is solving 2, but I think we can solve 1 at the same time. Here's what I've been pondering:
* For removing, try to acquire an exclusive lock on the `version:<version>` lock, but don't wait on the lock, use the `ifAvailable` option to fail fast
  * This is meant to handle the cases where one window is racing with the other: user clicks the remove button before the window updates the UI, or one window is doing "Remove All Downloads" and another window is changing active versions
* When downloading, first get the shared `version:<version>` lock, and then also try to get an exclusive lock on `downloading:<version>`, waiting on the lock. That way we don't end up with simultaneous downloads, which can currently happen if you switch to an undownloaded version in a window and then open a new window.

erikian:
I did what you suggested and came up with a way to simulate a user trying to do something simultaneously in multiple windows. It's a bit convoluted but the results seem to be reliable, so here it goes:

For the `user clicks the remove button before the window updates the UI` case, I did this (video below):

1. Open two Fiddle windows and their devtools.
2. In the first window, open Settings > Electron and create a BroadcastChannel in the console that will try to click on the "Remove" version for v25.3.0:
```ts
const receiverBroadcastChannel = new BroadcastChannel('testLocks');

receiverBroadcastChannel.addEventListener('message', async () => {
  // the version is actually removed if we don't wait for a bit, but I assume
  // users doing it manually would be at least this slow anyway :)  
  await new Promise(resolve => setTimeout(resolve, 20))

  console.log('trying to remove the version the other window just selected');
  const v25_3_0_removeButton = [
    ...document.querySelectorAll('table .bp3-button'),
  ][1];
    
  v25_3_0_removeButton.click();
});
```
3. In the second window, open the version selector, select the v25.3.0 button in the Elements tab in the devtools, and programatically click on it and create another BroadcastChannel that will post a message to the other window:
```ts
// $0 is the button that selects v25.3.0
$0.click();

const emitterBroadcastChannel = new BroadcastChannel('testLocks');

emitterBroadcastChannel.postMessage('whatever');
```

The first window will simulate the user clicking on the "Remove" button for 25.3.0 just after that version is selected in the second window, and nothing happens:

https://github.com/electron/fiddle/assets/19478575/f4db2aa6-b95e-4a8c-b08e-eacab8d0583f

-------

For the simultaneous downloads case (video below): 

1. Open two Fiddle windows and their devtools.
2. Open the version selector in both windows.
3. In both devtools, create a BroadcastChannel in the console that will trigger the click on the first version that's not been downloaded - should be the same version in both windows:
```ts
const receiverBroadcastChannel = new BroadcastChannel('testLocks');

receiverBroadcastChannel.addEventListener('message', () => {
  console.log('trying to download...');
  document.querySelector('a[label="Not downloaded"]').click();
});
```
4. In one of the devtools, create another BroadcastChannel that will post a message to the receiving BroadcastChannels in both windows:
```ts
const emitterBroadcastChannel = new BroadcastChannel('testLocks');

emitterBroadcastChannel.postMessage('whatever');
```

Both windows will receive the broadcast at the same time. Only one of the windows will be granted the lock at first (the window on the right side in the video), and the other window will only get the lock when the download finishes in the first window.

Even though the lock does its job, we still get an error and end up with an inconsistent state ("Checking status" in the download selector in the window that first got the lock, and a "dest already exists" error in the other window) when both windows try to download the same version in rapid succession (the same thing happens in the case you mentioned of the user opening a new window when a download is in progress), so maybe `fiddle-core` is not handling this properly based on the error stack:

https://github.com/electron/fiddle/assets/19478575/d7a4ba6a-00ea-4d16-88e4-6569671769d6

I think this is mostly ready for review, but I'd like some feedback on this before I add some new tests 🙂

PR #1398, opened 2023-07-14
refactor: use @observer class decorator

dsanders11:
Refactoring to use `@observer` instead of `observer(...)` makes typings work properly with `shallow<T>(...)`. I went through and cleaned up all the usage of `any` in the tests to use proper typings, and fixed a few things which fell out as a result.

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -41,430 +41,427 @@ interface IGistActionButtonState {
  * @class GistActionButton
  * @extends {React.Component<GistActionButtonProps, GistActionButtonState>}
  */
-export const GistActionButton = observer(
-  class GistActionButton extends React.Component<
-    GistActionButtonProps,
-    IGistActionButtonState
-  > {
-    public constructor(props: GistActionButtonProps) {
-      super(props);
-      this.handleClick = this.handleClick.bind(this);
-      this.performGistAction = this.performGistAction.bind(this);
-      this.setPrivate = this.setPrivate.bind(this);
-      this.setPublic = this.setPublic.bind(this);
+@observer
+export class GistActionButton extends React.Component<
+  GistActionButtonProps,
+  IGistActionButtonState
+> {
+  public constructor(props: GistActionButtonProps) {
+    super(props);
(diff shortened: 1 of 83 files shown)


BlackHole1 (commented):
I don't know if this is a good change because `decorators` are not currently recommended in `mobx`, and support for `decorators` has been discontinued since `mobx@6`. If we use `decorators`, it may have a negative impact on the future upgrade of `mobx`. See: https://mobx.js.org/enabling-decorators.html

dsanders11:
Thanks for the link @BlackHole1, I hadn't seen that.

Personally, I think the resulting proper typing in the tests outweighs any drawbacks here, as not having proper typing was letting bugs and other issues slip in. Since the syntax changes in `src` are small, we can always revert them in the future if they do cause a problem for upgrading packages. But, I think we have larger upgrade hurdles in the future, since [React considers the class based `React.Component` to be legacy](https://react.dev/reference/react/Component) and recommends against using it. So `@observer` may get dropped naturally in future upgrades to newer React if we modernize the codebase.

While MobX does not recommend the decorators, I don't think it will stop working any time soon, since they're supported in TypeScript.

> We do not recommend new codebases that use MobX use decorators until the point when they become an official part of the language, but you can still use them.

erickzhao (changes requested):
I agree with @BlackHole1 here. A huge part of the MobX 5 -> 6 upgrade was to remove the `@observer` decorators we had already due to the recommendations from the MobX team (see https://github.com/electron/fiddle/pull/1061).

I feel like this is a step backwards to solve a typing problem with Enzyme, which we want to eventually drop in favour of React Testing Library anyways (added in #1093).

I don't think the RTL tests require us to type stuff as `any`.

dsanders11:
> I feel like this is a step backwards to solve a typing problem with Enzyme, which we want to eventually drop in favour of React Testing Library anyways (added in https://github.com/electron/fiddle/pull/1093).
>
> I don't think the RTL tests require us to type stuff as any, so I think it would be better to fix the tests directly by migrating them over.

There should be an open issue to track that - I was unaware of the difference, or that there is a planned migration.

I'll close this and move some of the improvements to a different PR.

PR #1404, opened 2023-07-16
feat: cancel download all versions option

### LINKED ISSUE #1402: Add Cancel Download All Versions
This is similar to #1016 but should be easier to implement since we don't need `@electron/get` support for canceling an individual download. Currently if you click "Download All Versions" there's no way to cancel that, which is pretty bad since that can be hundreds of downloads. The only way to cancel is to quit the app.

This shouldn't be too difficult to fix, in the for loop which downloads all versions we should just check if `isDownloadingAll` is still true after each download and `break` if it's not. The other change would be when `isDownloadingAll` is true, render a "Stop Downloads" button in place of the "Download All Versions" button, and clicking that will set `isDownloadingAll` to false.

It will need to move `isDownloadingAll` to `AppState` as well (`isDeletingAll` too) since downloads continue after the user leaves settings, and it would probably be confusing if switching settings tabs or leaving settings stopped the downloads. Currently leaving the settings tab means the `isDownloadingAll` state is lost and the button could be clicked again or "Remove All Versions" could be clicked, so moving the state to `AppState` will fix that.

https://github.com/electron/fiddle/blob/11b54b90ce11082b031be85223340bb23f3b18d3/src/renderer/components/settings-electron.tsx#L115-L125

arnabrahman:
## Description of your changes
Currently, if we click "Download All Versions" there's no way to cancel that, which is pretty bad since that can be hundreds of downloads. The only way to cancel is to quit the app.

 Introduced a `Stop Downloads` button in place of the `Download All Versions` button during download. Moved `isDownloadingAll` & `isDeletingAll` inside `AppState`. Did all of these according to the #1402 description. Some tests were failing because of these updates. Fixed the tests & also added new tests to cater to the changes.
 
 Please let me know if I need any updates to the code. 

### Before (When clicked `Download All Versions`)
before

### After (When clicked `Download All Versions`)
after

### Related issues

Closes #1402

### DIFF
--- a/src/renderer/components/settings-electron.tsx
+++ b/src/renderer/components/settings-electron.tsx
@@ -30,11 +30,6 @@ interface ElectronSettingsProps {
   appState: AppState;
 }
 
-interface ElectronSettingsState {
-  isDownloadingAll: boolean;
-  isDeletingAll: boolean;
-}
-
 /**
  * Settings content to manage Electron-related preferences.
  *
@@ -42,10 +37,7 @@ interface ElectronSettingsState {
  * @extends {React.Component<ElectronSettingsProps, ElectronSettingsState>}
  */
 export const ElectronSettings = observer(
-  class ElectronSettings extends React.Component<
-    ElectronSettingsProps,
-    ElectronSettingsState
-  > {
+  class ElectronSettings extends React.Component<ElectronSettingsProps> {
     constructor(props: ElectronSettingsProps) {
       super(props);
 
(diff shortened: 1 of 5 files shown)


dsanders11 (changes requested):
Overall very clean PR, thanks! Just a couple of suggested changes.

arnabrahman:
@dsanders11 Thanks for the suggestions. I have added them.

dsanders11 (approved):

PR #1412, opened 2023-07-22
fix: added fiddle-scrollbar to resize titlebar responsively

### LINKED ISSUE #1274: Titlebar does not resize responsively enough when resized horizonally
When Fiddle's `BrowserWindow` width is shrunk past a certain point, items on the far right are cut off completely with no way to scroll to them:

Screenshot 2023-02-27 at 2 59 55 PM

We should either:

1. Allow users to scroll to the cut-off items somehow
2. Leverage media queries to restructure the title bar when it gets small enough

manishsalian09:
This fixes #1274
application resized horizontally to smaller screen size
#1274-fix-screenshot1

application resized horizontally to larger screen size
#1274-fix-screenshot2

### DIFF
--- a/src/renderer/components/commands.tsx
+++ b/src/renderer/components/commands.tsx
@@ -42,8 +42,8 @@ export const Commands = observer(
         <div
           className={
             window.ElectronFiddle.platform === 'darwin'
-              ? 'commands is-mac'
-              : 'commands'
+              ? 'commands is-mac fiddle-scrollbar'
+              : 'commands fiddle-scrollbar'
           }
           onDoubleClick={this.handleDoubleClick}
         >


codebytere (changes requested):
We appreciate this contribution but i'm not sure this is the right change here. Horizontally scrolling a titlebar is not native behavior, nor does it look or feel natural for the user. I think we want to do something more like changing the icon layout and removing text when the width is small enough to make this feel more responsive.

manishsalian09:
@codebytere Thank you for reviewing the changes. Apologies for committing the changes before getting more clarification. If i understand when the width is small we should try to display only logo and if possible text on some items it space permits?

PR #1419, opened 2023-07-26
ci: fix code signing on Windows

dsanders11:
There were a couple of issues preventing code signing on Windows from working properly, and it was silently failing. Fix those up and try to fail loudly if something isn't right.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,6 +100,8 @@ jobs:
           paths:
             - out
   win-build:
+    environment:
+      SIGN_OR_DIE: true
     parameters:
       arch:
         type: enum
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -100,6 +100,8 @@ jobs:
           paths:
             - out
   win-build:
+    environment:
+      SIGN_OR_DIE: true
```
VerteDinde, line 104: This is art

forge.config.ts
```diff
@@ -21,7 +21,15 @@ if (process.env['WINDOWS_CODESIGN_FILE']) {
...
     process.env['WINDOWS_CODESIGN_FILE'] = certPath;
+  } else if (process.env['SIGN_OR_DIE']) {
+    throw new Error('Did not find Windows codesign file');
   }
+
+  if (!process.env['WINDOWS_CODESIGN_PASSWORD'] && process.env['SIGN_OR_DIE']) {
+    throw new Error('Did not find "WINDOWS_CODESIGN_PASSWORD" env variable');
+  }
+} else if (process.env['SIGN_OR_DIE']) {
+  throw new Error('Did not find "WINDOWS_CODESIGN_FILE" env variable');
```
VerteDinde, line 32: Do we need this check if we're checking for the lack of WINDOWS_CODESIGN_PASSWORD in the line above? Could we just check for the presence of WINDOWS_CODESIGN_PASSWORD and WINDOWS_CODESIGN_FILE, and throw if either aren't present/are an empty string?

VerteDinde (approved):
One comment, not blocking

erickzhao (approved):

forge.config.ts
dsanders11: Tweaked the logic in 91e0fb84e3864dab47294ea77bf342adcb26902f, if that's good with you then merge away. :shipit:

PR #1422, opened 2023-07-26
ci: another Windows code signing fix

dsanders11:
The first attempt here in #1419 had some problems that this sorts out, namely:
* Environment variables on CircleCI changed names
* Wasn't properly getting and using the output of `tools/add-windows-cert.ts`

### DIFF
--- a/tools/add-windows-cert.ts
+++ b/tools/add-windows-cert.ts
@@ -7,7 +7,7 @@ export async function generateWindowsSigningCert() {
   try {
     const dir = await fs.mkdtemp(path.resolve(os.tmpdir(), 'builder-folder-'));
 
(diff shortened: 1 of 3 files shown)


tools/add-windows-cert.ts
```diff
@@ -28,6 +28,6 @@ export async function generateWindowsSigningCert() {
 
 if (require.main === module) {
   (async () => {
-    await generateWindowsSigningCert();
```
erickzhao, line 31: What does this log line do?

dsanders11: `tools/add-windows-cert.ts` is supposed to write the certificate to disk and then output the path, so the log is outputting it from the script.

erickzhao (approved):

PR #1428, opened 2023-08-02
refactor: move getting Node.js types to main process

dsanders11:
First PR to shift the file operations around Node.js types to the main process.

### DIFF
--- a/src/main/electron-types.ts
+++ b/src/main/electron-types.ts
@@ -1,0 +1,142 @@
+import * as path from 'node:path';
+
+import { ElectronVersions } from '@electron/fiddle-core';
+import fetch from 'cross-fetch';
+import { IpcMainEvent, app } from 'electron';
+import * as fs from 'fs-extra';
+import packageJson from 'package-json';
+import readdir from 'recursive-readdir';
+import semver from 'semver';
+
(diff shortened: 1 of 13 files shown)


src/main/electron-types.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    try {
+      const files = (await readdir(dir)).filter((f) => f.endsWith('.d.ts'));
+
+      for (const file of files) {
+        types[path.relative(dir, file) as keyof NodeTypes] = fs.readFileSync(
+          file,
+          'utf8',
+        );
+      }
```
erikian, line 55: nit: no big deal, but this could be async - always a good idea to avoid blocking I/O if we can help it

```suggestion
      for (const file of files) {
        types[path.relative(dir, file) as keyof NodeTypes] = await fs.readFile(
          file,
          'utf8',
        );
      }
```

erikian (approved):

src/main/electron-types.ts
dsanders11: Yep, seems reasonable. When doing these refactors I try to change as little as possible to minimize risk of introducing a new bug, but this is low risk.

Issue #1408, opened 2023-07-18
Migrate tests from Enzyme to React Testing Library


erickzhao:
Since #1093, we have the ability to test Fiddle's React components using React Testing Library (RTL).

We should gradually migrate all our Enzyme component tests to RTL to move away from Enzyme for the reasons detailed in the [Fiddle Test Improvements](https://docs.google.com/document/d/12aYOCxEXJ-NkAgU0odw3Q9rQBn-Rdg1ODYyqOnq8G14/edit#heading=h.6eui53oy77m4) document.

PR #1406, opened 2023-07-18
build: convert Forge and Webpack config to TypeScript

dsanders11:
TypeScript all the things.

PR #1432, opened 2023-08-04
test: update test dependencies and rearrange files

erikian:
Just some groundwork related to #1408:

- `jest.json` has been moved to `jest.config.js` - that's the default location and makes the test scripts config a little simpler
- updated some Jest- and RTL-related dependencies (including `@testing-library/dom` from v8 to v9, with no breaking changes that affect us)
- moved the RTL tests from `rtl-spec` to `rtl-spec/components` so we can mirror the current structure


dsanders11:
> `jest.json` has been moved to `jest.config.js` - that's the default location and makes the test scripts config a little simpler

Let's make this `jest.config.ts` and get type checking support for the config. [The Jest docs have an example using `jest.config.ts`](https://jestjs.io/docs/configuration), and testing it real quick shows it works well without any extra changes - it requires `ts-node` but I already added that in #1406 when moving other configs to TypeScript.

dsanders11:
Subjective nit: I'd call this a `test:` or `chore:` rather than `refactor:` since it's not changing code. My personal general rule of thumb is to use `test:` for changes like this which only touch testing stuff so that I can easily skip over them if tracking down which commit introduced a bug, while `refactor:` I use for changes which refactor code but shouldn't have user-visible changes.

Minor nit: let's shorten up the title if we can, it's nice when the title is < 72 chars so that the whole message can be seen when scrolling through commit history.

erikian:
> Subjective nit: I'd call this a test: or chore: rather than refactor: since it's not changing code

Yeah, agreed on using `test:` here 🙂

> Minor nit: let's shorten up the title if we can

Done 👍

erickzhao (approved):

PR #1434, opened 2023-08-05
test: update `testMatch` for RTL tests

erikian:
I moved RTL tests to a subfolder in #1432 and failed to update the regex to actually run these tests.

Should we have a coverage report in-PR to make it easier to spot any coverage drops?

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -8,7 +8,7 @@ const config: Config = {
   testEnvironmentOptions: {
     url: 'http://localhost',
   },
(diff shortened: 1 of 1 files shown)


dsanders11:
> Should we have a coverage report in-PR to make it easier to spot any coverage drops?

We had Coveralls in the past, but it didn't get re-enabled after the switch to CircleCI. I had #1301 but it was blocked on getting a token from @electron/wg-infra. I'll see if we can make that happen, you're right, it would have caught this case.

dsanders11 (approved):

PR #1435, opened 2023-08-06
test: use RTL in the version selector tests

erikian:
Another step towards #1408. I've added a couple of changes on top of just migrating the tests to RTL (open to feedback if any of them are a bad idea):

- extracted some duplicate version mocking logic to a separate file
- created a helper function to render the `VersionSelect` component - I didn't do it in a `beforeEach` block because we only need to render the whole component in one test
- added a missing test for the `renderVersionContextMenu` component

~~I've also included the `testMatch` change from #1434 just so the RTL tests will run on CI for this PR. This means that #1434 should land first to keep things separate and to prevent it from becoming an empty changeset.~~


tests/renderer/components/__snapshots__/commands-version-chooser-spec.tsx.snap
```diff
@@ -1,45 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
VerteDinde, line 1: This is me being ignorant of RTL - do we need to generate new snapshots for RTL to test against? I think RTL still uses snapshot testing, unless I'm mistaken here? Reading through: https://testing-library.com/docs/react-testing-library/migrate-from-enzyme

erikian: We could generate new snapshots, but from what I gather from the RFC @erickzhao linked in https://github.com/electron/fiddle/issues/1408, and I might be wrong here, we probably shouldn't:

> Most PRs to the React components end up regenerating the snapshot. We don’t actually pay attention to the diff too much during development, so these changes end up being low signal and high noise.

erickzhao (approved):

ckerr:
Looks like this was approved a a couple of months ago but not merged.

@erikian is this PR still something that should be merged; and if so, could you update the PR to resolve the conflicts that have built up in the interim? :smile_cat:

erikian:
@ckerr I'll try to get this PR rebased later this week, thanks for the reminder!

PR #1436, opened 2023-08-08
ci: don't use pull_request_target with semantic PR title check

dsanders11:
Refs https://github.com/electron/electron/pull/37745.


ckerr (approved):
:+1:

PR #1439, opened 2023-08-10
fix: error loading big gists

### LINKED ISSUE #1438: Gist loading will fail if the gist gets truncated for size
See https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#truncation

We don't handle the case where `octokit.gists.get` returns a file that's been truncated, and so Fiddle errors on invalid JSON input:

Screenshot 2023-08-10 at 6 17 26 PM

To reproduce, paste https://gist.github.com/codebytere/e06a6d4e650ee581f90d0159b701aaa2 into the gist loading section and try to load it. 

If a given file returned by `octokit.gists.get` has `truncated: true`, we should fetch the rest of the contents by making GET request to the URL specified by `raw_url` for that file.

Relevant code is [here](https://github.com/electron/fiddle/blob/295f74f50ce7d44315d53c31ac470d29a2cbfa23/src/renderer/remote-loader.ts#L121-L196)

hvohr:
Closes #1438 

Fixes an issue where loading gists over 1mb would throw an error due to the file being truncated. Per [Gist Docs](https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#truncation), when the file is truncated, the full contents can be obtained by fetching the raw url. This adds error handling for that issue.

Tested w/ https://gist.github.com/codebytere/e06a6d4e650ee581f90d0159b701aaa2

### DIFF
--- a/src/renderer/remote-loader.ts
+++ b/src/renderer/remote-loader.ts
@@ -125,8 +125,11 @@ export class RemoteLoader {
       const values: EditorValues = {};
 
       for (const [id, data] of Object.entries(gist.data.files)) {
+        const content = data.truncated
+          ? await fetch(data.raw_url).then((r) => r.text())
+          : data.content;
         if (id === PACKAGE_NAME) {
-          const { dependencies, devDependencies } = JSON.parse(data.content);
+          const { dependencies, devDependencies } = JSON.parse(content);
           const deps: Record<string, string> = {
             ...dependencies,
             ...devDependencies,
@@ -177,7 +180,7 @@ export class RemoteLoader {
         if (!isSupportedFile(id)) continue;
 
         if (isKnownFile(id) || (await this.confirmAddFile(id))) {
-          values[id] = data.content;
+          values[id] = content;
         }
       }
 
--- a/tests/renderer/remote-loader-spec.ts
+++ b/tests/renderer/remote-loader-spec.ts
@@ -167,25 +167,28 @@ describe('RemoteLoader', () => {
 
     it('handles gists with files over 1mb', async () => {
       const gistId = 'toobig';
+      const filename = 'index.js';
+      const content = 'hello im huge';
 
-      mockGistFiles['index.js'] = {
+      editorValues[filename] = content;
+      mockGistFiles[filename] = {
         truncated: true,
-        content: JSON.stringify('hello im huge', null, 2),
+        content: 'truncated',
         raw_url: 'https://gist.githubusercontent.com/IMTOOBIG',
       };
 
-      jest.spyOn(global, 'fetch').mockImplementationOnce(() =>
-        Promise.resolve({
-          text: () => Promise.resolve('hello im huge'),
-        } as Response),
-      );
+      jest.spyOn(global, 'fetch').mockResolvedValueOnce({
+        text: () => Promise.resolve(content),
+      } as Response);
 
       mocked(getOctokit).mockResolvedValue({
         gists: mockGetGists,
       } as unknown as Octokit);
+      instance.confirmAddFile = jest.fn().mockResolvedValue(true);
 
       const result = await instance.fetchGistAndLoad(gistId);
       expect(result).toBe(true);
+      expect(app.replaceFiddle).toBeCalledWith(editorValues, { gistId });
     });
 
     it('does not set an invalid Electron version from package.json', async () => {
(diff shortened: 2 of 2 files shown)


dsanders11 (commented):
Change looks good. 👍 Could you add a test to this section? 

https://github.com/electron/fiddle/blob/47f62e36767f43846645b7bee9eab7e70d0aa55e/tests/renderer/remote-loader-spec.ts#L72

src/renderer/remote-loader.ts
```diff
@@ -125,8 +125,11 @@ export class RemoteLoader {
       const values: EditorValues = {};
 
       for (const [id, data] of Object.entries(gist.data.files)) {
+        const content = data.truncated
+          ? await fetch(data.raw_url).then((r) => r.text())
```
BlackHole1, line 129: > From gist docs:
> If you need the full contents of the file, you can make a GET request to the URL specified by raw_url. Be aware that for files larger than ten megabytes, you'll need to clone the gist via the URL provided by git_pull_url.

Would it be better to use git_pul_url? 🤔

dsanders11: Wouldn't that require cloning the repo, which would add significant complexity? It's a good point though, would be good to check for that case and throw an error saying the gist is too large to load. I think it's reasonable for Fiddle to not handle that specific case since it's unlikely to ever come up. 🙂

BlackHole1: Oh, sorry. I thought `git_pull_url` is equivalent to `raw_url. 😅

hvohr:
> Thank you for the feedback! I can absolutely add a test as soon as I am able!

hvohr:
> Change looks good. 👍 Could you add a test to this section?
> 
> https://github.com/electron/fiddle/blob/47f62e36767f43846645b7bee9eab7e70d0aa55e/tests/renderer/remote-loader-spec.ts#L72


Thank you for the feedback! I can absolutely add a test as soon as I am able!

codebytere (approved):
🥳

tests/renderer/remote-loader-spec.ts
```diff
@@ -165,6 +165,29 @@ describe('RemoteLoader', () => {
...
+        } as Response),
+      );
+
+      mocked(getOctokit).mockResolvedValue({
+        gists: mockGetGists,
+      } as unknown as Octokit);
+
+      const result = await instance.fetchGistAndLoad(gistId);
+      expect(result).toBe(true);
+    });
```
dsanders11, line 189: Some tweaks are needed for the test, unfortunately GitHub doesn't show a pretty diff here.

The test before would also pass if you reverted the changes in `src/renderer/remote-loader.ts`. The suggested tweaks:
* Ensure the truncated content differs from the response to `raw_url` so we know it got the correct content
* Makes sure `instance.confirmAddFile` is mocked to true, since the filename is not one of the standard files
* Adds the expected contents to `editorValues` and adds an `expect(app.replaceFiddle)` at the end

```suggestion
    it('handles gists with files over 1mb', async () => {
      const gistId = 'toobig';
      const filename = 'index.js';
      const content = 'hello im huge';

      editorValues[filename] = content;
      mockGistFiles[filename] = {
        truncated: true,
        content: 'truncated',
        raw_url: 'https://gist.githubusercontent.com/IMTOOBIG',
      };

      jest.spyOn(global, 'fetch').mockResolvedValueOnce({
        text: () => Promise.resolve(content),
      } as Response);

      mocked(getOctokit).mockResolvedValue({
        gists: mockGetGists,
      } as unknown as Octokit);
      instance.confirmAddFile = jest.fn().mockResolvedValue(true);

      const result = await instance.fetchGistAndLoad(gistId);
      expect(result).toBe(true);
      expect(app.replaceFiddle).toBeCalledWith(editorValues, { gistId });
    });
```

dsanders11 (changes requested):

tests/renderer/remote-loader-spec.ts
hvohr: > Some tweaks are needed for the test, unfortunately GitHub doesn't show a pretty diff here.
> 
> The test before would also pass if you reverted the changes in `src/renderer/remote-loader.ts`. The suggested tweaks:
> 
> * Ensure the truncated content differs from the response to `raw_url` so we know it got the correct content
> * Makes sure `instance.confirmAddFile` is mocked to true, since the filename is not one of the standard files
> * Adds the expected contents to `editorValues` and adds an `expect(app.replaceFiddle)` at the end

Applied changes, thank you!

dsanders11 (approved):
Thanks @hvohr!

PR #1440, opened 2023-08-11
test: use RTL in the welcome tour tests

erikian:
Adresses #1408. This should be easier to review one commit at a time, as Github doesn't seem to be picking up the move across folders.

I removed the window resizing tests from `tour.spec.tsx` as it's not possible to access the component state in RTL (it's an implementation detail) and I didn't figure out a way to test that using something a user could see and interact with.


BlackHole1 (approved):

rtl-spec/components/tour.spec.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react';
+
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { mocked } from 'jest-mock';
+
+import { Tour, TourScriptStep } from '../../src/renderer/components/tour';
+import { overrideRendererPlatform } from '../../tests/utils';
+
+describe('VersionChooser component', () => {
```
erickzhao, line 10:
```suggestion
describe('Tour component', () => {
```

erickzhao (approved):

dsanders11:
Merging as the Coveralls failure on the Linux build is fixed on `main`.

PR #1442, opened 2023-08-15
refactor: finish splitting ElectronTypes to main process

dsanders11:
Follow-up to #1428 which finishes moving our Electron types code to the main process.

The only big change as a part of this refactoring is to the file watcher for local builds types. Since it is in the main process now, it has been streamlined to have one watcher per local build, and it keeps track of which windows need to be notified of changes to that path.


src/main/electron-types.ts
```diff
@@ -2,23 +2,84 @@ import * as path from 'node:path';
...
+
+        // If no watcher for that path yet, create it
+        if (!this.watchers.has(dir)) {
+          const watcher = watch(file, () => {
+            // Watcher should notify all windows watching that path
+            const windows = this.getWindowsForLocalPath(dir);
+            for (const window of windows) {
+              ipcMainManager.send(
+                IpcEvents.ELECTRON_TYPES_CHANGED,
+                [fs.readFileSync(file, 'utf8'), version],
```
erikian, line 61: I think we already have the file contents from L48, so it could be reused here instead of re-reading the file

erikian (approved):
LGTM, one non-blocking suggestion

src/main/electron-types.ts
dsanders11: This line is inside the callback for the file watcher, so the file contents have changed and need to be read again. I can pull this out into a private function to make that more clear.

erikian: Ah, makes sense. I think it's fine as it is then. One question though, won't this throw if the file is deleted or does the callback only run when the file exists?

erikian: nevermind, I don't think it can be deleted 😅

dsanders11: It's a good point, though. Prior to refactoring the read was in a `try/catch` block, now it's not - I'll fix that.

https://github.com/electron/fiddle/blob/47f62e36767f43846645b7bee9eab7e70d0aa55e/src/renderer/electron-types.ts#L89-L97

dsanders11: Refactored in 636e784831e5c5fc20d6e3cb5b82d6dff7a92173 - also noticed that the file contents were being read in the for-loop unnecessarily so cleaned that up as well. Thanks for the comment! 🙂

ckerr (approved):

PR #1452, opened 2023-08-27
refactor: move fiddle-core usage to main process

dsanders11:
Stacked on top of #1451 so in draft mode until that lands.

Refactor to move usage of `@electron/fiddle-core` to the main process, thus completing the full refactoring and setting up a PR to disable Node integration and enable context isolation.


src/renderer/state.ts
```diff
@@ -813,8 +809,8 @@ export class AppState {
         state === InstallState.installed ||
         state == InstallState.downloaded
       ) {
-        await this.installer.remove(version);
-        if (this.installer.state(version) === InstallState.missing) {
+        await window.ElectronFiddle.removeVersion(version);
+        if (this.getVersionState(version) === InstallState.missing) {
```
erikian, line 813: Since `window.ElectronFiddle.removeVersion` uses `ipcMain.invoke`, can we make it return the version state so we don't need another IPC round-trip?

src/preload/preload.ts
```diff
@@ -138,6 +149,8 @@ export async function setupFiddleGlobal() {
     },
     getTestTemplate: () => ipcRenderer.invoke(IpcEvents.GET_TEST_TEMPLATE),
     getUsername: () => ipcRenderer.sendSync(IpcEvents.GET_USERNAME),
+    getVersionState: (version: string) =>
+      ipcRenderer.sendSync(IpcEvents.GET_VERSION_STATE, version),
```
erikian, line 153: Can this be asynchronous? I realize we have a bunch of synchronous calls here, but IMO we should avoid them in new code if it's not too much of a hassle, they can add up and come back to bite us 😬

dsanders11: This was made sync to match the existing code without needing to do more extensive refactoring. Particularly this is used in a constructor where we can't use `async` calls. It can probably be updated to work with asynchronously populating the version states, but that will require more extensive testing so I don't want to touch that in this PR.

https://github.com/electron/fiddle/blob/74bde543f3cbb2818e9b0ab020a0e880d455b949/src/renderer/state.ts#L355-L362

I agree with the overall sentiment though - once this all lands we should look at doing future refactors to change any usages of `ipcRenderer.sendSync` that we can.

src/renderer/state.ts
dsanders11: Good suggestion! You're right, this can be optimized.

erikian (approved):

PR #1453, opened 2023-08-29
refactor: enable context isolation

dsanders11:
Stacked on top of #1452 so in draft mode until that lands.

Only a couple of changes needed:
* Had to push destructuring `RunnableVersion` up to the callers of `window.ElectronFiddle` functions - these are proxies since they're observable by MobX and you can't clone proxies across the context isolation interface
* Had to move `app` and `monaco` off of `window.ElectronFiddle` as they're set during initialization inside the renderer process and `window.ElectronFiddle` has read-only properties once context isolation is enabled


erikian (approved):

felixrieseberg:
Thanks for your hard work!

VerteDinde (approved):

forge.config.ts
```diff
@@ -64,7 +64,7 @@ const config: ForgeConfig = {
         },
         mainConfig: mainConfig,
         renderer: {
-          nodeIntegration: true,
+          nodeIntegration: false,
```
erickzhao, line 67: We could just remove this option altogether since it's false by default

erickzhao (approved):

PR #1465, opened 2023-09-22
ci: load macOS release secrets from secret service

dsanders11:
Shifting more usage to the secret service.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ commands:
       - node/install-packages
       - run: yarn run contributors
       - run: yarn run electron-releases
-  load-gh-token:
+  load-release-secrets:
     steps:
       - run: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -175,6 +176,7 @@ workflows:
             parameters:
               arch: [ x64 ]
       - mac-build:
+          context: fiddle-release
```
VerteDinde, line 179: I think we'll eventually need this context in all build steps but approving since Mac is ready now

VerteDinde (approved):

.circleci/config.yml
dsanders11: Yea, once we move the remaining secrets. I'm contemplating making a context per-platform just to silo them a bit more and only providing them on more of a "need to know" basis.

PR #1472, opened 2023-09-28
fix: use process.env when running fiddle

### LINKED ISSUE #1470: Fiddle fails to launch Electron process on Linux
Fiddle 0.34.2 fails to launch Electron process on Linux. I've tested it on both Fedora (RPM package) and Ubuntu 22.04 (Debian package).

Fiddle's console output and the symbolicated crash are attached.

[console_output.txt](https://github.com/electron/fiddle/files/12745099/console_output.txt)
[symbolicated_crash.txt](https://github.com/electron/fiddle/files/12745100/symbolicated_crash.txt)

dsanders11:
Fixes #1470.

### DIFF
--- a/src/renderer/runner.ts
+++ b/src/renderer/runner.ts
@@ -331,19 +331,9 @@ export class Runner {
   }
 
   private buildChildEnvVars(): { [x: string]: string | undefined } {
-    const { isEnablingElectronLogging, environmentVariables } = this.appState;
+    const { environmentVariables } = this.appState;
 
-    const env = { ...process.env };
-
-    if (isEnablingElectronLogging) {
-      env.ELECTRON_ENABLE_LOGGING = 'true';
-      env.ELECTRON_DEBUG_NOTIFICATIONS = 'true';
-      env.ELECTRON_ENABLE_STACK_DUMPING = 'true';
-    } else {
-      delete env.ELECTRON_ENABLE_LOGGING;
-      delete env.ELECTRON_DEBUG_NOTIFICATIONS;
-      delete env.ELECTRON_ENABLE_STACK_DUMPING;
-    }
+    const env: Record<string, string> = {};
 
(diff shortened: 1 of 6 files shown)


src/renderer/runner.ts
```diff
@@ -333,7 +333,7 @@ export class Runner {
   private buildChildEnvVars(): { [x: string]: string | undefined } {
     const { isEnablingElectronLogging, environmentVariables } = this.appState;
 
-    const env = { ...process.env };
+    const env: Record<string, string> = {};
```
erikian, line 336: The `else` block on lines 342-345 can be removed, there's nothing to `delete` anymore

erikian (approved):

src/renderer/runner.ts
dsanders11: Good catch, didn't notice those lines. That logic should actually probably be shifted to the main process so that the behavior matches what it was before the refactoring. Otherwise if those variables are in the environment in general, there will be a behavior change from before the main process refactor.

I'll look at refactoring this PR.

dsanders11:
Updated this to better match the original behavior and added test coverage.

erikian (approved):
Minor nit, otherwise LGTM!

tests/main/fiddle-core-spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  return {
+    Installer: InstallerMock,
+    Runner: FiddleRunnerMock,
+  };
+});
+
+describe('fiddle-core', () => {
+  let runner: FiddleRunnerMock;
+  let env: NodeJS.ProcessEnv;
```
erikian, line 25: Can we make this somthing like `originalEnv`? There's a bunch of other `env`s elsewhere, this should make it slightly easier to tell them apart when skimming over the code

PR #1474, opened 2023-09-29
feat: basic support for ESM fiddles

dsanders11:
Will create an issue for this, but not going to block this PR on it:
- Handle case where a fiddle is loaded (local or Gist) and has multiple `main.{cjs,js,mjs}` files.


src/renderer/utils/editor-utils.ts
```diff
@@ -1,44 +1,65 @@
...
-  [MAIN_JS, `Main Process (${MAIN_JS})`],
-  ['renderer.js', 'Renderer Process (renderer.js)'],
-  ['index.html', 'HTML (index.html)'],
-  ['preload.js', 'Preload (preload.js)'],
-  ['styles.css', 'Stylesheet (styles.css)'],
-]);
-
 export function getEditorTitle(id: EditorId): string {
-  return TITLE_MAP.get(id) || id;
+  switch (id) {
```
felixrieseberg, line 39: I love how clean this entire section is, reads really well!

PR #1475, opened 2023-09-30
fix: mark fiddle as edited if files added/deleted

dsanders11:
Currently adding or deleting a file doesn't count as editing a fiddle which can lead to non-ideal behavior like adding a file without content and then switching selected versions and having the file disappear, or deleting a file and then closing the window and not being prompted to save.

### DIFF
--- a/src/renderer/editor-mosaic.ts
+++ b/src/renderer/editor-mosaic.ts
@@ -165,6 +165,8 @@ export class EditorMosaic {
     } else {
       this.hide(id);
     }
+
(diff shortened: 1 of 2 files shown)


erikian (approved):

aabdoaa:
> Currently adding or deleting a file doesn't count as editing a fiddle which can lead to non-ideal behavior like adding a file without content and then switching selected versions and having the file disappear, or deleting a file and then closing the window and not being prompted to save.

PR #1483, opened 2023-10-12
fix #1361 callback-to-promise

### LINKED ISSUE #1361: Show Me Examples Still Use Callback APIs
There's a handful of outdated show me examples which need to be migrated to using the promise APIs, e.g.

https://github.com/electron/fiddle/blob/f8deba908f8b72942002d89ef7350ee21b4848a5/static/show-me/session/main.js#L19-L21

shankeleven:
fixes #1361 
replaced outdated callbacks with promise.then()

### DIFF
--- a/static/show-me/session/main.js
+++ b/static/show-me/session/main.js
@@ -16,7 +16,7 @@ app.whenReady().then(() => {
   console.log(defaultSession.getUserAgent())
 
   // Cache Size
-  defaultSession.getCacheSize((result) => {
+  defaultSession.getCacheSize().then((result) => {
     console.log(result)
   })
 })


BlackHole1 (approved):

dsanders11 (changes requested):
Thanks for the PR, @shankeleven! The referenced code in that issue is just one example, to close that issue we need to address all cases. Do you plan to update this PR to fix all of them? Otherwise we'll need to tweak the description so that issue remains open.

[This old doc](https://github.com/electron/electron/blob/5-0-x/docs/api/promisification.md) should help with identifying APIs that were changed from callback to promises.

shankeleven:
Sure, i'll try to find and correct as many as i can

codebytere:
@shankeleven do you still plan to work on this?

shankeleven:
Sorry, i guess i wouldn't be able to do it due to some medical reasons, i apologise for taking the job and not being able to deliver

dsanders11:
No problem @shankeleven. Get well. ❤️

PR #1499, opened 2023-11-06
ci: remove branch filter

dsanders11:
This slipped in as part of #1498 but I don't think we want it here.


erickzhao (approved):
LGTM :shipit:

BlackHole1 (approved):

PR #1501, opened 2023-11-08
ci: restore usage of `load-release-secrets`

dsanders11:
This was removed to work around an issue that has now been resolved in the secrets service.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,6 +144,7 @@ jobs:
       - install
       - attach_workspace:
           at: .
+      - load-release-secrets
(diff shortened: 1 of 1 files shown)


felixrieseberg (approved):
Heads-up: I've removed it from two places (https://github.com/electron/fiddle/commit/b2a2051b6927faa834d4859c98e7ed33837e4e9c)

dsanders11:
Leaving it removed from the second spot for the moment, all we need is the `GITHUB_TOKEN` which is only needed in the publish step.

PR #1502, opened 2023-11-10
feat: add support for JSON files in editors

codebytere:
This PR adds support for JSON files. These can be used in Fiddle for data storage amongst other things - the impetus for this specifically is the ability to add `manifest.json` files for Chrome extensions.

### DIFF
--- a/src/renderer/components/sidebar-file-tree.tsx
+++ b/src/renderer/components/sidebar-file-tree.tsx
@@ -14,7 +14,7 @@ import { ContextMenu2, Tooltip2 } from '@blueprintjs/popover2';
 import classNames from 'classnames';
 import { observer } from 'mobx-react';
 
-import { EditorId } from '../../interfaces';
+import { EditorId, PACKAGE_NAME } from '../../interfaces';
 import { EditorPresence } from '../editor-mosaic';
 import { AppState } from '../state';
 import { isRequiredFile, isSupportedFile } from '../utils/editor-utils';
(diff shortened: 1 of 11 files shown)


dsanders11 (changes requested):
Running this locally, I get `package-lock.json` when I start up with 27.0.0 selected. Probably because we pull the content from https://github.com/electron/electron-quick-start/tree/27-x-y and that has a `package-lock.json`.

dsanders11:
Minor, but while you can't add a new file called `package.json`, if you rename an existing file to `package.json` it just disappears.

dsanders11 (approved):
Overall LGTM. Somewhat of a nit, but can you swap some of the hardcoded 'package.json' text with `PACKAGE_NAME`? It's nice to be consistent so searching finds all instances.

src/renderer/components/sidebar-file-tree.tsx
```diff
@@ -189,9 +189,19 @@ export const SidebarFileTree = observer(
 
       if (!id) return;
 
+      if (
+        id.endsWith('.json') &&
+        [PACKAGE_NAME, 'package-lock.json'].includes(id)
+      ) {
+        await appState.showErrorDialog(
+          'Cannot add package.json or package-lock.json as custom files',
```
dsanders11, line 197:
```suggestion
          `Cannot add ${PACKAGE_NAME} or package-lock.json as custom files`,
```

src/renderer/editor-mosaic.ts
```diff
@@ -140,8 +140,20 @@ export class EditorMosaic {
...
   /** Add a file. If we already have a file with that name, replace it. */
   private addFile(id: EditorId, value: string) {
-    if (!isSupportedFile(id))
-      throw new Error(`Cannot add file "${id}": Must be .js, .html, or .css`);
+    if (
+      id.endsWith('.json') &&
+      [PACKAGE_NAME, 'package-lock.json'].includes(id)
+    ) {
+      throw new Error(
+        'Cannot add package.json or package-lock.json as custom files',
```
dsanders11, line 148:
```suggestion
        `Cannot add ${PACKAGE_NAME} or package-lock.json as custom files`,
```

tests/main/utils/read-fiddle-spec.ts
```diff
@@ -47,10 +52,26 @@ describe('read-fiddle', () => {
     setupFSMocks(mockValues);
 
     const fiddle = await readFiddle(folder);
-
     expect(fiddle).toStrictEqual(mockValues);
   });
 
+  it('reads JSON files only if they are not package.json', async () => {
```
dsanders11, line 58:
```suggestion
  it(`reads JSON files only if they are not ${PACKAGE_NAME}`, async () => {
```

Issue #1504, opened 2023-11-13
Environment variables with extra '=' don't work


codebytere:
To reproduce:

Go to settings, the fill the environment variable section with the placeholder:

<img width="875" alt="Screenshot 2023-11-13 at 10 26 35 AM" src="https://github.com/electron/fiddle/assets/2036040/78f09c2f-eb92-4867-905f-0dcd56c7ee2d">

Run the Fiddle, and observe failure:

```console
Saving files to temp directory...
Saved files to /var/folders/x_/clmjt4bs0dx09xjy7k3xt1z80000gn/T/electron-fiddle-8492-ufiHikyVXPgI
Electron v9999999.9.9 started.
electron: invalid value for NODE_OPTIONS (unterminated string)

Electron exited with code 9.
```

PR #1505, opened 2023-11-13
chore: fix environment variable parsing

codebytere:
Closes https://github.com/electron/fiddle/issues/1504.

Fixes an issue where complex environment variables with multiple equal signs would be incorrectly parsed. This was happening because our env var parsing was [very simplistic](https://github.com/electron/fiddle/blob/main/src/renderer/runner.ts#L333-L343) and always split at equals signs.

We should use an existing env var parser instead, written and maintained by a Node.js maintainer. It has no dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "node-watch": "^0.7.3",
     "p-debounce": "^2.0.0",
     "package-json": "^7.0.0",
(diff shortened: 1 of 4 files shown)


VerteDinde (approved):
Nice!

felixrieseberg (approved):
Nice polish!

Issue #1523, opened 2024-01-05
`Follow Link` does not work in console or editor


codebytere:
<img width="522" alt="Screenshot 2024-01-05 at 6 38 34 PM" src="https://github.com/electron/fiddle/assets/2036040/d8a163ca-f04b-45aa-aee7-8ca671d8d19f">

Neither clicking on the tooltip directly nor `Cmd+Click` has any effect.

PR #1524, opened 2024-01-05
feat: enable opening external URLs in editor

codebytere:
Closes https://github.com/electron/fiddle/issues/1523.

Fixes an issue where external links did not function properly. The user can now open external links after confirming the link to open (for security purposes).

Screenshot 2024-01-05 at 6 42 48 PM


src/renderer/components/editor.tsx
```diff
@@ -113,6 +119,30 @@ export class Editor extends React.Component<EditorProps> {
...
+          })
+          .then((open) => {
+            if (open) {
+              window.ElectronFiddle.openExternal(url);
+            }
+          });
+      },
+    };
+  }
+
```
erikian, line 145: Can we move this to an utility function (maybe in `/src/utils/editor-utils.ts`) to avoid duplication?

erikian (approved):

dsanders11 (commented):
Do we need to add the IPC call? I think we can just use `window.open` in this case, see #1260.

src/renderer/components/editor.tsx
codebytere: i'm open to this but it feels a little strange passing appState to a util function 🤔

codebytere:
@dsanders11 good call - changed!

dsanders11 (approved):

src/renderer/components/editor.tsx
erikian: Fair point, we do access `appState` via `window.app.state` elsewhere though, this variable is so ubiquitous that I think it's acceptable — but either way is fine by me, I don't feel strongly.

BlackHole1 (approved):
LGTM

Issue #1529, opened 2024-01-13
Handle missing local Electron.app


codebytere:
At the moment, we handle the case where a folder _containing_ a local Electron app has changed or disappeared, but not where the actual binary is missing. This means that selecting an unavailable local version results in endless status check spin:

<img width="494" alt="Screenshot 2024-01-13 at 10 17 04 AM" src="https://github.com/electron/fiddle/assets/2036040/08e90bc7-acda-43cc-b623-06af79c9a24d">

PR #1530, opened 2024-01-13
fix: handle missing local Electron.app

codebytere:
Closes https://github.com/electron/fiddle/issues/1529

Handles the case where a local Electron executable has moved or been deleted.

Before:

Screenshot 2024-01-13 at 10 17 04 AM

After:

Screenshot 2024-01-13 at 10 29 54 AM

### DIFF
--- a/src/renderer/components/version-select.tsx
+++ b/src/renderer/components/version-select.tsx
@@ -66,8 +66,9 @@ const itemListRenderer: ItemListRenderer<RunnableVersion> = ({
  * @returns {string}
  */
 export function getItemLabel({ source, state, name }: RunnableVersion): string {
+  // If a version is local, either it's there or it's not.
   if (source === VersionSource.local) {
-    return name || 'Local';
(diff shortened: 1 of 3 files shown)


dsanders11 (approved):

src/renderer/components/version-select.tsx
```diff
@@ -66,8 +66,9 @@ const itemListRenderer: ItemListRenderer<RunnableVersion> = ({
  * @returns {string}
  */
 export function getItemLabel({ source, state, name }: RunnableVersion): string {
+  // If a version is local, either it's there or it's not.
   if (source === VersionSource.local) {
-    return name || 'Local';
+    return state === 'missing' ? 'Unavailable' : name || 'Local';
```
erikian, line 71:
```suggestion
    return state === InstallState.missing ? 'Unavailable' : name || 'Local';
```

erikian (approved):

PR #1537, opened 2024-01-25
test: increase timeout in ElectronTypes watch test

dsanders11:
Noticing this test started flaking after #1533, I think the filesystem operations in that test is delayed when it's running under Rosetta, causing the flaky timeout. Doing a quick and dirty bump of the timeout to mitigate the flake for now.


ckerr (approved):
LGTM w/the followup TODO to find a better answer.

Everyone (including me) always likes to complain about hacks like this, but :man_shrugging: there's a lot to be said for silencing test flakes.

Issue #1539, opened 2024-01-27
Release links broken on electronjs.org


ajs5710:
Wasn't sure where was appropriate to put this but it seems that the release links for fiddle are broken from documentation.

What is happening: 
In safari navigate to https://www.electronjs.org/fiddle and click on orange release buttons to download but get a 404 from GitHub.

What is expected:
Link goes to valid release download/download starts

Release links for me are going to something like: 
  https://github.com/electron/fiddle/releases/download/v0.36.0/Electron.Fiddle-darwin-arm64-0.36.0.zip
This results in a 404

Actual releases appear to be something more like:
  https://github.com/electron/fiddle/releases/download/v0.36.0/Electron-Fiddle-darwin-arm64-0.36.0.zip

Note the expected name of the zip begins with `Electron.Fiddle` but the actual name starts with `Electron-Fiddle`

Issue #1503, opened 2023-11-10
Broken Darwin downloads


mxschmitt:
Repro:

1. Navigate to https://www.electronjs.org/fiddle
2. Click on Download on darwin-arm64

It ends up using 

https://github.com/electron/fiddle/releases/download/0.34.5/Electron.Fiddle-darwin-arm64-0.34.5.zip

which is a 404 but it should use:

https://github.com/electron/fiddle/releases/download/v0.34.5/Electron.Fiddle-darwin-arm64-0.34.5.zip

(missing v version prefix)

mxschmitt:
Probably caused by https://github.com/electron/website/commit/c41589bce35c65e28459b4ab5515f408dc387a4c.

cc @dsanders11

dsanders11:
Thanks!

PR #1540, opened 2024-02-01
ci: check download links daily

dsanders11:
We've broken the download links on the website a few times and not noticed, so let's add some lightweight automation to confirm the links aren't 404ing after a new release.

Refs:
* electron/website/issues/467
* #1503
* #1539

### DIFF
--- a/.github/workflows/check-download-links.yml
+++ b/.github/workflows/check-download-links.yml
@@ -1,0 +1,44 @@
+name: Check Download Links
+'on':
+  workflow_dispatch: null
+  schedule:
+    - cron: 0 1 * * *
+permissions: {}
+jobs:
+  check-download-links:
+    name: Check Download Links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


erickzhao (changes requested):
The site is currently built on a cron if no new commits come in, so Fiddle download links get updated periodically instead of immediately on publish.

To make this work, I think we would need to dispatch https://github.com/electron/electron-website-updater events for Fiddle builds in the same way that we trigger commits every time the `e/e` docs update.

dsanders11:
Good catch, I'd forgotten the links won't update on the website immediately.

> To make this work, I think we would need to dispatch https://github.com/electron/electron-website-updater events for Fiddle builds in the same way that we trigger commits every time the e/e docs update.

There'd still be an indeterminate delay after the release was published before that website build would finish and go live, which risks the check not reviewing the latest state, which could provide a false negative.

How about for simplicity we just put this on a cron as well (say 1 AM PT)? That would catch broken download links faster than we currently do.

erickzhao (approved):
SGTM :shipit:

PR #1542, opened 2024-02-07
fix(show-me/tray): correct import

JEntler:
Changed import from 'electron.main' to 'electron/main'.


dsanders11 (approved):
Good catch, thanks!

PR #1544, opened 2024-02-09
fix: allows users to drag app when settings are open

yangannyx:
This PR fixes an issue where the app is not draggable when a user is on the Settings page. This is because the `tabbing-hidden` style applied to the header also nullifies the header's `-webkit-app-region: drag;` style.

This PR:

1. Moves the `tabbing-hidden` style to the `commands` component. This ensures that tabbing on the settings page does not focus these hidden elements.
2. Moves the `Dialogs` component to be rendered after `Header` and `OutputEditorsWrapper`. This ensures that when elements in the Settings page scroll over the header area, the header's `-webkit-app-region: drag;` doesn't swallow click events. I think this also makes sense from a `z-index` perspective as dialogs are usually rendered over main content.
3. Adds the `-webkit-app-region: no-drag;` style to all focusable elements in the settings component, similar to how `commands` applies this style.

Testing:
1. Verified that focused elements cycle through elements on the settings page when tabbing.
2. Verified that inputs, buttons, and selects in the settings content can be clicked when they are in the header region.
3. Verified that the app can be dragged in the header region when settings are open.

### DIFF
--- a/src/renderer/components/commands.tsx
+++ b/src/renderer/components/commands.tsx
@@ -36,15 +36,15 @@ export const Commands = observer(
 
     public render() {
       const { appState } = this.props;
-      const { isBisectCommandShowing, title } = appState;
+      const { isBisectCommandShowing, title, isSettingsShowing } = appState;
 
       return (
         <div
-          className={
+          className={`${
             window.ElectronFiddle.platform === 'darwin'
               ? 'commands is-mac'
               : 'commands'
-          }
+          }${isSettingsShowing ? ' tabbing-hidden' : ''}`}
           onDoubleClick={this.handleDoubleClick}
         >
           <div>
--- a/src/renderer/components/commands.tsx
+++ b/src/renderer/components/commands.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { Button, ControlGroup } from '@blueprintjs/core';
+import classNames from 'classnames';
 import { observer } from 'mobx-react';
 
 import { GistActionButton } from './commands-action-button';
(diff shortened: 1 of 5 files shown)


dsanders11:
Note: Windows test failure is an unrelated CircleCI issue.

src/renderer/components/commands.tsx
```diff
@@ -36,15 +36,15 @@ export const Commands = observer(
...
 
       return (
         <div
-          className={
+          className={`${
             window.ElectronFiddle.platform === 'darwin'
               ? 'commands is-mac'
               : 'commands'
-          }
+          }${isSettingsShowing ? ' tabbing-hidden' : ''}`}
```
dsanders11, line 47: This can be simplified using the `classnames` package which we use in some other components:

```tsx
className={classNames(
  'commands',
  { 'is-mac': window.ElectronFiddle.platform === 'darwin' },
  { 'tabbing-hidden': isSettingsShowing },
)}
```

dsanders11 (approved):
Nice! This has bugged me for a long time. 😄

Tested it out and it works as expected for me. Dropped one suggestion regarding code style. Test coverage would be nice, but I don't think it's trivial, so I'm fine with merging without it.

src/renderer/components/commands.tsx
yangannyx: Oh, neat! Thanks for the pointer

codebytere (approved):
sweet - thanks! also been bugging me for a long time 😁

yangannyx:
Thanks for the reviews @dsanders11 @codebytere! Heads up, it looks like I don't have write access to merge this PR, if either of yall want to merge it instead

Issue #1546, opened 2024-02-19
Confusing error if Gist contains invalid `package.json`


codebytere:
If a user tries to load a Gist containing invalid JSON in their package.json, a confusing error is currently thrown:

<img width="431" alt="Screenshot 2024-02-19 at 11 07 54 AM" src="https://github.com/electron/fiddle/assets/2036040/2d55d876-4984-4605-84cd-5a2eef99c1ac">

We should improve this.

PR #1547, opened 2024-02-19
fix: better error message for bad `package.json`

codebytere:
Closes https://github.com/electron/fiddle/issues/1546.

Fixes an issue I noticed whereby if a user tried to load a Gist containing invalid JSON in their package.json, a confusing error would be thrown.

<details><summary>Before</summary>
<p>

Screenshot 2024-02-19 at 11 07 54 AM

</p>
</details> 

<details><summary>After</summary>
<p>

Screenshot 2024-02-19 at 11 13 28 AM

</p>
</details>

### DIFF
--- a/src/renderer/file-manager.ts
+++ b/src/renderer/file-manager.ts
@@ -71,10 +71,10 @@ export class FileManager {
       if (name === PACKAGE_NAME) {
         const { remoteLoader } = window.app;
 
-        let deps: Record<string, string> = {};
+        const deps: Record<string, string> = {};
         try {
           const { dependencies, devDependencies } = JSON.parse(value);
-          deps = { ...dependencies, ...devDependencies };
+          Object.assign(deps, dependencies, devDependencies);
         } catch {
           await this.appState.showErrorDialog(
             'Could not open Fiddle - invalid JSON found in package.json',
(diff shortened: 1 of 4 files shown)


src/renderer/file-manager.ts
```diff
@@ -70,11 +70,16 @@ export class FileManager {
...
-        const { dependencies, devDependencies } = JSON.parse(value);
-        const deps: Record<string, string> = {
-          ...dependencies,
-          ...devDependencies,
-        };
+
+        let deps: Record<string, string> = {};
+        try {
+          const { dependencies, devDependencies } = JSON.parse(value);
+          deps = { ...dependencies, ...devDependencies };
```
dsanders11, line 77:
```suggestion
        const deps: Record<string, string> = {};
        try {
          const { dependencies, devDependencies } = JSON.parse(value);
          Object.assign(deps, dependencies, devDependencies);
```

Would let `deps` remain `const`, which is nice.

src/renderer/remote-loader.ts
```diff
@@ -130,11 +130,13 @@ export class RemoteLoader {
...
         if (id === PACKAGE_NAME) {
-          const { dependencies, devDependencies } = JSON.parse(content);
-          const deps: Record<string, string> = {
-            ...dependencies,
-            ...devDependencies,
-          };
+          let deps: Record<string, string>;
+          try {
+            const { dependencies, devDependencies } = JSON.parse(content);
+            deps = { ...dependencies, ...devDependencies };
```
dsanders11, line 136:
```suggestion
          const deps: Record<string, string>;
          try {
            const { dependencies, devDependencies } = JSON.parse(content);
            Object.assign(deps, dependencies, devDependencies);
```

Same as above.

dsanders11 (approved):

PR #1550, opened 2024-02-22
build: bump to Electron 29.1.4

VerteDinde:
This PR bumps Fiddle to use Electron 29.0.1


dsanders11 (changes requested):
Blocking until the bug preventing running is fixed.

dsanders11 (approved):
Updated this to bump to the latest stable release and include the fix in `@electron/fiddle-core` needed for E28+. 👍

PR #1552, opened 2024-02-23
chore: improve preload comment

yangannyx:
Simple text change to the comment at the top of the the default `preload.js` file. Clarifies what the preload script runs before.

Screenshot 2024-02-23 at 12 16 11 PM

### DIFF
--- a/static/electron-quick-start/preload.js
+++ b/static/electron-quick-start/preload.js
@@ -1,7 +1,8 @@
 /**
- * The preload script runs before. It has access to web APIs
- * as well as Electron's renderer process modules and some
- * polyfilled Node.js functions.
+ * The preload script runs before `index.html` is loaded
+ * in the renderer. It has access to web APIs as well as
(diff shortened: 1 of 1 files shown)


dsanders11 (changes requested):
Thanks for the PR! Agreed the current wording is less than ideal.

The content in Fiddle actually gets loaded from `electron/electron-quick-start` by looking for a branch there (e.g. `29-x-y`) and losing the content there first. The file you've touched in Fiddle is the fallback if the other is not found.

As such, let's first make this change in that repo so that these don't diverge, and then we can land it here.

See https://github.com/electron/electron-quick-start/blob/main/preload.js

yangannyx:
Thanks for the pointer! Made the same change to `electron/electron-quick-start` [here](https://github.com/electron/electron-quick-start/pull/892)

dsanders11 (approved):
Changed the title to be `chore:` rather than `fix:` as I think that's more accurate in this case.

PR #1559, opened 2024-03-12
I don't know how to resole the lint error i m getting 

hachimetsu:
error was in contribution.json ?
[lint error](https://app.circleci.com/pipelines/github/electron/fiddle/1391/workflows/1ee8aeda-e348-4f1e-b442-7aae2791909e/jobs/5759?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary)

[previous pull](https://github.com/electron/fiddle/pull/1558)

[changes i did](https://github.com/electron/fiddle/pull/1558/files)

### DIFF
--- a/src/renderer/components/output.tsx
+++ b/src/renderer/components/output.tsx
@@ -94,6 +94,14 @@ export const Output = observer(
             openerService: this.openerService(),
           },
         );
+        /**
+         * TODO: Clear Console on CTRL + K.
+         */
+        if(this.editor){
+          this.editor.addCommand(MonacoType.KeyMod.CtrlCmd | MonacoType.KeyCode.KEY_K, () => {
+            if(this.model) this.model.setValue("Console Cleared !");
+          });
+        }
       }
     }
 
--- a/src/renderer/components/output.tsx
+++ b/src/renderer/components/output.tsx
@@ -94,13 +94,13 @@ export const Output = observer(
             openerService: this.openerService(),
           },
         );
-        /**
-         * TODO: Clear Console on CTRL + K.
-         */
-        if(this.editor){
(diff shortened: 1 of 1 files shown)


hachimetsu:
can anyone help me with `text:cli`

hachimetsu:
Screenshot from 2024-03-12 11-32-05

error on test:cli

src/renderer/components/output.tsx
```diff
@@ -94,6 +94,14 @@ export const Output = observer(
             openerService: this.openerService(),
           },
         );
+        if (this.editor) {
+          this.editor.addCommand(
+            MonacoType.KeyMod.CtrlCmd | MonacoType.KeyCode.KEY_K,
+            () => {
+              if (this.model) this.model.setValue('Console Cleared !');
```
erickzhao, line 101: I don't think I'd want a message when clearing the console. 😅

hachimetsu: is that mean i m doing right ? 
or i m doing something else which is not actually required ?
what if i do `if (this.model) this.model.setValue("');`

hachimetsu: and i did use this `this.model.setValue('Console Cleared !');` because i use `CTRL+L` or`CTRL+SHIFT+L`
it will move the cursor to the top of the terminal window and clear all the text that was previously displayed. 
However, it doesn't erase the command history; you can still scroll up to see previous commands and output.

diivi, line 101:
```suggestion
              if (this.model) this.model.setValue('');
```
works for me

VerteDinde:
@hachimetsu Thanks for the PR! I'm going to close this because there has been no movement on it since March, but we appreciate you trying to contribute

PR #1561, opened 2024-03-16
fix: reload window button not working

gaspardruan:
When clicking the reload window button, nothing happened. The console reported the error
image
errors occur when reload window button is clicked

I just changed one line and it works fine.
```
- onClick={window.ElectronFiddle.reloadWindows}
+ onClick={() => window.ElectronFiddle.reloadWindows()}
```


codebytere (approved):

dsanders11 (approved):
Thanks!

PR #1584, opened 2024-04-25
fix: use async showOpenDialog

dsanders11:
The original issue behind using `showOpenDialogSync`, #553, no longer repros. `showOpenDialogSync` doesn't play nice with multiple Fiddle windows as it will block the UI on other windows.


BlackHole1 (approved):

erickzhao (approved):
LGTM

PR #1605, opened 2024-07-06
fix: limit usage of shell.openExternal to web URLs

dsanders11:
Follow [our own security guidelines](https://www.electronjs.org/docs/latest/tutorial/security#15-do-not-use-shellopenexternal-with-untrusted-content) here and limit what we pass through to `shell.openExternal`.

### DIFF
--- a/src/main/windows.ts
+++ b/src/main/windows.ts
@@ -24,6 +24,15 @@ export function mainIsReady() {
   mainIsReadyResolver();
 }
 
+export function safelyOpenWebURL(url: string) {
+  try {
+    const { protocol } = new URL(url);
+    if (['http:', 'https:'].includes(protocol)) {
+      shell.openExternal(url);
+    }
+  } catch {}
+}
+
 /**
  * Gets default options for the main window
  *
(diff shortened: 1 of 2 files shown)


yangannyx (approved):
LGTM!

src/main/windows.ts
```diff
@@ -24,6 +24,15 @@ export function mainIsReady() {
   mainIsReadyResolver();
 }
 
+function isWebProtocol(url: string) {
```
codebytere, line 27: non-blocking, but we could also do something like `safelyOpenWebURL`:

```js
function safelyOpenWebURL(url: string) {
   try {
      const { protocol } = new URL(url);
      if (['http:', 'https:'].includes(protocol)) {
        shell.openExternal(details.url);
      }
   } catch {}
}
```

where we could then call e.g.:

```js
safelyOpenWebURL(details.url);
```

instead of 

```js
if (isWebProtocol(details.url)) {
   shell.openExternal(details.url);
}
```

at the callsites?

dsanders11: Good suggestion! Implemented, and also added test coverage which I should have added from the start. 😅

erickzhao (approved):

PR #1616, opened 2024-07-14
fix: set parent window for open/save dialogs

### LINKED ISSUE #1572: Open/save dialog is not modal
On Linux, the open/save dialog is not modal, as I can click into the application behind it and make it disappear into the background.

### Steps to Reproduce

1. Open Fiddle
2. Go to File, Open
3. You see an Open dialog box
4. Click the publish button in the application.

### Expected behavior

Clicking into the application doesn't have any effect, and the open dialog remains open.

### Actual Behavior

On Linux, clicking in the application while the open dialog is present will hide the open dialog in the background. It will then become impossible to find again without using inconvenient techniques like Alt-Tab through all open windows.

### Information

Electron Fiddle: 0.36.2
OS: Ubuntu 22.04.4 LTS x86_64
Kernel: 5.15.0-101-generic
Desktop Environment: KDE Plasma
Window Manager: KWin
Theme: Breeze Dark [Plasma], Breeze [GTK2/3]
Icons: [Plasma], breeze-dark [GTK2/3]

### Context

I know the problematic behavior is not intrinsic to Linux because my other KDE applications work as expected, as does Firefox. So this seems specific to Fiddle and/or Electron. For example, my other KDE applications gray out the application when the open dialog is open (like in macOS) and clicking into the grayed out area doesn't background the open dialog.

dsanders11:
Fixes #1572.

### DIFF
--- a/src/main/dialogs.ts
+++ b/src/main/dialogs.ts
@@ -56,8 +56,10 @@ export function setupDialogs() {
 
   ipcMainManager.handle(
     IpcEvents.LOAD_LOCAL_VERSION_FOLDER,
-    async (): Promise<SelectedLocalVersion | undefined> => {
-      const folderPath = await showOpenDialog();
+    async (event): Promise<SelectedLocalVersion | undefined> => {
+      const folderPath = await showOpenDialog(
+        BrowserWindow.fromWebContents(event.sender)!,
+      );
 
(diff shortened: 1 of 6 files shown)


BlackHole1 (approved):
LGTM

PR #1618, opened 2024-07-29
test: migrate publish button tests to RTL

erikian:
Ref: #1408.

I'm also adding a helper function that allows us to access class components' instances, since RTL doesn't expose them and a bunch of our tests are messing with component internals directly; that's probably something we should move away from at some point, but for now that helper should make it a bit easier to migrate away from Enzyme.


dsanders11 (approved):

dsanders11:
> that's probably something we should move away from at some point, but for now that helper should make it a bit easier to migrate away from Enzyme.

I think this is reasonable. Migrating off of Enzyme will unblock us on other stuff and long-term we can further improve our testing methodology.

PR #1620, opened 2024-07-30
test: fix flake under Rosetta

dsanders11:
Follow-up to #1537. Re-ran the workflow 5 times to confirm it didn't flake.

Screenshot 2024-07-30 at 4 04 01 PM


erickzhao (approved):
yay :)

PR #1621, opened 2024-08-02
build: remove rimraf dev dependency

dsanders11:
Seems a bit unnecessary to pull in `rimraf` since Node.js now supports the `recursive` option.


MarshallOfSound (approved):

codebytere (approved):

tools/clean-webpack.ts
```diff
@@ -0,0 +1,4 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+fs.rmSync(path.join(__dirname, '.webpack'), { force: true, recursive: true });
```
erikian, line 4: `__dirname`-free alternative:
```suggestion
fs.rmSync(path.resolve('.webpack'), { force: true, recursive: true });
```

erikian (approved):

ckerr (approved):
:green_heart:

tools/clean-webpack.ts
dsanders11: Yea, the only reason I went with the `__dirname` version was I wanted a little bit of extra protection so it won't go deleting `${CWD}/.webpack` if a user inadvertently invoked it from a different CWD. Probably not an actual concern, but I don't love blindly deleting a relative path.

PR #1622, opened 2024-08-02
build: refactor tools scripts to remove third-party deps

dsanders11:
Minor step towards removing `cross-fetch` and `fs-extra` dependencies.

Also updates `log-symbols` and deals with the ESM fun.


ckerr (approved):
LGTM

PR #1626, opened 2024-08-05
build: upgrade `@testing-library` deps

erikian:
Keeping up with RTL dependencies:
- `@testing-library/react` can't be upgraded because v12 is the last version that supports React 16 and 17, so we're stuck with that until we can upgrade React
- only a minor change was required for the `@testing-library/jest-dom` v5 -> v6 migration
- the latest major version of `@testing-library/dom` just works with the current setup
- the latest version of `@testing-library/user-event` is just a patch update for us

Ref: #1408


dsanders11 (approved):
Thanks! Good to know about `@testing-library/react` being blocked by the React 18 upgrade.

PR #1652, opened 2024-11-22
feat: register-local-version protocol command

samuelmaddock:
Implements protocol handler for `electron-fiddle://register-local-version/` to allow registering local versions.

Example request:
> electron-fiddle://register-local-version/?name=feat%3A+service+worker+preload+scripts&version=35.0.0-dist&path=%2FUsers%2Fsamuelmaddock%2F.electron_build_tools%2Fartifacts%2Fpr_44411_darwin_arm64

This is intended to be used by build-tools as a way to quickly [download build artifacts](https://github.com/electron/build-tools/pull/679) and test them in Fiddle.

https://github.com/user-attachments/assets/e0a8f15b-fe16-4a12-a2d5-b6fc0c19b581


erickzhao (approved):
This is a really elegant implementation!

dsanders11 (changes requested):
This should have a confirmation dialog similar to the one we show when we load Fiddles from the docs.

https://github.com/electron/fiddle/blob/3452e12a67682aba8898019d1ae07df8e3194d61/src/renderer/remote-loader.ts#L251-L261

I'm also a bit weary from a security perspective to expose modifying Fiddle's list of local builds like this. I think it warrants stronger consideration of how it might be abused.

samuelmaddock:
@dsanders11 I've added a confirmation dialog. While testing I noticed that multiple requests will clobber the previous dialog. I've gone ahead and added a fix for that as well.


https://github.com/user-attachments/assets/7d320cad-d0ea-489d-9bf4-b613395f79e6

MarshallOfSound (changes requested):
This shouldn't be a protocol URL that can in theory be triggered by remote pages, make it a CLI argument to the fiddle binary instead and have build tools **spawn** fiddle, trigger the second instance event and handle that.

Also fiddle that confirm with the user they actually want to register this thing before it just Does What It Was Told

samuelmaddock:
> This shouldn't be a protocol URL that can in theory be triggered by remote pages, make it a CLI argument to the fiddle binary instead and have build tools **spawn** fiddle, trigger the second instance event and handle that.

I feel like this gets exceedingly difficult when there are multiple installations or apps installed outside of the default directory (especially on Windows).

> Also fiddle that confirm with the user they actually want to register this thing before it just Does What It Was Told

Added in the most recent commit.

PR #1679, opened 2025-03-05
fix: allow renaming between main entrypoint files

dsanders11:
Fixes a regression introduced in #1563 that somewhat broke #1474 since you can't rename `main.js` to `main.mjs`.


BlackHole1 (approved):
LGTM

PR #1687, opened 2025-03-06
fix: inconsistent outline color on focus when navigating using tab key

### LINKED ISSUE #1085: Inconsistent focus indicators throughout Fiddle
👋🏼  While tabbing around, I noticed that the focus indicators are inconsistent:
* Text inputs with the `bp3-input` class have a default blue outline
* Some input elements have a `background-color: @foreground-2;` style applied on `focus` (which I'm guessing is meant to change the outline)
* Other focusable elements (like checkboxes) have an additional white outline
* https://github.com/electron/fiddle/issues/1084

I've opened a separate issue for the buttons as this is a more severe accessibility issue, but the rest are more of a matter of styling and preference 🤷🏼‍♀️ 

Screenshot from the settings page for reference:

AlokYadavCodes:
Fix #1085  by ensuring consistent focus outlines for input fields and buttons in dark mode.
Previously, input fields and buttons had inconsistent focus outlines, leading to a poor user experience.

**All elements are focused in the images below:**
### Before:  
before1

### After:
after1

### Before:
before2

### After: 
after2

### Changes
Added `outline: @blue3 solid 2px;` to following classes in dark mode:

_.bp3-control input:focus ~ .bp3-control-indicator_

 _.bp3-button:focus_

### DIFF
--- a/src/less/blueprint.less
+++ b/src/less/blueprint.less
@@ -6,6 +6,10 @@
 
 // Override some of the colors
 .fiddle.bp3-dark {
+  .bp3-control input:focus ~ .bp3-control-indicator {
+    outline: @blue3 solid 2px;
+  }
+
   .bp3-menu,
   .bp3-popover .bp3-popover-content {
     background-color: @background-1;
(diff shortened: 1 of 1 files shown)


AlokYadavCodes:
@codebytere
It looks like this workflow requires maintainer approval to run. Could you please approve it when possible?

PR #1702, opened 2025-04-08
feat: allow packaging a local electron version

codebytere:
Allow packaging an app using forge with a local build of Electron using `@electron-forge/plugin-local-electron`.


erickzhao (approved):
LGTM

PR #1712, opened 2025-06-19
refactor: improve Electron settings version table

codebytere:
Prior to this the table loaded all possible electron versions with no real ability to filter or search.

This PR virtualizes the list and adds input filtering.

Screenshot 2025-06-19 at 10 37 23 AM


dsanders11 (changes requested):
When I tested this locally, it seems the download/delete button no longer works properly in the table. No visual update happens unless you scroll such that the button goes out of view, then scroll it back into view.

codebytere:
@dsanders11 fixed in [bff81a9](https://github.com/electron/fiddle/pull/1712/commits/bff81a9095612c6144ea5fb86e63e16a9ba18fae)!

dsanders11:
@codebytere, looks like that change works well if the version is at the top of the list, but if you're scrolled down the list and click one of the buttons, I'm seeing it snap back to the top of the list, and am unable to scroll down at all until the operation (download or delete) has finished.

codebytere:
@dsanders11 i think i've finally gotten it when you get a chance to re-review :D

BlackHole1 (approved):
LGTM. After local testing, the issue raised by @dsanders11 has been fixed.

dsanders11 (changes requested):
Can also confirm that issue is cleared up. 👍

There's a couple more visual things I think could use polish before merging.

1. The buttons are no longer rendering as equal width like they do in Fiddle currently, so the button changes sizes while downloading, and when it says "Delete".
2. If you type in gibberish for your filter, in dark mode the "No versions match your filter" text is dark when it should be light so that it's legible.

codebytere:
@dsanders11 sorry for accidental review request! will comment when changes are addressed :)

PR #1714, opened 2025-07-02
chore: update electron-quick-start references

yangannyx:
This PR updates references to `electron-quick-start` to the renamed repo `minimal-repro`.

Fiddle currently uses `electron-quick-start` in the folder names of its template caching logic. I decided not to change this logic since it's internal to how fiddle operates and not something general electron consumers will be exposed to


tests/mocks/npm-response-main.json
yangannyx, line 1: These mocks are unused to my knowledge

src/main/content.ts
```diff
@@ -13,7 +13,7 @@ import { IpcEvents } from '../ipc-events';
 // parent directory of all the downloaded template fiddles
 const TEMPLATES_DIR = path.join(app.getPath('userData'), 'Templates');
 
-// location of the fallback template fiddle used iff downloading failed
+// location of the fallback template fiddle used if downloading failed
```
erickzhao: I think this was intentional ([iff = if and only if](https://en.wikipedia.org/wiki/If_and_only_if))

dsanders11:
@yangannyx, I think this PR might be missing a commit? I don't see any actual changes to the `electron-quick-start` references.

BlackHole1:
@yangannyx, Is this PR still ongoing? If it is, it might be better to mark it as a Draft.

yangannyx:
Hey @dsanders11 , @BlackHole1 apologies for the late reply. This PR is complete, it removes the `electron-quick-start` references in the mocks that are deleted in commit https://github.com/electron/fiddle/pull/1714/commits/6690c62d38e38da4d7d9111bda5946c659df4414.
The remaining usages are in reference to the folder names where we cache copies of the minimal repro files for each electron major version. Since this is an implementation detail, I didn't think it was worth the lift of refactoring these folders to be migrated and prefixed `minimal-repro` instead of `electron-quick-start`. Let me know if you all think it's still a good idea!

PR #1716, opened 2025-07-06
docs: add Quickstart section to README

Misba0019:
Added a 'Quickstart' section after the download link and image in the README.
It guides new users through opening, editing, running, saving, and packaging their first Fiddle.


dsanders11:
Thanks for opening a PR. I don't think this adds much value to the README as it's pretty basic stuff that doesn't really help users get started with Fiddle. As such I'm going to close this PR.

PR #1717, opened 2025-07-09
build: update typescript and use @tsconfig/node22

dsanders11:
Need to bump TypeScript to be able to pick up the new version of `@electron/fiddle-core`, and extend from `@tsconfig/node22` to ensure the right values for things like `target` and `moduleResolution`.

Doing this surfaced ~100 errors when running `tsc` so this PR fixes them all up to get a clean build again. Some of these are legitimate things that were wrong but for some reason weren't being caught before (we were using `IpcMainEvent` in a lot of places where we should have been using `IpcMainInvokeEvent`), but some are just throwing in type assertions to keep things building. There should be no functional changes as a result of this PR, just types changes.


jest.config.ts
```diff
@@ -21,6 +21,7 @@ const config: Config = {
   collectCoverageFrom: ['src/**/*.{ts,tsx}', '!**/*.d.ts', '!**/*constants.ts'],
   coveragePathIgnorePatterns: ['/node_modules/', '/tests/.*\\.(ts|js)$'],
   moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1',
```
VerteDinde, line 24: For future viewers: asked David about this line and it's to handle ts-jest dynamic imports temporarily, lol

VerteDinde (approved):
Asked a few questions in Slack DMs, but looks good!

src/renderer/components/editors-toolbar-button.tsx
```diff
@@ -17,7 +17,7 @@ interface ToolbarButtonProps {
 
 abstract class ToolbarButton extends React.PureComponent<ToolbarButtonProps> {
   public static contextType = MosaicWindowContext;
-  public context: MosaicWindowContext;
```
dsanders11, line 20: Note, these `declare` changes were necessary and follow [the pattern in `@types/react`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d3a514567da8d4d146dab60eadda33bcd1e1b29c/types/react/v16/index.d.ts#L464-L480):

```ts
/**
 * If using the new style context, re-declare this in your class to be the
 * `React.ContextType` of your `static contextType`.
 * Should be used with type annotation or static contextType.
 *
 * ```ts
 * static contextType = MyContext
 * // For TS pre-3.7:
 * context!: React.ContextType<typeof MyContext>
 * // For TS 3.7 and above:
 * declare context: React.ContextType<typeof MyContext>
 * ```
 *
 * @see https://react.dev/reference/react/Component#context
 */
// TODO (TypeScript 3.0): unknown
context: any;
```

src/renderer/components/editors-toolbar-button.tsx
```diff
@@ -17,7 +17,7 @@ interface ToolbarButtonProps {
...
+  public declare context: MosaicWindowContext;
```
dsanders11, line 20: Note, these `declare` changes were necessary and follow [the pattern in `@types/react`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d3a514567da8d4d146dab60eadda33bcd1e1b29c/types/react/v16/index.d.ts#L464-L480):

```ts
/**
 * If using the new style context, re-declare this in your class to be the
 * `React.ContextType` of your `static contextType`.
 * Should be used with type annotation or static contextType.
 *
 * ```ts
 * static contextType = MyContext
 * // For TS pre-3.7:
 * context!: React.ContextType<typeof MyContext>
 * // For TS 3.7 and above:
 * declare context: React.ContextType<typeof MyContext>
 * ```
 *
 * @see https://react.dev/reference/react/Component#context
 */
// TODO (TypeScript 3.0): unknown
context: any;
```

PR #1718, opened 2025-07-10
test: switch to Vitest

dsanders11:
Stacks on top of #1717, and prerequisite to bumping `@electron/fiddle-core` since I don't want to mess with the rat's nest which is Jest + ESM.

The majority of these line changes are mechanical `jest` -> `vi`, but there are some non-trivial bits which I'll comment on inline.


tests/main/devtools-spec.ts
```diff
@@ -1,47 +1,59 @@
...
-      expect(e).toMatch('error');
-    }
+    vi.mocked(installExtension).mockRejectedValue(error);
+    vi.mocked(isDevMode).mockReturnValue(true);
+
+    await setupDevTools();
+    expect(console.warn).toHaveBeenCalledWith(
+      expect.stringMatching(/installDevTools/),
+      error,
+    );
```
dsanders11, line 57: This test wasn't actually testing what it intended to, since the `expect` inside of the `catch` block never actually ran. Inside of `setupDevTools` any exceptions are caught and logged with `console.warn`, so this test is updated to actually test that.

tests/renderer/components/dialogs-spec.tsx
dsanders11, line 1: Not sure why the DOM output changed here, but updated so the test goes green.

tests/main/dialogs-spec.ts
```diff
@@ -75,16 +78,16 @@ describe('dialogs', () => {
...
-      mocked(fs.existsSync).mockReturnValue(false);
-      mocked(dialog.showOpenDialog).mockResolvedValue({
+      vi.mocked(fs.existsSync).mockReturnValue(false);
+      vi.mocked(dialog.showOpenDialog).mockResolvedValue({
         filePaths: paths,
         canceled: false,
       });
 
-      expect(ipcHandler()).resolves.toStrictEqual({
+      await expect(ipcHandler()).resolves.toStrictEqual({
```
dsanders11, line 90: These weren't being awaited before, there was a warning from Vitest about it. Doesn't change the outcome of the test, but it's more correct now.

tests/mocks/electron-types.ts
dsanders11, line 1: More accurate types, that fell out of using `with: { type: 'json' }` for the import of our JSON fixtures.

tests/setup.ts
```diff
@@ -63,13 +63,24 @@ class FakeBroadcastChannel extends EventTarget {
...
+// Don't serialize Timeout objects in depth
+// TODO(dsanders11): This feels like we shouldn't need to do this?
+expect.addSnapshotSerializer({
+  serialize(val) {
+    return val.toString();
+  },
+  test(val) {
+    return val?.constructor.name === 'Timeout';
+  },
+});
```
dsanders11, line 75: Spun my wheels on this one for a bit, the `layoutDebounce` property in `EditorMosaic` was serializing to a full complex object, unlike a simple integer under Jest. The solution I came up with was to just use this custom serializer so that it serializes to `[object]` in the snapshot, which is fine for our needs.

tests/utils.ts
```diff
@@ -97,55 +106,30 @@ export function objectDifference(a: any, b: any): Record<string, unknown> {
...
-  timeout: number;
-}
-
-/**
- * Waits up to `timeout` msec for a test to pass.
- *
- * @param test - function to test
- * @returns a promise that returns the test result on success, or rejects on timeout
- */
-export async function waitFor(
```
dsanders11, line 119: Not needed, `vi.waitUntil` is an equivalent in Vitest.

tests/utils.ts
```diff
@@ -97,55 +106,30 @@ export function objectDifference(a: any, b: any): Record<string, unknown> {
...
 }
+
+/**
+ * Helper function to mock CommonJS `require` calls with Vitest.
+ *
+ * @see https://github.com/vitest-dev/vitest/discussions/3134
+ * @param mockedUri - mocked module URI
+ * @param stub - stub function to assign to mock
+ */
+export async function mockRequire(mockedUri: string, stub: any) {
```
dsanders11, line 126: This is cribbed from Forge.

tools/contributors.ts
```diff
@@ -35,12 +37,7 @@ interface ContributorInfo {
   location: string;
 }
 
-// Helper function to work around import issues with ESM module
-const dynamicImport = new Function('specifier', 'return import(specifier)');
```
dsanders11, line 39: No longer needed under Vitest.

tests/utils.ts
```diff
@@ -84,7 +87,13 @@ export class FetchMock {
...
 // return an object containing props in 'a' that are different from in 'b'
 export function objectDifference(a: any, b: any): Record<string, unknown> {
-  const serialize = (input: any) => JSON.stringify(toJS(input));
+  const serialize = (input: any) =>
+    JSON.stringify(toJS(input), (_key, value) => {
+      if (value?.constructor.name === 'Timeout') {
+        return value[Symbol.toPrimitive]();
+      }
+      return value;
+    });
```
dsanders11, line 96: There was an error regarding stringifying a circular structure due to `layoutDebounce` in `EditorMosiac`, so this is done to break the cycle.

tests/renderer/components/__snapshots__/editors-spec.tsx.snap
```diff
@@ -901,7 +901,7 @@ exports[`Editors component renders a toolbar 1`] = `
       appState={
         {
           "editorMosaic": {
-            "layoutDebounce": 26,
+            "layoutDebounce": [object Object],
```
dsanders11, line 904: This is what the custom snapshot serializer does for us.

tests/main/utils/exec-spec.ts
dsanders11, line 1: Cleaned up the test implementaiton in here.

tests/renderer/electron-types-spec.ts
```diff
@@ -40,7 +44,9 @@ describe('ElectronTypes', () => {
     } as const;
 
     electronTypes = new ElectronTypes(monaco as any);
-    nodeTypesData = require('../fixtures/node-types.json');
+    ({ default: nodeTypesData } = await import('../fixtures/node-types.json', {
+      with: { type: 'json' },
+    }));
```
dsanders11, line 49: Gets us some improved typing on the shape of the fixture.

tests/renderer/electron-types-spec.ts
```diff
@@ -121,17 +133,23 @@ describe('ElectronTypes', () => {
...
+        types,
+      );
+      vi.mocked(window.ElectronFiddle.getNodeTypes).mockResolvedValue({
         version: version.version,
         types: nodeTypesData
-          .flatMap((entry) => (entry.files ? entry.files : entry))
+          .flatMap((entry) =>
+            entry.type === 'directory'
+              ? (entry as NodeTypesDirectory).files
+              : entry,
```
dsanders11, line 151: Changed as a result of more accurate types.

tests/renderer/components/sidebar-file-tree-spec.tsx
```diff
@@ -217,10 +218,10 @@ describe('SidebarFileTree component', () => {
     setTimeout(() => {
       expect(editorMosaic.files.get('index.html')).toBe(EditorPresence.Visible);
       expect(editorsInstance.state.focused).toBe('index.html');
-    });
+    }, 1000);
```
BlackHole1: Is there a specific reason for doing this? It seems that this will increase the duration of unit tests.

tests/renderer/state-spec.ts
```diff
@@ -75,23 +84,23 @@ describe('AppState', () => {
   });
 
   afterAll(async () => {
-    // Wait for all the async task to resolve before the jest
+    // Wait for all the async task to resolve before the
```
BlackHole1:
```suggestion
    // Wait for all the async task to resolve before the vitest
```

tests/renderer/components/sidebar-file-tree-spec.tsx
dsanders11: Good catch, this was something I intended to return to but forgot to when fixing other things. There's a debounce timeout in the code that was rendering this change necessary here, but it looks like making use of `vi.useFakeTimers()` lets me revert the longer timeout change. Fixed in db20456525005c8d7761ec04fa955c970d52ee89.

PR #1720, opened 2025-07-17
chore: add Devtron in development mode

dsanders11:
Now that Devtron is being revived, let's use it here. 🙂

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -43,7 +43,9 @@ export async function onReady() {
     process.env.NODE_ENV = 'production';
   } else {
     const { devtron } = await import('@hitarth-gg/devtron');
-    await devtron.install();
+    await devtron.install().catch(err => {
+      console.warn(`cannot install devtron: ${err.message}`)
+    });
   }
 
   setupAboutPanel();
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -39,7 +39,12 @@ let argv: string[] = [];
  */
 export async function onReady() {
   await onFirstRunMaybe();
-  if (!isDevMode()) process.env.NODE_ENV = 'production';
+  if (!isDevMode()) {
+    process.env.NODE_ENV = 'production';
+  } else {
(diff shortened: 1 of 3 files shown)


src/main/main.ts
```diff
@@ -39,7 +39,12 @@ let argv: string[] = [];
  */
 export async function onReady() {
   await onFirstRunMaybe();
-  if (!isDevMode()) process.env.NODE_ENV = 'production';
+  if (!isDevMode()) {
+    process.env.NODE_ENV = 'production';
+  } else {
+    const { devtron } = await import('@hitarth-gg/devtron');
+    await devtron.install();
```
BlackHole1:
```suggestion
    await devtron.install().catch(err => {
      console.warn(`cannot install devtron: ${err.message}`)
    });
```

When the installation fails, the entire application should not be rendered unable to open.

BlackHole1 (changes requested):

src/main/main.ts
dsanders11: I believe `devtron.install()` is made to [catch any errors itself and log](https://github.com/electron-userland/devtron/blob/7e10e55d26b7e112fc27acc04a928f90456bea49/src/index.ts#L132-L134) to prevent this, but sure, we can be defensive here just in case. 👍

src/main/main.ts
```diff
@@ -39,7 +39,14 @@ let argv: string[] = [];
  */
 export async function onReady() {
   await onFirstRunMaybe();
-  if (!isDevMode()) process.env.NODE_ENV = 'production';
+  if (!isDevMode()) {
+    process.env.NODE_ENV = 'production';
+  } else {
+    const { devtron } = await import('@hitarth-gg/devtron');
+    await devtron.install().catch(err => {
+      console.warn(`cannot install devtron: ${err.message}`)
```
BlackHole1:
```suggestion
    await devtron.install().catch((err) => {
      console.warn(`cannot install devtron: ${err.message}`);
```

BlackHole1 (approved):

PR #1726, opened 2025-07-24
feat: add support to update github gist token without sign out

### LINKED ISSUE #407: Allow to change token without signout

vhashimotoo:
Fixes #407

This PR adds support to change the GitHub token without especially sign out from the previous one. The previous token is also not be removed until the new one is confirmed as valid.

image


tests/renderer/components/dialog-token-spec.tsx
```diff
@@ -186,6 +188,22 @@ describe('TokenDialog component', () => {
       expect(store.gitHubToken).toEqual(null);
     });
 
+    it('handles the invalid token update', async () => {
```
codebytere, line 191: We should also check that a valid token update works as expected!

PR #1727, opened 2025-07-24
feat: improve error dialog when gist publish is failing

### LINKED ISSUE #1549: Better UX for expired PAT for Gist publishing
When a GitHub PAT expires, we pop up a dialog that says that the auth was invalid. To remedy this, the user needs to manually sign out in the settings afterwards and generate a new token.

As a UX improvement, we should redirect to the settings prompt as an option in the dialog, or at least clear the invalid token.

vhashimotoo:
Fixes #1549

This PR improves the UX of the error dialog that is showed when the user tries to publish a Gist with invalid credentials (or with some other error). This updates the dialog and adds the button that will move the user to the settings page and opens the token update window when the error is related to bad credentials and shows the standard error dialog if something other happened.

image

### DIFF
--- a/src/renderer/components/commands-action-button.tsx
+++ b/src/renderer/components/commands-action-button.tsx
@@ -16,6 +16,7 @@ import { observer } from 'mobx-react';
 import {
   EditorId,
   EditorValues,
+  GenericDialogType,
   GistActionState,
   GistActionType,
   PACKAGE_NAME,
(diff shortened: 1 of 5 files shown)


codebytere (changes requested):
I'm not sure I think we should always show an update token button for failure dialogs - there are a lot of reasons publish could fail.

PR #1728, opened 2025-07-28
build: use the node22 branch of @electron/windows-sign

dsanders11:
We're going to do a new Fiddle release and use the opportunity to validate this E2E.


VerteDinde (approved):
Let’s try it!

PR #1731, opened 2025-07-29
chore: fix TS compilation

dsanders11:
This slipped in during #1722 but wasn't caught until the release jobs ran because our CI was never running `tsc`, and the error was just being logged as a warning during tests rather than a fatal error. This PR updates our CI to ensure we run `tsc`, and then fixes the error that slipped in.


erickzhao:
pushed up a commit cuz CI was failing

dsanders11:
Oddly that one only happened some of the time. 🤔 Not sure what's up with that being intermittent. I think it may have to do with the actual content on disk for the contributors JSON at the time, but not sure why that would differ - would have thought it would be an all or nothing thing.

erickzhao (approved):

PR #1734, opened 2025-07-29
build: fix contributors fetch script silently failing

dsanders11:
This has been intermittently biting us in CI and now manifests in a TypeScript error after recent changes that are doing better typing for the imported JSON file.

The existing script wasn't checking if the fetch response was good before using it, which led to a situation where the initial fetch of contributors would succeed, but then fetching further details on each contributor would get rate limited and silently fail causing partial data.

I took the opportunity to clean up the script in general, and made some tweaks to the CI workflows to minimize the odds of being rate limited by GitHub. The CircleCI jobs are still at risk since it's non-trivial to give them a token to avoid it, but when we one day migrate to GHA that should go away.

### DIFF
--- a/tools/contributors.ts
+++ b/tools/contributors.ts
@@ -32,9 +32,9 @@ interface ContributorInfo {
   api: string;
   login: string;
   avatar: string;
-  name: string;
-  bio: string;
-  location: string;
+  name: string | null;
+  bio: string | null;
+  location: string | null;
 }
 
 export async function maybeFetchContributors(silent?: boolean): Promise<void> {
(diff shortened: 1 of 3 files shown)


tools/contributors.ts
```diff
@@ -87,71 +87,67 @@ export async function maybeFetchContributors(silent?: boolean): Promise<void> {
...
+      const details = await fetchDetailsContributor(contributor);
+
+      return {
+        ...contributor,
+        name: details.name,
+        bio: details.bio,
+        location: details.location,
+      };
+    }),
+  );
```
BlackHole1: If concurrency is not used here, but instead a simple for…await, would it reduce the risk of failure?

Ref: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits

> Make too many concurrent requests. No more than 100 concurrent requests are allowed. This limit is shared across the REST API and GraphQL API.

dsanders11: @BlackHole1, I think it's a good suggestion! It might help, and if it doesn't it has little adverse impact on the overall runtime, so I think it's worth trying. Thanks!

BlackHole1 (approved):
LGTM

PR #1738, opened 2025-07-30
build: update `@electron/windows-sign`

erickzhao:
cc @dsanders11


dsanders11:
Closing this as we've rolled back in #1740, will try `@electron/windows-sign` again once we migrate to GHA.

PR #1745, opened 2025-08-07
feat: allow access to fiddle history

codebytere:
This PR adds the ability to optionally roll back to previous versions of Gists. The option can be enabled and disabled via Settings, and is disabled by default.

<details><summary>Screenshots</summary>
<p>

Screenshot 2025-08-07 at 4 35 35 PM

Screenshot 2025-06-23 at 11 21 22 AM

Screenshot 2025-06-23 at 11 21 25 AM

</p>
</details>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
         run: chmod +x tools/add-macos-cert.sh && . ./tools/add-macos-cert.sh
       - name: Signing Manager Setup (Windows)
         if: ${{ startsWith(matrix.os, 'windows-') }}
-        uses: digicert/ssm-code-signing@fb61e357690ad6aaa11c372000c37fb74d35c000 # v1.1.1
+        uses: digicert/ssm-code-signing@af7f4b07694120e514fe5b19f0598af3c20256e3 # v1.2.0
       - name: Write authentication cert to disk (Windows)
         if: ${{ startsWith(matrix.os, 'windows-') }}
         shell: bash
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@vitest/coverage-v8": "3.2.4",
     "copy-webpack-plugin": "^11.0.0",
     "css-loader": "^6.7.1",
-    "electron": "^39.1.1",
+    "electron": "^40.1.0",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.7",
     "enzyme-to-json": "^3.6.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3305,7 +3305,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/node@npm:*, @types/node@npm:>= 8, @types/node@npm:^22.5.5, @types/node@npm:^22.7.7":
+"@types/node@npm:*, @types/node@npm:>= 8, @types/node@npm:^22.5.5":
   version: 22.18.8
   resolution: "@types/node@npm:22.18.8"
   dependencies:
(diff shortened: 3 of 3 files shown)


tests/renderer/components/__snapshots__/settings-general-gist-spec.tsx.snap
```diff
@@ -0,0 +1,27 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dsanders11: Looks like this snapshot is outdated, probably this branch existed before the Vitest migration. Can you nuke it and update the snapshot?

dsanders11 (changes requested):
In addition to the inline comment, the `catch (error)` blocks need to be updated to `catch (error: any)` (or do additional type checking) to make `tsc` happy.

src/renderer/remote-loader.ts
```diff
@@ -115,13 +126,40 @@ export class RemoteLoader {
...
+        gist_id: gistId,
+      });
+
+      console.log('Gist revisions', revisions.data);
+
+      return revisions.data.map((r) => ({
+        sha: r.version,
+        date: r.committed_at,
+        changes: r.change_status,
+      }));
```
erickzhao: suggestion: Should we reverse this list? Intuitively, I would have guessed that "Revision 3" was the latest one and "Revision 1" was the oldest, but it seems like this is reversed in the actual.

image

src/renderer/state.ts
```diff
@@ -536,6 +542,7 @@ export class AppState {
       this.save(GlobalSetting.acceleratorsToBlock, this.acceleratorsToBlock),
     );
     autorun(() => this.save(GlobalSetting.packageAuthor, this.packageAuthor));
+    autorun(() => this.save(GlobalSetting.isShowingGistHistory, false));
```
erickzhao: issue(blocking): I think this conflicts with the following snippet from line 121 above:

```ts
  public isShowingGistHistory = !!(
    this.retrieve(GlobalSetting.isShowingGistHistory) ?? true
  );
```

From my testing, the first load shows the Gist history feature but subsequent loads turn the setting off without explicitly doing so.

src/renderer/components/history.tsx
```diff
@@ -0,0 +1,187 @@
...
+    return (
+      <li
+        key={revision.sha}
+        className="revision-item"
+        onClick={() => this.handleRevisionSelect(revision)}
+      >
+        <div className="revision-content">
+          <h4>
+            <Icon icon="history" className="revision-icon" />
+            {`Revision ${index + 1}`}
```
erickzhao: suggestion: can we say "Created" for the first entry instead of "Revision 1"?

src/renderer/components/history.tsx
```diff
@@ -0,0 +1,187 @@
...
+        />
+      );
+    }
+
+    if (revisions.length === 0) {
+      return (
+        <NonIdealState
+          icon="history"
+          title="No Revision History"
+          description="This Gist doesn't have any previous revisions"
```
erickzhao:
```suggestion
          description="This Gist doesn't have any revisions"
```

src/renderer/components/history.tsx
```diff
@@ -0,0 +1,187 @@
...
+    this.state = {
+      isLoading: true,
+      revisions: [],
+      error: null,
+    };
+  }
+
+  public async componentDidMount() {
+    await this.loadRevisions();
+  }
```
erickzhao: I'm not sure why, but this lifecycle function triggers when the modal closes as well, which doubles the necessary amount of network requests.

src/renderer/remote-loader.ts
```diff
@@ -115,13 +126,40 @@ export class RemoteLoader {
     }
   }
 
+  public async getGistRevisions(gistId: string): Promise<GistRevision[]> {
+    console.log('getGistRevisions!');
```
erickzhao: suggestion: should we remove this log line if we already have the response logged below?

src/renderer/components/history.tsx
codebytere: fixed!

erickzhao (dismissed):
Two things to change after checking it out again:

* Can we reverse the list such that the latest revision is the first list we see?
* In the current iteration, I notice that if the Enable History setting is already on in my app, I get a "No Gist ID" error message until I toggle the setting off/on again.

dsanders11 (changes requested):
Played around with this some more and a couple comments:

* Can we filter out revisions that are +0/-0? It looks like there's an existing bug where the initial upload is pushing two revisions, the second is empty. Clicking update with no changes will also push up an empty revision. We can fix these, but for existing gists I think it makes sense to filter these out - there's no reason you'd want to go to an empty commit in a revision history, that I can think of?

* Can we save off the revision ID and in the UI show which one is the active revision? It's easy to switch between them, but then it's also very easy to get confused as to which revision you're now on. 😅

codebytere:
@dsanders11 finally got back to this - pushed up last few suggestions!

PR #1762, opened 2025-10-08
build: upgrade Yarn v4

erickzhao:


.github/workflows/coveralls.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           node-version: 22.x
           cache: yarn
       - name: Install
-        run: yarn --frozen-lockfile --network-timeout 100000 || yarn --frozen-lockfile --network-timeout 100000 || yarn --frozen-lockfile --network-timeout 100000
+        run: yarn --immutable || yarn --immutable || yarn --immutable
```
MarshallOfSound: Ain't no way we still need to run it three times?

erickzhao: LOL it's a fallback in case the Yarn registry fails that we added way back when. I don't think it hurts to keep this because I don't know if upgrading Yarn fixes that particular issue.

MarshallOfSound: No other repo needs this 🤷 Doesn't hurt it's just funky

erickzhao: https://github.com/electron/fiddle/pull/1762/commits/71661b88956fb252e4af67f891c4f257312fda33

PR #1796, opened 2025-11-05
build(deps): use Electron 39

erickzhao:
So we're not EOL ourselves. :)


erickzhao:
Holding due to Electron 39.1 crash

PR #1878, opened 2026-04-02
build: Electron Forge 8 Alpha 7

erickzhao:
Mostly testing how things shake out after the ESM + Node 22 PR.


src/main/main.ts
```diff
@@ -229,7 +225,4 @@ export function main(argv_in: string[]) {
   });
 }
 
-// only call main() if this is the main module
-if (typeof module !== 'undefined' && require.main === module) {
-  main(process.argv);
-}
+main(process.argv);
```
dsanders11, line 228: Do we know what originally motivated this? Wondering if we should keep it.

erickzhao: Not sure but as a note I specifically changed this around because `require.main === module` was tripping over `module: preserve`.

dsanders11: Context for the change here: https://github.com/electron/fiddle/pull/618/changes/b30d4c33f1916983eb7abecb16da67aa3bb8c0c4

I'm a little concerned this might cause a regression in tests, but at the same time they're going green. 🤔

dsanders11 (approved):

PR #1917, opened 2026-05-05
ci: use D:\ drive for Windows CI runners

erickzhao:
## Summary

ref https://github.com/electron/forge/pull/4038

- Redirect `YARN_GLOBAL_FOLDER` and `TEMP` to the faster D:\ scratch disk on GitHub Actions Windows runners
- Applied to both `build.yml` and `test.yml` workflows
- Matches the approach already used in Electron Forge

## Test plan
- [x] Verify Windows CI jobs pass on this PR
- [x] Confirm no regression on macOS/Linux jobs (step is gated with `if: runner.os == 'Windows'`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dsanders11 (approved):
Unclear if this actually helps our situation here as the build times aren't markedly different, but doesn't hurt to give this a try.

ckerr (approved):

PR #1920, opened 2026-05-05
build: remove package-json dependency

ckerr:
Does what it says on the tin: remove `package-json` dependency.


mlaurencin (approved):

jkleinsc (approved):

erickzhao (approved):

src/main/electron-types.ts
```diff
@@ -197,12 +196,22 @@ export class ElectronTypes {
...
-      const types = await packageJson('@types/node', {
-        version: semver.major(version).toString(),
-        fullMetadata: false,
-      });
+      const url = 'https://registry.npmjs.org/@types%2Fnode';
+      const headers: HeadersInit = {
+        Accept: 'application/vnd.npm.install-v1+json',
+      };
+      const res = await fetch(url, { headers });
+      const data = (await res.json()) as { versions: Record<string, unknown> };
```
dsanders11: We need to check `res.ok` before parsing it as JSON.

dsanders11 (changes requested):
One specific change requested inline, but overall I'd feel more comfortable with this change if we pulled it out into a helper util and added test coverage of it.

src/main/electron-types.ts
ckerr: Good call. Added in c062733. I changed it to throw, which is what package-json did in this case.

ckerr:
> One specific change requested inline, but overall I'd feel more comfortable with this change if we pulled it out into a helper util and added test coverage of it.

Moved into a helper w/tests in 3b20bfbca5ffd32ffca0bd2386ffab987cfa6142

dsanders11 (approved):

PR #1926, opened 2026-05-06
chore: security hardening renderer→main IPC surface

jkleinsc:
This PR adds additional hardening of our IPC calls from the renderer:

1. `NPM_PACKAGE_RUN` — Command Injection
`packageRun()` was calling `exec()` with a shell string (`${packageManager} run ${command}`), enabling shell injection via a crafted `command` value. 
**Fix:** Replaced with `execFile(dir, pm, ['run', command])`. Arguments are passed as an array; no shell is involved.

2. `SAVE_FILES_TO_TEMP` — Path Traversal → Arbitrary File Write
Filename keys in the `files` Map were written directly via `fs.outputFile(path.join(dir.name, name), content)` with no validation, allowing `../../../.bashrc` style names to escape the temp directory.
**Fix:** Added `isSafeFileName()` helper that rejects names containing `..`, `/`, or the OS path separator. Unsafe filenames are skipped with a console warning.

3. `START_FIDDLE` — Multiple Hardening Fixes -
Three related vulnerabilities were addressed:
* Dangerous Env Key Injection
`Object.assign(env, params.env)` merged the renderer-supplied `env` object directly into `process.env` before spawning, allowing keys like `LD_PRELOAD` and `DYLD_INSERT_LIBRARIES` to hijack dynamic linker behaviour.
**Fix:** Added a `BLOCKED_ENV_KEYS` set (`LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, `DYLD_FRAMEWORK_PATH`, `DYLD_LIBRARY_PATH`). These keys are stripped from `params.env` before the merge.
* Arbitrary Working Directory
The renderer-supplied `dir` was used as `cwd` for the spawned process without validation, allowing Electron to be launched from any directory on the system.
**Fix:** Added `isInsideTempDir()` guard — `startFiddle` throws immediately if `dir` does not resolve to a path inside `os.tmpdir()`.
* Renderer-Controlled Executable Path
`isValidBuild` was supplied by the renderer and trusted to decide whether `localPath` was a valid Electron build directory, allowing the renderer to steer process spawning toward an arbitrary binary.
**Fix:** The `isValidBuild` parameter is ignored. Instead, `startFiddle` always calls `Installer.getExecPath(localPath)` and uses `fs.existsSync()` itself to verify the resolved executable exists before using it. The renderer never controls which code path is taken.
* Null Byte Injection in CLI Arguments
Renderer-supplied `options` strings were passed to `spawn()` without filtering, allowing null bytes to truncate argument strings at the OS level.
**Fix:** `options` is filtered to retain only strings that contain no null bytes (`\0`).

4. `PATH_EXISTS` — Filesystem Oracle
`fs.existsSync()` was called on any renderer-supplied string, allowing arbitrary filesystem probing.
**Fix:** Added `isAllowedPath()` helper that accepts only absolute paths whose resolved value falls under one of four known roots: `os.tmpdir()`, `app.getPath('appData')`, `ELECTRON_DOWNLOAD_PATH`, or `ELECTRON_INSTALL_PATH`. All other paths — including relative paths and paths containing `..` — return `false` without calling `existsSync`.

5. `CREATE_THEME_FILE` / `READ_THEME_FILE` — Path Traversal (fixed)
`name` was joined onto `THEMES_PATH` without validation, so `../../.config/evil` could read or write files outside the themes directory.
**Fix:** Both `readThemeFile()` and `createThemeFile()` verify the resolved path starts with `THEMES_PATH + path.sep` after `path.join`. Out-of-bounds paths are rejected (null return / thrown error respectively).

6. `GET_TEMPLATE_VALUES` — Path Traversal Read (fixed)
`name.toLowerCase()` was joined onto the `show-me` static directory with no validation.
**Fix:** `getTemplateValues()` returns a rejected Promise if `name` contains `..`, `/`, or the OS path separator.

7. `NPM_ADD_MODULES` — Unvalidated `dir` (fixed)
The renderer-supplied `dir` was passed directly to `execFile` without confirming it was inside the temp directory, allowing module installs (and their `postinstall` scripts) to run in arbitrary directories.
**Fix:** `addModules()` resolves `dir` and throws if it is not inside `os.tmpdir()`.

#### Test Coverage

All fixes are covered by new or updated tests:

| Test file | Changes |
|---|---|
| `tests/main/npm.spec.ts` | `packageRun` assertions updated to `execFile` array form; `addModules` dirs updated to tmpdir; added out-of-tmpdir rejection test |
| `tests/main/files.spec.ts` | Added `saveFilesToTemp` unsafe-filename skip test; added `PATH_EXISTS` allowlist tests (rejects relative paths, `..`, and paths outside all roots; accepts paths inside `os.tmpdir()` and `app.getPath('appData')`) |
| `tests/main/fiddle-core.spec.ts` | Added tests: blocked env keys stripped; dir outside tmpdir rejected; `localPath` executable validated via `fs.existsSync`; null bytes stripped from options |
| `tests/main/themes.spec.ts` | Added traversal-rejection tests for `readThemeFile` and `createThemeFile` |
| `tests/main/templates.spec.ts` | Added traversal-rejection test for `getTemplateValues` |

All 82 tests pass (6 pre-existing failures in `tests/main/utils/exec.spec.ts` are unrelated to these changes).


src/main/files.ts
```diff
@@ -38,12 +39,49 @@ function isInsideTempDir(dir: unknown): dir is string {
...
+function isSafeFileName(name: unknown): name is string {
+  return (
+    typeof name === 'string' &&
+    name.length > 0 &&
+    !path.isAbsolute(name) &&
+    !name.includes('..') &&
+    !name.includes('/') &&
+    !name.includes(path.sep)
+  );
+}
```
dsanders11: This is near identical to `isSafeDataName` a bit higher up in this file, I think we can extra the commonality here.

src/main/files.ts
```diff
@@ -38,12 +39,49 @@ function isInsideTempDir(dir: unknown): dir is string {
...
+  });
+}
+
 /**
  * Ensures that we're listening to file events
  */
 export function setupFileListeners() {
   ipcMainManager.on(IpcEvents.PATH_EXISTS, (event, filePath: string) => {
-    if (typeof filePath !== 'string') {
+    if (!isAllowedPath(filePath)) {
```
dsanders11: We're dropping `PATH_EXISTS` in #1927 so let's drop this line and `isAllowedPath`. This wouldn't work regardless, as `PATH_EXISTS` is used to check user supplied local build paths which can be anywhere on their system.

src/main/templates.ts
```diff
@@ -12,6 +12,11 @@ import { IpcEvents } from '../ipc-events';
  * Returns expected content for a given name.
  */
 export function getTemplateValues(name: string): Promise<EditorValues> {
+  if (name.includes('..') || name.includes('/') || name.includes(path.sep)) {
```
dsanders11:
```suggestion
  if (path.basename(name) !== name) {
```

We can take advantage of `path.basename` stripping everything for us here.

src/main/themes.ts
```diff
@@ -25,6 +25,11 @@ export async function readThemeFile(
   const file = name.endsWith('.json') ? name : `${name}.json`;
   const themePath = path.join(THEMES_PATH, file);
 
+  if (!themePath.startsWith(THEMES_PATH + path.sep)) {
+    console.warn(`readThemeFile: rejected path outside themes dir: ${name}`);
+    return null;
+  }
```
dsanders11, line 30:
```suggestion
  const file = name.endsWith('.json') ? name : `${name}.json`;

  if (path.basename(file) !== file) {
    console.warn(`readThemeFile: rejected path outside themes dir: ${file}`);
    return null;
  }
  
  const themePath = path.join(THEMES_PATH, file);
```

Again, let's use `path.basename`.

src/main/themes.ts
```diff
@@ -55,6 +60,12 @@ export async function createThemeFile(
   const file = name.endsWith('.json') ? name : `${name}.json`;
   const themePath = path.join(THEMES_PATH, file);
 
+  if (!themePath.startsWith(THEMES_PATH + path.sep)) {
```
dsanders11: Same as above.

tests/main/files.spec.ts
```diff
@@ -351,4 +354,74 @@ describe('files', () => {
...
+
+    // Only the safe filename should be written
+    expect(fs.outputFile).toHaveBeenCalledTimes(1);
+    expect(fs.outputFile).toHaveBeenCalledWith(
+      expect.stringContaining('main.js'),
+      '',
+    );
+  });
+
+  describe('PATH_EXISTS IPC handler', () => {
```
dsanders11: Drop this test coverage since we're removing `PATH_EXISTS`.

dsanders11 (changes requested):

dsanders11 (approved):

ckerr (approved):

PR #1927, opened 2026-05-06
chore: move local builds to being stored in main process

dsanders11:
This turned into a whole big thing, but essentially this moves the storage of the local build paths out of the renderer and into the main process.

* Does a one time (enforced by main process) migration of `local-electron-versions` in `localStorage` to the main process
* Uses an opaque token from the main process when adding a new local build so the renderer can reference it but clearly isn't dictating a specific local path
* Drops the overpowered `PATH_EXISTS` IPC since it's no longer needed


src/renderer/versions.ts
```diff
@@ -9,6 +9,32 @@ import {
...
+    }
+
+    const accepted = window.ElectronFiddle.migrateLocalVersions(versions);
+    if (accepted) {
+      window.localStorage.removeItem(GlobalSetting.localVersion);
+    }
+  } catch {
+    // Corrupt data - remove it
+    window.localStorage.removeItem(GlobalSetting.localVersion);
+  }
```
ckerr, line 29: All of these code paths lead to the same `removeItem()` calls _except_ for the impossible/unsupported state of having valid local versions but main has already migrated.

```suggestion
  try {
    const versions: Array<Version> = JSON.parse(raw);
    if (Array.isArray(versions) && versions.length !== 0)
      window.ElectronFiddle.migrateLocalVersions(versions);
  } finally {
    window.localStorage.removeItem(GlobalSetting.localVersion);
  }
```

ckerr (approved):

PR #1928, opened 2026-05-06
feat: add github token management to `main` & IPC endpoints

ckerr:
Part 1 in a series (of 2?) to move GitHub token handling out of the renderer process.

This interim step:

- adds octokit wrangling & SafeStorage token storage to the main process
- adds new IPC endpoints to use thes new features
- adds new tests for the new features.

TODO:

- Wire UI up to this IPC
- Remove the code that's currently duplicated between main and renderer, eg Octokit interaction
- Delete the token from local storage


tests/mocks/electron.ts
```diff
@@ -173,6 +173,15 @@ const systemPreferences = {
   getUserDefault: vi.fn(),
 };
 
+const safeStorage = {
+  isEncryptionAvailable: vi.fn().mockReturnValue(true),
```
dsanders11:
```suggestion
  isEncryptionAvailable: vi.fn(() => true),
```

tests/main/github.spec.ts
```diff
@@ -0,0 +1,739 @@
...
+  vi.mocked(safeStorage.decryptString).mockImplementation((buffer: Buffer) => {
+    const str = buffer.toString();
+    return str.startsWith('encrypted:') ? str.slice('encrypted:'.length) : str;
+  });
+}
+
+describe('github', () => {
+  beforeEach(() => {
+    vi.mocked(fs.existsSync).mockReturnValue(false);
+    restoreSafeStorageMocks();
```
dsanders11:
```suggestion
describe('github', () => {
  beforeEach(() => {
    vi.mocked(fs.existsSync).mockReturnValue(false);
```

With the associated suggested change in `tests/mocks/electron.ts` the need for this should go away as we reset mocks after each test already.

dsanders11 (changes requested):
Left some suggested comments on how we can drop `restoreSafeStorageMocks`.

tests/main/github.spec.ts
ckerr: Added in dfe9a6ea5d6604b3ca9b3726ff3d1a8e3fc8cde6

tests/mocks/electron.ts
ckerr: dfe9a6ea5d6604b3ca9b3726ff3d1a8e3fc8cde6

dsanders11 (approved):

PR #1935, opened 2026-05-07
refactor: move more of start fiddle flow to main process

dsanders11:
Currently when a fiddle is started, the `START_FIDDLE` IPC event is sent from the renderer to the main process, with all the parameters needed to start the fiddle. This PR changes the flow to drop the parameters on `START_FIDDLE` and instead add a new `GET_START_FIDDLE_OPTIONS` IPC request from main -> renderer to get those parameters. This lets us decouple the starting of a fiddle and will let us isolate the run button.

This also lets us consolidate more of the run fiddle logic into the main process and simplifies the flow instead of having a split brain where the renderer was driving some of the process. As a result we can drop the `CLEANUP_DIRECTORY` and `DELETE_USER_DATA` IPC events.


ckerr (approved):

dsanders11:
Note that 6c4a09f8ff377cf0f83750805a10f0083298f4b8 is a drive by fix for an existing bug since I was in the area and it annoyed me.

ckerr (approved):

ckerr (approved):

jkleinsc (approved):

PR #1940, opened 2026-05-08
feat: add `GITHUB_FETCH_EXAMPLE` to main

ckerr:
Third in a [series](https://github.com/electron/fiddle/pull/1928) to move GitHub token + gist management into the main process.

This PR:

1. Adds "fetch example" to `src/main/github.ts`
2. Adds an IPC endpoint `GITHUB_FETCH_EXAMPLE` to invoke it
3. Adds test coverage for the endpoint
4. Updates the renderer to fetch examples using this IPC method instead of using Octokit.


src/renderer/constants.ts
```diff
@@ -1,6 +1,3 @@
-export const ELECTRON_ORG = 'electron';
-export const ELECTRON_REPO = 'electron';
```
dsanders11, line 2: nit(non-blocking): should we move these to `src/main/constants.ts`?

src/main/github.ts
```diff
@@ -1,12 +1,14 @@
 import * as fs from 'node:fs';
-import * as path from 'node:path';
+import { join as pathJoin } from 'node:path';
```
dsanders11, line 2: nit(non-blocking): personally I prefer the existing pattern of `path.join`

dsanders11 (approved):
Two non-blocking comments.

src/renderer/constants.ts
ckerr: For now I think it's fine as-is since they're only used in github.ts, so the current narrower scope makes sense to me. If they are needed outside of that file, then I'm +1 on moving them to main/constants

*edit:* although, better for them to be  symbolic constants than magic strings. I've re-added `ELECTRON_ORG`, `ELECTRON_REPO` in `github.ts` in aa16c07bf10e23b8c772665f73849df479c160d8

src/main/github.ts
ckerr: In general I agree.

I avoided importing `path` to avoid a shadowed variable with the `path` arg to an octokit method which requires an object with a `path` property.

So we either have a minor `pathJoin` wart, or a minor `{ path: examplePath }` wart, or a minor shadowed variable wart. I chose `pathJoin` because it feels least bad to me but I can see arguments for all three choices :shrug:

PR #1942, opened 2026-05-08
refactor: move GitHub tokens to main

ckerr:
Part six in a [series](https://github.com/electron/fiddle/pull/1928) to move GitHub token + gist management into the main process.

This PR:

- Finishes updating the renderer code to make IPC calls to use the main process' renderer code.
- Removes all references to Octokit from the renderer
- Removes the token information from localStorage
- Removes dead octokit-related code
- Updates tests


src/renderer/components/commands-action-button.tsx
```diff
@@ -22,15 +22,12 @@ import {
...
 import { ensureRequiredFiles } from '../utils/editor-utils';
-import { getOctokit } from '../utils/octokit';
 
 interface GistActionButtonProps {
   appState: AppState;
 }
 
 interface IGistActionButtonState {
-  readonly isUpdating: boolean;
-  readonly isDeleting: boolean;
```
dsanders11, line 33: Nice cleanup, looks like this has been dead code for a while.

src/renderer/app.tsx
```diff
@@ -161,7 +161,17 @@ export class App {
...
 
-    window.ElectronFiddle.sendReady();
+    // Restore signed-in state from main's encrypted credential, if any.
+    // Wait for auth restore before signalling ready so that queued IPC
+    // messages (e.g. deep-linked private gist loads) use the authenticated
+    // Octokit instance.
+    window.ElectronFiddle.gitHubCheckAuth()
+      .then(({ login }) => {
+        this.state.gitHubLogin = login;
+      })
```
dsanders11, line 175:
```suggestion
      })
      .catch((e) => console.warn('Failed to check GitHub auth status', e))
```

dsanders11 (changes requested):
I think there's a regression here in regards to Internet access that we should resolve before merging.

From my reading of the code, if you successfully logged in, and then at a later time start Fiddle while offline, we'll keep the token on disk but the renderer will show the user as not being logged in with GitHub. Becoming online while Fiddle is still running would leave the app in a logged out state even though the user has a valid token.

I think the behavior we want is that when offline the user should still be shown as logged in as long as they have a token on disk. They should be able to click the log out button and clear the token. Logging in while offline will fail.

If we don't use the user's login name anywhere in the renderer, we may want to change `ElectronFiddle.gitHubCheckAuth()` to return `{ loggedIn: boolean }` or something like that. In the offline case the main process would still return `{ loggedIn: true }` if there's a token on disk, as that's all the renderer needs to know.

ckerr:
> From my reading of the code, if you successfully logged in, and then at a later time start Fiddle while offline, we'll keep the token on disk but the renderer will show the user as not being logged in with GitHub. Becoming online while Fiddle is still running would leave the app in a logged out state even though the user has a valid token.

Good call. 76dbba09b6f031ea6e16a8bcb56654eb65654d0a

src/main/github.ts
```diff
@@ -161,52 +159,42 @@ async function handleTokenSignIn(
...
-
     if (error?.status === 401 || error?.status === 403) {
+      octokit_ = null;
       deleteToken();
+      return { login: null, hasToken: false };
     }
 
-    return { login: null };
+    // If we're offline, don't 't invalidte the token or octokit_.
+    // Keep them as-is for use when we're back online.
```
dsanders11:
```suggestion
    // If we're offline, don't invalidate the token or octokit_.
    // Keep them as-is for use when we're back online.
```

dsanders11 (approved):

PR #1943, opened 2026-05-09
refactor: use more production code, fewer mocks in github tests

ckerr:
Fifth in a [series](https://github.com/electron/fiddle/pull/1928) to move GitHub token + gist management into the main process.

This is a test cleanup PR:

- Adds support for `app.setPath()` in the Electron mock
- Throws an error if the Electron mock's `app.getPath()` is called with an unexpected key.
- Makes `STATIC_DIR` in `main/constants` work in both production and test environments
- In github tests, remove the mocks for
    - node:fs
    - fs-extra
    - getTemplate()
- in github tests, try to not rely on implementation details, eg not mocking/spying low-level filesystem calls. The tests shouldn't care which of `fs.existsSync()` or `fsPromises.exists()` is called by production code.

### DIFF
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -1,8 +1,21 @@
+import * as fs from 'node:fs';
 import * as path from 'node:path';
 
 import { app } from 'electron';
 
-export const STATIC_DIR = path.resolve(__dirname, '../static');
+// Find the root dir for static assets (eg `show-me/`, `electron-quick-start/`).
+// In production, the bundled main script lives in `.webpack/main/` and webpack
+// copies static assets to `.webpack/static/`.
+// In tests (vitest loads the source TypeScript directly), `__dirname` is
+// `src/main/` and the static folder lives at the repository root.
+function resolveStaticDir(): string {
+  const paths = ['../static', '../../static'].map((p) =>
+    path.resolve(__dirname, p),
+  );
+  return paths.find(fs.existsSync) ?? paths[0];
+}
+
+export const STATIC_DIR = resolveStaticDir();
 
 export const ELECTRON_DOWNLOAD_PATH = path.join(
   app.getPath('userData'),
--- a/src/main/constants.ts
+++ b/src/main/constants.ts
@@ -1,4 +1,3 @@
-import * as fs from 'node:fs';
 import * as path from 'node:path';
 
 import { app } from 'electron';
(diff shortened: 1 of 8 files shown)


src/main/constants.ts
```diff
@@ -1,8 +1,21 @@
...
+// In tests (vitest loads the source TypeScript directly), `__dirname` is
+// `src/main/` and the static folder lives at the repository root.
+function resolveStaticDir(): string {
+  const paths = ['../static', '../../static'].map((p) =>
+    path.resolve(__dirname, p),
+  );
+  return paths.find(fs.existsSync) ?? paths[0];
+}
+
+export const STATIC_DIR = resolveStaticDir();
```
dsanders11:
```suggestion
export const STATIC_DIR = path.resolve(__dirname, process.env.VITEST ? '../../static' : '../static');
```

Not sure we need a dynamic way to resolve this if it's just the two possible paths.

tests/main/github.spec.ts
```diff
@@ -85,67 +117,36 @@ function mockOctokitInstance(overrides: Record<string, any> = {}) {
...
+    fs.rmSync(userDataPath, { recursive: true, force: true });
+  });
 
   describe('handleTokenSignIn()', () => {
-    it('signs in with valid token formats', async () => {
+    it('saves encrypted tokens to a permission-protected userData file', async () => {
       mockOctokitInstance();
+      expect(loadToken()).toBeNull();
 
+      const expectedSignInResult = { success: true, login: MOCK_LOGIN };
```
dsanders11: There are warnings that this is unused.

dsanders11 (changes requested):
Left some concrete comments, but I also have a meta comment.

I understand the motivation here of removing some of the implementation details leaking into tests regarding which FS operations are called, but I also have some concerns about unmocking the FS operations and letting tests write to the FS as tests run amock could leave a lot of garbage on disk. These tests are still making an implementation assumption about the `userData` path being where GitHub tokens are written to, so if that changes then the writes go somewhere other than intended.

[Vitest recommends](https://vitest.dev/guide/mocking/file-system) using `memfs` so that you don't have to mock individual FS calls, but you also don't write to disk. Unfortunately I think `memfs` has a larger dependency graph than we'd want to pull in at the moment.

So not blocking this PR, but I do wonder if there's a happy middle ground, before we start ripping out all the FS mocks. Perhaps something like an FS mock that wraps the real FS operations but ensures all paths are inside of `tmpdir` to prevent any unintended FS operations from the tests if we miss something.

tests/main/github.spec.ts
ckerr: Fixed in 5990348485c3007a1f8c693e4274d90d191bd12a

ckerr:
> These tests are still making an implementation assumption about the `userData` path being where GitHub tokens are written to, so if that changes then the writes go somewhere other than intended.

> Perhaps something like an FS mock that wraps the real FS operations but ensures all paths are inside of tmpdir to prevent any unintended FS operations from the tests if we miss something.

I've added a safeguard in the `beforeEach()` call in 8e66d1d21314b35afd5d2f80d7622fda63ef6519 to confirm that the credentials file is inside of our tmpdir. This should prevent the github tests from doing any mutating calls elsewhere. 

There are other tests in the file that read from the static template directory. Since those are read-only, there's zero risk to the live filesystem, it exercises more production code, and it simplifies the tests since we don't have to mock the templates, IMO the template part of this is a win.

ckerr:
@dsanders11 are there any other changes you'd like to see?

dsanders11 (approved):

PR #1952, opened 2026-05-22
feat: isolate run button

dsanders11:
Isolates the run button to a cross-origin iframe to ensure it's rendered as an out-of-process iframe (OOPIF), and is isolated in a separate process to protect against a compromised renderer. This prevents the renderer from being able to programmatically run fiddles, and only the main process or user interaction can start them.


src/preload/preload.ts
```diff
@@ -44,37 +44,70 @@ const channelMapping: Record<FiddleEvent, IpcEvents> = {
...
+
+// This preload runs in every frame in the default session, so
+// expose a different API surface depending on the protocol.
 async function preload() {
-  await setupFiddleGlobal();
+  if (location.protocol === ISOLATED_ACTIONS_PROTOCOL) {
+    setupIsolatedActionsGlobal();
+  } else {
+    await setupFiddleGlobal();
+  }
```
ckerr, line 89: Let's add an `isMainFrame` check here, exposing only isolated if we're not the main frame

```suggestion
  if (isMainAppFrame()) {
     await setupFiddleGlobal();
  } else {
    setupIsolatedActionsGlobal();
  }
```

src/main/isolated-actions.ts
```diff
@@ -0,0 +1,95 @@
...
+): WebFrameMain | null {
+  const frames = webContents.mainFrame?.framesInSubtree;
+  if (!frames) return null;
+  return (
+    frames.find((frame) => {
+      try {
+        return new URL(frame.url).protocol === `${ISOLATED_ACTIONS_SCHEME}:`;
+      } catch {
+        return false;
+      }
```
ckerr: We might want to gate on more than just the protocol? For example we could check the host and pathname too. That would limit the "run fiddle" permissions a little more tightly.

Not sure if I'm overthinking this, open to push back on this. If you're onboard, other related changes inline with `(limit-run-permissions-to-run-button)` notation.

```suggestion
      return isIsolatedRunButtonDocumentUrl(frame.url);
```

with

```ts
const RUN_BUTTON_ENTRY_PATH_PREFIX = `/${RUN_BUTTON_ENTRY_NAME}/`;

export function isIsolatedRunButtonDocumentUrl(rawUrl: string): boolean {
  try {
    const url = new URL(rawUrl);
    return (
      url.protocol === `${ISOLATED_ACTIONS_SCHEME}:` &&
      url.host === ISOLATED_ACTIONS_RUN_BUTTON_HOST &&
      url.pathname === '/'
    );
  } catch {
    return false;
  }
}
```

src/main/isolated-actions.ts
```diff
@@ -0,0 +1,95 @@
...
+      return new Response(null, { status: 404 });
+    }
+
+    let pathname = url.pathname;
+
+    // Refuse path traversal — anything with `..` could escape the
+    // upstream output directory once we resolve it.
+    if (pathname.includes('..')) {
+      return new Response(null, { status: 400 });
+    }
```
ckerr: This won't safeguard against encoded dots. Not sure what the risk level is here, but there are safer ways to test for this, eg decodeURIComponent + path.posix.normalize(decoded) + normalized.startsWith(entryDirSuffix)

src/preload/preload.ts
dsanders11: Not sure about this change, only because in the future there may be another protocol that'll need different a different preload as well. If we isolate the editors themselves, they'll need additional third-party code (`monaco-editor`, etc) so we'll want to separate those as well. I think going off the protocol here will scale better for future changes.

ckerr (approved):

PR #1957, opened 2026-06-04
fix: replace extract-zip with @electron-internal/extract-zip

erickzhao:
Swaps the `extract-zip` dependency for the `@electron-internal/extract-zip` drop-in replacement.

- Renamed the dependency in `package.json` (`@electron-internal/extract-zip@^1.0.1`)
- Updated import/require specifiers in source
- Added `@electron-internal/*` to the npm age-gate allowlist in `.yarnrc.yml`
- Refreshed `yarn.lock`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dsanders11 (changes requested):
Putting requested changes on this just because I'd like to let `@electron-internal/extract-zip` bake a bit more before we pick it up here.

dsanders11:
@erickzhao, ready to approve this if you want to bump it to version 1.0.5 and pick up those fixes. 🙂

dsanders11 (approved):

dsanders11:
@erickzhao, I think we'll need to merge in `main` as well here to be mergeable, if I do it Farday will be a problem. 😁

erickzhao:
Done!

PR #1960, opened 2026-06-05
ci: use macos-latest-large for x64 macOS release job

dsanders11:
Latest release job hung, most likely due to OOM which we've run into in the past. Bump to `macos-latest-large` which has more RAM (30GB vs 14GB) but is still Intel macOS.


erickzhao (approved):
chonky

PR #1966, opened 2026-06-10
test: run tests under ELECTRON_AS_NODE

dsanders11:
In the future we should look at moving tests to run inside Electron itself to ensure we catch issues, but for now let's use `ELECTRON_RUN_AS_NODE` so we are at least sure we're using the same Node.js version as our `electron` package.


erickzhao (approved):
lgtm

PR #2004, opened 2026-09-04
build(win32): sign with Azure Trusted Signing via OIDC

felixrieseberg:
Replaces the DigiCert KeyLocker flow for Windows code signing with Azure Trusted Signing, authenticated through GitHub OIDC.

**Changes**

- `forge.config.ts`: a shared `@electron/windows-sign` options builder for the Squirrel and MSIX makers. It writes the Trusted Signing metadata file, points signtool at the dlib, uses the ACS timestamp server and SHA-256 only. Returns `undefined` when no Azure variables are set, so local and CI builds stay unsigned.
- `forge.config.ts`: MSIX publisher updated to the Trusted Signing certificate subject (`CN=OpenJS Foundation, O=OpenJS Foundation, L=San Francisco, S=California, C=US`). No MSIX has shipped in a public release, so no publisher bridging is needed.
- `release.yml`: `azure/login` replaces the three DigiCert steps, `id-token: write` added to the build job, Trusted Signing account and endpoint as literals (they are resource names, not secrets), identity IDs from repository secrets.
- `@electron/windows-sign` added as a devDependency for the type import.

**Verification**

Two runs of a temporary Windows-only workflow on this branch (since removed) exercised the full path:

- https://github.com/electron/fiddle/actions/runs/33822462896 — OIDC login, Squirrel installer signed, `signtool verify /pa /v` passed with timestamp
- https://github.com/electron/fiddle/actions/runs/33823193313 — same plus the MSIX maker with the new publisher; all installers verified

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,6 +32,8 @@ jobs:
     permissions:
       actions: write
       contents: read
+      # Required for the OIDC login to Azure used for Windows code signing
+      id-token: write
     environment: release
     steps:
       - run: git config --global core.autocrlf input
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
@claude review

dsanders11 (commented):
Rather than adding `@electron/windows-sign` to `package.json` only for a type (we'll have to keep it up-to-date and in-sync with Forge), can we just derive the type directly from Forge?

```ts
type WindowsSignOptions = NonNullable<
  MakerMSIX['config']['windowsSignOptions']
>;
```

erickzhao (approved):

